Issue #2608, opened 2026-04-18
ops: plugin freshness rotation for running workspaces


deruelle:
## What

The Soleur plugin is installed in each workspace via a symlink from `<workspace>/plugins/soleur` to `$SOLEUR_PLUGIN_PATH` (default `/app/shared/plugins/soleur`). "Latest" is tied to image deploy cadence — when the container image updates, all running workspaces see the new plugin on the next symlink dereference.

Gap: there is no mechanism to force-update the plugin for a running workspace between image deploys (e.g., hotfix a skill without redeploying the whole web-platform). Design one of:

- Versioned plugin paths (`/app/shared/plugins/soleur-vX.Y.Z/`) with a rotator.
- An API endpoint that re-points the symlink for a named workspace.
- Scheduled plugin refresh from a known-good source at session start.

## Why deferred

Not an MU1 AC. Deploy cadence is currently fast enough that plugin-hotfix latency is not operationally painful.

## Re-evaluation criteria

- First time a critical plugin hotfix needs to land faster than a full container deploy permits.

## References

- Parent: #1448
- Symlink logic: `provisionWorkspace` + `provisionWorkspaceWithRepo` in `apps/web-platform/server/workspace.ts`.
- Runbook: `knowledge-base/engineering/ops/runbooks/mu1-signup-workspace-verification.md` (Known Deferrals).

PR #3046, opened 2026-04-29
fix(web-platform): populate /mnt/data/plugins/soleur via image-baked seed (#3045)

### LINKED ISSUE #3045: investigate: /mnt/data/plugins/soleur mount is empty in production
## Discovered during #3033 investigation

While verifying the Layer 3 canary mount path in PR for #3033, the deepen-plan pass turned up that `/mnt/data/plugins/soleur` is **also empty** in production — the same class of "configured-but-not-populated mount" that #3033 found for `/mnt/data/apps`.

The canary container in `apps/web-platform/infra/ci-deploy.sh` mounts this path read-only into `/app/shared/plugins/soleur`:

\`\`\`
-v /mnt/data/plugins/soleur:/app/shared/plugins/soleur:ro
\`\`\`

But:

- Cloud-init's runcmd creates `/mnt/data/plugins/soleur` as an empty directory (line 299 of `cloud-init.yml`) — no `git clone`, no rsync, no terraform provisioner populates it.
- No CI workflow rsyncs plugin content there.
- App code references the mount target at runtime:
  - `apps/web-platform/server/workspace.ts:39`
  - `apps/web-platform/server/cc-dispatcher.ts:387`
  - `apps/web-platform/server/agent-runner.ts:542`

`workspace.ts:381-384` warns on failure rather than throwing, suggesting this might be a best-effort no-op — but that needs explicit confirmation. If the symlink fails silently, the canary's verification surface for plugin-aware code paths is **non-existent**, not "false negative" but "no signal at all."

## Action required

1. Audit the three call sites above and document what they do when the mount is empty.
2. Decide whether the empty-mount behavior is **intended** (best-effort, treat as no-op) or a **latent bug** (callers expect content, get unexplained empty-dir behavior).
3. If latent bug: file a fix that either populates the mount via Terraform/cloud-init/CI, or removes the mount and updates callers to a different source-of-truth.

## Re-evaluation criteria

Close when: a written disposition exists for each of the three call sites, AND either (a) the mount is removed cleanly with callers updated, or (b) a reproducible mechanism populates the mount across fresh-server provisioning and existing-server updates.

## Why this is filed separately from #3033

Per `wg-when-an-audit-identifies-pre-existing` — pre-existing issues identified during investigation get tracked, not folded into the current PR. The #3033 fix is scoped to Layer 3; the plugins mount is a separate (and possibly older) configuration drift.

deruelle:
## Summary

Populates `/mnt/data/plugins/soleur` on prod hosts so the Soleur Agent SDK can
load skills/agents/MCP servers in user-facing web-platform sessions. The mount
has been silently empty since the web-platform MVP commit (`5b8e2420`) — every
session was returning "skill not found" or degrading to a generic Claude.

- Vendor plugin tree into `apps/web-platform/_plugin-vendored` at build time, `COPY` into image at `/opt/soleur/plugin`, then seed `/mnt/data/plugins/soleur` from an ephemeral container at deploy time (`docker create` + `docker cp src/.` + `docker rm`).
- Sentinel file `.seed-complete` written **after** `docker cp` returns 0 — distinguishes a SIGKILL-mid-cp partial copy from a healthy mount. Boot probe (`verifyPluginMountOnce`) fires `reportSilentFallback` to Sentry on missing path / empty / manifest-missing / partial-seed states.
- Three-file lockstep (Dockerfile, ci-deploy.sh, cloud-init.yml) unified on POSIX `find -mindepth 1 -delete` so bash + dash paths exercise the same cleanup form.

## User-Brand Impact

**Brand-survival threshold:** `none` — capability gap, not a privacy or
data-loss surface. The mount is `:ro` and contains the same public plugin
content already published to the marketplace; no auth/credentials/data/
payments/user-resources surface is crossed.

**If this lands broken:** session loads but every Soleur skill returns
"skill not found" — i.e., the current production behavior for every
web-platform user (months of empty mount). The fix restores the advertised
feature surface.

**If this leaks:** N/A — populating a read-only public-content mount does
not expose user data. The `:ro` flag prevents container write-back.

## Test plan

- [x] Unit: `apps/web-platform/test/plugin-mount-check.test.ts` — 6 vitest scenarios (path missing / empty / manifest missing / partial seed / fully-seeded / memoization) using `vi.resetModules()` + dynamic import (no test-only `_resetForTesting` export).
- [x] Integration: `apps/web-platform/infra/cloud-init-plugin-seed.test.sh` — builds a fixture image, runs the seed sequence verbatim, asserts manifest + skill stub + sentinel land, stale dotfiles removed, manifest mode bits include world-read, `:ro` UID-1001 readability via `docker run --user 1001:1001 -v target:/p:ro busybox`.
- [x] Multi-agent review (8 reviewers + semgrep + test-design + simplicity) — findings fixed inline; one finding (#3053, empty-mount window during deploy) scope-out as `contested-design` with reviewer CONCUR; durable fix deferred to #2608's plugin-rotation API.
- [x] Compound learning recorded: `knowledge-base/project/learnings/2026-04-29-bind-mount-seed-detection-needs-late-sentinel.md` (key insight: late-extracted payload file ≠ post-success sentinel).

Closes #3045
Ref #3053 (deferred: empty-mount window during deploy)
Ref #2608 (plugin freshness rotation API — durable design venue)

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -35,6 +35,10 @@ on:
         description: "Skip path change detection (for workflow_dispatch)"
         type: boolean
         default: false
+      vendor_plugin:
+        description: "Vendor plugins/soleur into <docker_context>/_plugin-vendored before docker build (#3045)"
+        type: boolean
+        default: false
     outputs:
       version:
         description: "Computed version without prefix (e.g., 0.1.1)"
@@ -389,6 +393,19 @@ jobs:
             exit 1
           fi
           echo "::notice::NEXT_PUBLIC_SUPABASE_ANON_KEY passes JWT-claims validation (ref=$ref_safe)"
+
+      - name: Vendor plugin into build context
+        if: steps.version.outputs.next != '' && inputs.docker_image != '' && inputs.vendor_plugin
+        env:
+          DOCKER_CONTEXT: ${{ inputs.docker_context }}
+        run: |
+          set -euo pipefail
+          DEST="${DOCKER_CONTEXT}/_plugin-vendored"
+          rm -rf "$DEST"
+          cp -r plugins/soleur "$DEST"
+          # Drop test/docs trees that aren't needed at runtime — keeps image small.
+          rm -rf "$DEST/test" "$DEST/docs/_site"
+          echo "Vendored $(find "$DEST" -type f | wc -l) plugin files into $DEST"
 
       - name: Build and push Docker image
         id: docker_build
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -400,11 +400,32 @@ jobs:
           DOCKER_CONTEXT: ${{ inputs.docker_context }}
         run: |
           set -euo pipefail
+          # Defense-in-depth: bound DOCKER_CONTEXT to the apps/* prefix so a
+          # future caller passing untrusted input cannot rm -rf an arbitrary
+          # path under "$DOCKER_CONTEXT/_plugin-vendored".
+          if [[ ! "$DOCKER_CONTEXT" =~ ^apps/[a-z0-9-]+$ ]]; then
+            echo "::error::vendor_plugin: DOCKER_CONTEXT='$DOCKER_CONTEXT' must match ^apps/[a-z0-9-]+$"
+            exit 1
+          fi
           DEST="${DOCKER_CONTEXT}/_plugin-vendored"
           rm -rf "$DEST"
-          cp -r plugins/soleur "$DEST"
-          # Drop test/docs trees that aren't needed at runtime — keeps image small.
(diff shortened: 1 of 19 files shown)


deruelle:
## Deploy Pipeline Fix Drift Gate (auto-defer)

This PR modifies `apps/web-platform/infra/ci-deploy.sh` (added the plugin seed block at line ~240). The Deploy Pipeline Fix Drift Gate fires for any change to ci-deploy.sh / cloud-init.yml.

**Headless auto-defer:** Drift verification (canonicalize file+systemd verification contract per #3038) will run on the first deploy after merge. If drift is detected, follow-up issue will be auto-filed.

**Scope:** This PR's ci-deploy.sh changes are additive (new `docker create / cp / rm` block + sentinel write) and do not modify the existing systemd verification contract. Drift expected: none.

PR #3883, opened 2026-05-16
feat(runtime): PR-D attachments-storage tenant RLS (#3244 §4)

deruelle:
## Summary

PR-D is the final slice of runtime tenant-isolation umbrella #3244. Migrates the two callers of `persistAndDownloadAttachments` (`cc-dispatcher.ts:1435` + `agent-runner.ts:2305`) from service-role to tenant-scoped Supabase clients, adds the missing Storage write RLS policies (`storage.objects` INSERT/UPDATE/DELETE on `chat-attachments` + `message_attachments` INSERT), fixes the `attachment-display.tsx` permanent-skeleton silent-failure bug, amends Article 30 register PA2 (closes a pre-existing Art. 30 gap), and shrinks `.service-role-allowlist` from 14 → 13 PERMANENT entries.

Ref #3244 (umbrella stays open until PR-E #3887 lands). Ref #3869 — items 4 (sentinel sweep) and 5 (no-WITH-CHECK ADR) resolved by this PR; operator marks them resolved on the deferrals tracker post-merge (the umbrella issue stays open for items 1–3 and 6).

## User-Brand Impact

- **Brand-survival threshold:** `single-user incident` — carry-forward from PR-C umbrella #3244. Closes the 3rd of 3 named brand-survival vectors (cross-tenant attachment read).
- **Artifact:** Founder B's `chat-attachments` Storage bucket bytes at `${userB.id}/${convB}/*.{png,jpeg,gif,webp,pdf}` (may incidentally contain Art. 9 special-category data per PA2 amendment).
- **Vector closed (1/3):** Cross-tenant Storage SELECT via tenant-JWT. Pre-PR-D, `persistAndDownloadAttachments` ran service-role and would silently return any founder's bytes to any other founder's LLM context. Migration 045's `storage.objects FOR ALL USING ((storage.foldername(name))[1] = auth.uid()::text)` policy now enforces tenant isolation at the database boundary regardless of application-layer state. Test-covered at `test/server/attachment-pipeline.tenant-isolation.test.ts:FR1` (deny + positive control).
- **Vector closed (2/3):** Cross-tenant `message_attachments` INSERT — a tenant client could previously claim another founder's `message_id` and leak filenames / MIME / sizes via the sibling join. Migration 045's INSERT policy joining `messages → conversations.user_id = auth.uid()` rejects this at the DB; tests assert RLS-deny shape NOT FK-violation (which would pass-for-wrong-reason per the 2026-05-16 RLS-deny-payload learning).
[...]

Issue #3244, opened 2026-05-05
feat: Command Center server-side agentic runtime (alignment + hardening)


deruelle:
## Summary

Ship the runtime the roadmap already implies. Three sequenced increments inside `apps/web-platform/`:

1. **Cross-tenant isolation hardening (gate-zero).** Replace `SUPABASE_SERVICE_ROLE_KEY` escape hatch in `server/agent-runner.ts` with per-invocation user-scoped Supabase JWTs; per-invocation BYOK lease + audit row.
2. **Multi-turn continuity (#1044) fix.** Stateful per-(founder, domain, conversation) message history. Prerequisite for background reactions.
3. **Daily Priorities home + one Inngest-driven background trigger.** Dashboard answers 'what needs you today' across 8 domains. Demo: Stripe failed-payment → CFO drafts response.

Launches as **Soleur Command Center** (Web Platform sub-brand). CLI plugin remains 'Soleur'. Closed free preview until E&O + DPA + audit log + scope-grants land. Paid tier $99/mo flat BYOK.

## Reframing from original ask

This brainstorm started as 'Cosmos competitor for solo founders.' CPO and CMO independently challenged that framing:
- The roadmap, brand guide, and Phase 1.10/1.11 already describe this work — alignment, not pivot.
- Real comp is Polsia (\$1.5M ARR), Lindy, Notion AI 3.3 — not Cosmos.
- Cosmos is reference design + competitive context, not direct competitor.

CTO discovered the substrate already exists (`server/agent-runner.ts` runs durable per-user Claude Agent SDK sessions on Hetzner). Pivot scope reduces to: Inngest agent kit + RLS hardening + priorities surface.

## Artifacts

- Brainstorm: \`knowledge-base/project/brainstorms/2026-05-05-command-center-runtime-brainstorm.md\`
- Spec: \`knowledge-base/project/specs/feat-agent-runtime-platform/spec.md\`
- Branch: \`feat-agent-runtime-platform\`
- Draft PR: #3240
- Trigger: Augment Cosmos public preview (https://www.augmentcode.com/blog/cosmos-now-in-public-preview)

## Acceptance criteria (high-level)

- [x] \`server/agent-runner.ts\` no longer uses \`SUPABASE_SERVICE_ROLE_KEY\` for tenant operations; per-invocation user-scoped JWT minted; service-role allowlisted to JWT-mint + audit only. _(PR-A #3240 + PR-B #3395 + PR-C #3854)_
[...]

PR #3911, opened 2026-05-16
fix(runtime): PR-D migration RLS chain — same-tenant INSERT denied (#3883 follow-up)

deruelle:
## Summary

Hotfix for PR-D #3883 — the merged migration 045 contains an RLS policy on
`message_attachments` whose `WITH CHECK` predicate joins through `messages JOIN
conversations` under a tenant JWT. The chained RLS evaluation against
`messages` (whose own SELECT policy from migration 019 contains an EXISTS over
`conversations`) returns zero rows for the **legitimate same-tenant insert**,
denying every founder's attachment upload with 42501.

The cross-tenant DENY test passed in PR-D's CI run — but for the wrong reason:
the predicate denies ALL tenant inserts, not just cross-tenant ones. The
positive-control assertion (added in PR-D as required by AC11) caught the
regression in the same `tenant-integration` CI job, but auto-merge fired before
the fix could be authored.

PR-D's `tenant-integration` job is not yet in the branch-protection
required-checks set (added by PR #3893 to main during PR-D's session window),
which is why the merge proceeded despite the failure. The follow-up to wire
this check as required is a separate concern from this hotfix.

Ref #3244. Ref #3883 — PR-D's auto-merge fired before fix; this branch repairs
the migration before any operator runs `supabase db push` against prod.

## User-Brand Impact

- **Brand-survival threshold:** `single-user incident` — carry-forward from
  PR-D / PR-C umbrella #3244. The broken migration would cause every founder's
  attachment upload to 42501 in production once Phase 8 `supabase db push`
  fires. Pre-deploy ack-gate operator runbook is the only thing preventing the
  prod outage right now; merging this hotfix BEFORE Phase 8 closes the
  exposure window.
- **Artifact:** `message_attachments` row INSERT under tenant JWT — every
  founder hitting the attachment upload path.
- **Vector:** Production-side, not security — the policy is over-restrictive,
  not under-restrictive. Cross-tenant deny still functions (verified by PR-D's
  tenant-integration test); same-tenant denial is the regression.

### Web Platform — Migration 045
- Introduce `public.is_message_owner(p_message_id uuid, p_user_id uuid)`
  SECURITY DEFINER helper, `search_path` pinned to `pg_catalog, pg_temp` per
  `cq-pg-security-definer-search-path-pin-pg-temp`. `STABLE` for planner
  inline-eligibility; `EXECUTE` granted only to the `authenticated` role.
- Rewrite the `message_attachments` INSERT policy to call the helper instead
  of the EXISTS+JOIN predicate. The function runs at superuser privilege so
  the inner JOIN is unaffected by the caller's tenant-JWT RLS chain.
- Header rollback block updated to include `DROP FUNCTION IF EXISTS
  public.is_message_owner(uuid, uuid);` for completeness.

The 019 SELECT policy on `message_attachments` retains its EXISTS+JOIN shape
— it was never hit by a positive-control integration test, so the same
latent bug presumably exists there. A follow-up issue should swap that
policy to use the same SECURITY DEFINER helper; for this hotfix, scope is
limited to the WRITE path (which is what PR-D actually wires into the
tenant-client call sites).

## Test plan

- [x] `tsc --noEmit` clean
- [ ] ⏳ `TENANT_INTEGRATION_TEST=1 vitest run test/server/attachment-pipeline.tenant-isolation.test.ts` — the same-tenant positive control that failed on PR-D's run must pass on this hotfix's run; the cross-tenant deny must continue to pass.
- [ ] ⏳ Operator: do NOT run `doppler run -p soleur -c prd -- bash scripts/run-migrations.sh` for PR-D until this hotfix is also on main.


deruelle:
🔍 **Diagnosis update — CI failure is not the policy logic**

Three consecutive `tenant-integration` runs on this branch failed with the same shape:
> `new row violates row-level security policy for table "message_attachments"`

Same error fired on PR-D #3883 pre-merge (commit `e9991d23`) AND post-merge against main (`144edb12`). Same error fires now on the hotfix's `SECURITY DEFINER + LANGUAGE plpgsql` revision (`2ed2557c`).

**The CI workflow at `.github/workflows/tenant-integration.yml` does NOT apply migrations** — it runs vitest against the dev Supabase project's *current state*. `run-migrations.sh` is operator-invoked per `hr-menu-option-ack-not-prod-write-auth`.

If dev does not have migration 045 applied yet, tenant-client INSERT into `message_attachments` fires the RLS default-deny path (RLS is enabled by migration 019, no INSERT policy exists, deny is default → 42501). That is byte-identical to the error we'd see if the policy were applied and returned `false`.

**Unblock path:**
1. Operator applies migration 045 to dev: `cd apps/web-platform && doppler run -p soleur -c dev -- bash scripts/run-migrations.sh`
2. Re-run the tenant-integration workflow on this PR.
3. If the same-tenant positive control PASSES post-apply → the plpgsql + SECURITY DEFINER policy is correct, hotfix is mergeable.
4. If the positive control still FAILS post-apply → the chained-RLS analysis was wrong and the policy needs further iteration (next attempt: scalar subqueries instead of EXISTS+JOIN; failing that, a separate `user_id` denormalized column on `message_attachments` checked directly).

Ref #3244 / #3883.

Issue #4384, opened 2026-05-24
workflow-gate: promote legal-doc-cross-document-gate to required-check on main ruleset


deruelle:
## Background

PR #4353 (closing #4333) added the DSAR / Art. 17 cascade disclosure for the `workspace_member_removals` WORM ledger (PA-19) to all three canonical legal docs + Eleventy mirrors. That PR exercised the cross-document lockstep against the `legal-doc-cross-document-gate` advisory check — which trivially passes on docs-only diffs (no `surface_pattern` is hit), per session-state §Decisions of feat-one-shot-4333-dsar-legal-doc-lockstep.

The load-bearing verification path for the lockstep was AC10's local bash one-liner (`grep -c 'workspace_member_removals' …` ≥ 4), NOT the CI gate.

## Scope

Promote the `legal-doc-cross-document-gate` advisory job in `.github/workflows/ci.yml` to a required-status-check on the `main` branch ruleset, so a future PR that:

- Adds a new DSAR-surface file (matching the `surface_pattern` regex), AND
- Fails to update the corresponding canonical + Eleventy-mirror legal docs

is hard-blocked at the merge boundary instead of relying on plan-author discipline + multi-agent review to catch the lockstep gap.

## Acceptance criteria

- [ ] `legal-doc-cross-document-gate` is in the required-status-checks list of the `main` branch ruleset (verify via `gh api repos/{owner}/{repo}/rulesets`)
- [ ] A test PR that introduces an artificial `surface_pattern`-matching file without touching `docs/legal/` is blocked from merging
- [ ] The PR closing this issue includes a paragraph in `knowledge-base/legal/article-30-register.md` referencing the gate's promotion

## Out of scope

- Widening the `surface_pattern` regex (separate change)
- Promoting any other advisory gate (e.g., `tc-document-sha-guard` is already a required-check)

## Re-evaluate when

The next DSAR-surface PR (adds a new substrate matching `surface_pattern`) lands. If the operator catches a lockstep gap that the gate should have caught, this issue is overdue.

Source PR: #4353
Source: /ship Phase 6.1 follow-up of feat-one-shot-4333-dsar-legal-doc-lockstep

Issue #4392, opened 2026-05-25
post-merge probes for PR #4385: AC19/AC20/AC21 synthetic probes


deruelle:
## Post-merge synthetic probes from PR #4385

PR #4385 promoted `legal-doc-cross-document-gate.yml`'s `enforce` job to a required-status-check on the `CI Required` ruleset (`5 → 15` transition, verified at apply run 26386154924). Three post-merge synthetic probes remain to fully prove the gate end-to-end. They cannot be automated in a workflow without minting probe PRs as a recurring artifact, so they are operator-triggered.

### AC19 — synthetic deadlock-test PR (gate FAILS on a missing-legal-doc PR)

```bash
git checkout -b test-4384-deadlock-probe
mkdir -p apps/web-platform/app/api/account/export
printf 'test: #4384 deadlock probe — DO NOT MERGE\n' > apps/web-platform/app/api/account/export/_probe-4384.md
git add apps/web-platform/app/api/account/export/_probe-4384.md
git commit -m 'test(4384): probe legal-doc-gate deadlock — DO NOT MERGE'
git push -u origin test-4384-deadlock-probe
gh pr create --title 'test(4384): probe legal-doc-gate deadlock — DO NOT MERGE' \
  --body 'Synthetic probe per #4384 AC19. Closing after verification.' \
  --base main --head test-4384-deadlock-probe --draft
# Wait ~2 min for enforce job to run, then:
gh pr view --json statusCheckRollup \
  | jq -r '.statusCheckRollup[] | select(.name == "enforce") | "\(.conclusion // "PENDING") \(.name)"'
# Expected: FAILURE enforce
gh pr close --delete-branch test-4384-deadlock-probe
```

Expected: `FAILURE enforce`. Cleanup: close the probe PR and delete the branch.

### AC20 — synthetic ack-destroy probe (closes #3915)

Open a throwaway PR that REMOVES the new `enforce` `required_check` block (only) from `infra/github/ruleset-ci-required.tf`. Push. The `apply-github-infra.yml` `terraform plan` step's destroy-guard must REJECT the merge unless `[ack-destroy]` is in the commit message. Close the probe PR without merging once the destroy-guard's rejection is observed in the workflow run.

Then: `gh issue close 3915 --reason completed --comment "<link to probe>"`.

### AC21 — bot-PR happy-path probe (next bot PR via `bot-pr-with-synthetic-checks`)

[...]

PR #4394, opened 2026-05-25
test(4384): probe legal-doc-gate deadlock — DO NOT MERGE

deruelle:
Synthetic probe per #4384 AC19. Edits an export endpoint to verify the legal-doc cross-document gate FAILS when DPA/PP/SLA are not co-edited. Closing after enforce conclusion is observed.


deruelle:
AC19 probe complete. Observed FAILURE COMPLETED enforce — legal-doc cross-document gate correctly rejected a PR that edits an export endpoint without DPA/PP/SLA co-edits. See #4392 for tracking.

PR #4395, opened 2026-05-25
test(4384): probe ack-destroy destroy-guard — DO NOT MERGE

### LINKED ISSUE #3915: follow-through: destroy-guard end-to-end test ([ack-destroy] path) for apply-github-infra
## Follow-Through Item

Destroy-guard end-to-end test: open a PR that removes a required check from `infra/github/ruleset-ci-required.tf`. On merge WITHOUT `[ack-destroy]` on its own line in the merge commit message, expect the `apply-github-infra.yml` workflow to fail at the plan-guard step with the documented error. Then re-test WITH `[ack-destroy]` to verify the acknowledgement path proceeds with the apply.

**Source PR:** #3903
**Created by:** /ship Phase 7 Step 3.5
**Created:** 2026-05-16

## Verification

```yaml
type: manual
manual_because: subjective-design-call
sla_business_days: 30
```

`type: manual` justified: end-to-end destroy-guard validation requires opening, merging, and observing a destructive PR — high coordination cost and not automatable. The guard logic itself is unit-testable (PR #3903's review confirmed the regex anchor + numeric-validation pattern); this issue covers the production-path validation specifically.

## Status

Awaiting operator action.

deruelle:
Synthetic probe per #4392 AC20 / closes #3915.

Removes the Tier 3 `enforce` required_check block from `infra/github/ruleset-ci-required.tf` (without `[ack-destroy]` in the commit). The PR-time `infra-validation.yml` plan job should post a plan comment showing the destroy operation (proxy for the apply-time destroy-guard at `apply-github-infra.yml:243-251`).

The apply-time destroy-guard fires only on `push: branches: [main]` (verified via code-read of `apply-github-infra.yml`), so end-to-end validation would require merging to main. Proxy verification via the PR plan comment was chosen instead per /soleur:go user direction (2026-05-25) to avoid temporarily drifting the live ruleset.

Closing after plan-output confirmation.

### DIFF
--- a/infra/github/ruleset-ci-required.tf
+++ b/infra/github/ruleset-ci-required.tf
@@ -112,20 +112,6 @@ resource "github_repository_ruleset" "ci_required" {
         context        = "tc-document-sha-guard"
         integration_id = var.actions_integration_id
       }
-
-      # --- Tier 3: legal-doc cross-document lockstep gate (#4384, closes the
-      # advisory-bypass-via-auto-merge gap that produced #4333). Context
-      # string is the JOB name (`enforce`) at
-      # .github/workflows/legal-doc-cross-document-gate.yml:36, NOT the
-      # workflow display name — per ADR-032 job-name contract. Workflow
-      # `paths:` filter removed in the same PR (#4384) so the job posts on
-      # every PR; the existing `surface_hit=false` short-circuit (lines
-      # 82-85) keeps non-DSAR PRs at O(seconds). See learning
-      # 2026-03-20-github-required-checks-skip-ci-synthetic-status.md.
-      required_check {
-        context        = "enforce"
-        integration_id = var.actions_integration_id
-      }
     }
   }
 }


deruelle:
AC20 probe complete — but the result REVEALS A GUARD GAP.

Terraform plan output (sticky comment above):
```
Plan: 0 to add, 1 to change, 0 to destroy.
```

Removing a nested `required_check` block from `infra/github/ruleset-ci-required.tf` is classified by Terraform as an in-place UPDATE of the parent `github_repository_ruleset.ci_required`, not as a delete. The destroy-guard at `apply-github-infra.yml:234-235` counts `resource_changes[] | select(.change.actions | index("delete"))` — which is 0 for this change. The `[ack-destroy]` gate would NOT fire on a silent removal of a required-status-check.

This is exactly the class of change #3915 / #4384 wanted to protect against. Reporting back to #4392 and #3915 for follow-up.

PR #4731, opened 2026-06-01
fix(sentry-infra): add 3 untargeted cron monitors to apply allowlist + parity guard

deruelle:
## Summary

`apply-sentry-infra.yml` applies a **hardcoded `-target=` allowlist**, not the whole Terraform module. Three `sentry_cron_monitor` resources existed in `cron-monitors.tf` without a matching `-target` line, so they were **never created in prod** — the apply reported success over a plan that excluded them.

Caught by `/soleur:postmerge` after PR #4714: a live Sentry probe showed `scheduled-content-generator` was still absent (20 monitors, not 21) despite "Apply sentry infra: success". The #4684 output-aware heartbeat is **inert without the monitor** — the producer POSTs `?status=error` to a slug Sentry never knew about, so no issue opens.

## Fix
- Added 3 missing monitors to the workflow `-target` set (now **21 targets = 21 tf resources**):
  - `scheduled_content_generator` (the #4714 regression)
  - `scheduled_bug_fixer` (pre-existing gap)
  - `scheduled_ux_audit` (pre-existing gap)
- Added parity guards **(f)/(f2)** to `function-registry-count.test.ts`:
  - (f) every `sentry_cron_monitor` resource has a `-target` line
  - (f2) no `-target` references a non-existent resource
  - Verified **non-vacuous**: (f) fails when a target is removed.

## Test plan
- [x] `function-registry-count.test.ts` 9/9 pass with fix; (f) fails on revert (anti-vacuous)
- [x] `tsc --noEmit` clean
- [ ] ⏳ Post-merge: `apply-sentry-infra.yml` creates the 3 monitors; verify via Sentry monitors API that `scheduled-content-generator` is present (closes the #4684 alerting gap for real)

Refs #4684

### DIFF
--- a/.github/workflows/apply-sentry-infra.yml
+++ b/.github/workflows/apply-sentry-infra.yml
@@ -202,6 +202,9 @@ jobs:
             -target=sentry_cron_monitor.scheduled_compound_promote \
             -target=sentry_cron_monitor.scheduled_stale_deferred_scope_outs \
             -target=sentry_cron_monitor.scheduled_inngest_cron_watchdog \
+            -target=sentry_cron_monitor.scheduled_bug_fixer \
+            -target=sentry_cron_monitor.scheduled_content_generator \
+            -target=sentry_cron_monitor.scheduled_ux_audit \
             -target=sentry_uptime_monitor.soleur_apex \
             -target=sentry_uptime_monitor.soleur_www \
             -target=sentry_uptime_monitor.soleur_changelog_deep \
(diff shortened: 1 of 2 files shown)


deruelle:
Post-merge verified ✅

The path-filtered auto-apply did **not** fire (this PR changed `apply-sentry-infra.yml` + a test, not `cron-monitors.tf`, which is the watched path). Manually dispatched the workflow (run 26745574017 → success). Confirmed via the Sentry monitors API:

- `scheduled-content-generator` => **active** (closes the #4684 alerting gap for real)
- `scheduled-bug-fixer` => **active**
- `scheduled-ux-audit` => **active**

The new parity guard (f)/(f2) will block any future tf↔workflow target drift.

Issue #4882, opened 2026-06-03
observability: alert on diverged/stale workspace KB clone (KB-sync-stale PIR follow-up)


deruelle:
## Follow-up from KB-sync-stale PIR (PR #4878)

The 2026-06-03 KB-sync-stale incident was **user-reported, not alerted**. The
reconcile failure recorded an `ok:false` row to `kb_sync_history` + mirrored to
Sentry per-attempt, but there was no alert specific to the durable failure
state: "a workspace clone has diverged from `origin/<default>`" OR "a
workspace KB is stale beyond N hours since the last successful sync."

PR #4878 makes the failure **recoverable** (gated self-heal + restored manual
"Sync now" affordance) and **correctly classified** (`non_fast_forward` now has
a producer), but detection still depends on a user noticing a missing file.

### Proposed
Add a liveness/age alert (Sentry rule or scheduled probe) that fires when:
- a `kb_sync_history` row records `error_class = non_fast_forward` with
  `recovered != true` (diverged clone the self-heal could NOT fix — un-pushed
  local commits present), OR
- the newest `kb_sync_history` row for an active workspace is `ok:false` and
  older than N hours.

This closes the detection half of the incident (PR #4878 closed the recovery
half).

**Source:** PR #4878 / `knowledge-base/engineering/ops/post-mortems/kb-sync-stale-no-manual-recovery-postmortem.md`

deruelle:
## Brainstorm complete — scope narrowed to alert-rule-only

**Premise correction:** the "scheduled probe" half of this issue is already shipped. `cron-workspace-sync-health.ts` runs daily (\`23 6 * * *\`) with three arms, all merged **2026-06-01** (two days before this issue was filed):

- Arm 1 — `repo_status='ready'` + `github_installation_id IS NULL` → `op=ready-null-installation`
- **Arm 2 (#4712)** — latest `kb_sync_history` row `ok:false` on an installed workspace → `op=stale-sync-failed`
- Arm 3 (#4717) — went-quiet (default-branch commits never synced) → `op=went-quiet`

Both conditions this issue proposes already emit Sentry events. The bullet-1 case (`non_fast_forward`, `recovered != true`, un-pushed local commits) writes `{ok:false, error_class:'non_fast_forward'}` (verified at `kb-route-helpers.ts:307-309`), and Arm 2 fires on *any* `latest.ok === false` — so it's already caught. Bullet-2 *is* Arm 2.

[...]

PR #4886, opened 2026-06-03
fix(cron): cron-workspace-gc reclaims leaked clones + isolate off persistent KB volume (no-SSH)

deruelle:
## Summary

Restores the org-workspace KB reconcile (no SSH) and isolates ephemeral cron clones off the persistent KB volume — the fix for the 2026-06-02 KB-sync freeze (`Ref #4882`).

**Root cause:** every cron clones the repo via `mkdtemp(.../soleur-<cron>-…)`; cleanup is a caller-side `finally { rm }` that OOM/ENOSPC/SIGKILL bypass. #4770 wired `CRON_WORKSPACE_ROOT=/workspaces` (to dodge the 256 MB `/tmp` tmpfs) onto the **same** 20 GB Hetzner volume that holds the persistent UUID KB workspace clones. Leaked `soleur-*` clones accumulated → volume filled → the reconcile's `git pull` ENOSPC'd → **zero** `kb_sync_history` rows since 07:46 (the error path couldn't even write a row). This is a distinct layer from #4878 (diverged-clone self-heal), which stays.

**The fix (three parts):**
1. **`cron-workspace-gc` Inngest cron** — 6-hourly + on-demand, statfs-reports the cron-clone root to Sentry, sweeps aged `soleur-*` dirs (prefix + maxdepth-1 + age>1h, per-dir fail-soft `rm`), heartbeats Sentry Crons. The disk-reclaim lever, fireable via `/soleur:trigger-cron` with **zero SSH**.
2. **Isolation** — `CRON_WORKSPACE_ROOT=/workspaces/.cron` (both docker-run sites) namespaces cron clones one level below the persistent UUID dirs.
3. **Out-of-band alerting** — `scheduled-workspace-gc` Sentry Crons monitor (missed check-in pages) + a low-disk `warnSilentFallback` — neither depends on the wedged DB-write path that hid the freeze.

Also gold-tints the KB "Sync now" button for UX consistency (color-only).

## User-Brand Impact

**Artifact at risk:** the user's persistent KB-workspace clone (`/workspaces/<uuid>`) and `users.kb_sync_history` freshness.

**Exposure vector:** (1) data loss — the GC's `rm -rf` could delete a user's KB clone if mis-scoped; mitigated structurally (only `soleur-` prefix is ever swept, maxdepth-1, age-gated, run against the isolated `.cron` subdir one level below the UUID dirs, with a load-bearing unit test asserting a 36-char UUID dir is never swept). (2) silent KB freeze — pushes never reach the KB the agent reads, with no error shown; the fix restores the disk-reclaim path AND makes recurrence loud (Sentry heartbeat + low-disk warn, out-of-band of the DB write).

### Web Platform
- Add `cron-workspace-gc` scheduled Inngest function (in-process, ADR-033-compliant) that reclaims leaked ephemeral cron clones and reports free-space deltas to Sentry; fireable on demand via `cron/workspace-gc.manual-trigger` (zero SSH).
- Isolate cron clones onto a dedicated `/workspaces/.cron` subdir at both deploy docker-run sites so a leak cannot starve the persistent KB volume.
- Add the `scheduled-workspace-gc` Sentry Crons monitor (+ apply-workflow target) for missed-check-in paging.
- Exclude `.cron` from the active-workspace metric; extract `freeMb` as the shared single source of truth in `_cron-shared`.
- Gold-tint the KB "Sync now" button for UX consistency.

## Test plan

- [x] `cron-workspace-gc` unit tests (incl. the load-bearing "UUID dir never swept", non-directory skip, ENOSPC/EACCES fail-soft, statfs-after-failure degrade)
- [x] Registry/manifest/allowlist drift guards green (count 48→49); `ci-deploy.test.sh` 79/79 with the `/.cron` literal pinned
- [x] `tsc --noEmit` clean; full `scripts/test-all.sh` green
- [x] Multi-agent review (security/data-integrity/architecture/user-impact/code-quality/test-design) — no P1; destructive sweep confirmed structurally safe
- [ ] ⏳ Post-merge: fire the GC via `/soleur:trigger-cron`, confirm Sentry `scheduled-workspace-gc` heartbeat + `freedMb > 0`, and a fresh `kb_sync_history` `webhook_push ok:true` row

Ref #4882


deruelle:
[deploy_pipeline_fix-drift-gate] This PR touches `ci-deploy.sh` (a trigger file). `apply-deploy-pipeline-fix.yml` will auto-apply on merge — no operator action required. Kill switch: add `[skip-deploy-fix-apply]` to a commit message to defer.

Issue #5092, opened 2026-06-09
Post-merge: widen CF token + verify Bulk Redirects apply (PR #5082, GSC legal redirects)


deruelle:
# Post-merge verification chain for the GSC legal-redirects Bulk Redirects apply (PR #5082)

deferred-automation backlog item; re-evaluate when: (a) a Cloudflare credential with `API Tokens:Edit` (account-owned token management) exists in Doppler, which would make step 1 scriptable; or (b) a GSC service-account credential is provisioned, which would make step 4's read-only verification half scriptable via the URL Inspection API.

PR #5082 lands `cloudflare_list.legal_redirects` + `cloudflare_ruleset.bulk_redirects` (account-level, phase `http_request_redirect`, **12 redirect pairs**) — implementing the refactor tracked in #3367. The `cf_api_token_rulesets` token is **zone-scoped only** — the auto-apply on merge will fail on the two new resources until the token is widened. Verified 2026-06-09: neither `CF_API_TOKEN` nor any other Doppler `prd_terraform` credential can read or edit API tokens (`GET /accounts/{id}/tokens` and `GET /user/tokens` both return 9109 Unauthorized), so step 1 is genuinely operator-only (Cloudflare dashboard, MFA-gated).

Until step 2 completes, `scheduled-terraform-drift.yml` will report the 2 net-new resources as drift every 12h — expected signal, not an incident.

## Checklist

- [ ] **1. (Operator, BLOCKING) Widen the CF token.** Cloudflare dashboard → My Profile → API Tokens (or Account API Tokens) → edit the token backing Doppler `CF_API_TOKEN_RULESETS` → add account-level permissions **`Account Rulesets:Edit`** + **`Account Filter Lists:Edit`** (account: Soleur). If the dashboard forces a roll, update Doppler: `doppler secrets set CF_API_TOKEN_RULESETS -p soleur -c prd_terraform`.
- [ ] **2. Re-fire the apply** (after step 1) and watch it to completion — no dashboard gaze:

  ```bash
  gh workflow run apply-web-platform-infra.yml --ref main -F reason='bulk-redirects apply after token widen'
  sleep 10
  gh run watch "$(gh run list --workflow=apply-web-platform-infra.yml -L1 --json databaseId -q '.[0].databaseId')" --exit-status
  ```

- [ ] **3. Curl verification suite** (after step 2) — runnable as-is; every line must print `301` + the exact expected Location, on apex AND `www.`:

  ```bash
[...]

PR #5116, opened 2026-06-10
fix(playwright-mcp): enable Chromium sandbox via Chrome SUID helper

deruelle:
## Summary

Playwright launches Chromium with `--no-sandbox` by default (`chromiumSandbox: false` is the upstream default), so every Playwright-MCP browser ran unsandboxed — Chrome banners it as "Stability and security will suffer", and an unsandboxed renderer crash takes down the whole browser. This destabilized the long credential-handoff session that stalled the #5092 CF token widen.

This pins in `.claude/playwright-mcp.config.json`:
- `"channel": "chrome"` — use real Chrome, whose root SUID sandbox helper (`/opt/google/chrome/chrome-sandbox`, mode 4755) works even on hosts where AppArmor restricts unprivileged user namespaces (Ubuntu 24.04+, `kernel.apparmor_restrict_unprivileged_userns=1`). Playwright's bundled Chromium cannot sandbox there (its `chrome_sandbox` helper is not setuid), so pinning the channel is load-bearing, not cosmetic.
- `"chromiumSandbox": true` — drop `--no-sandbox`.

## Validation

Applied live in-session 2026-06-10 (MCP server reconnect to pick up config):
- `chrome://version` command line contains **no** `--no-sandbox`; executable is `/opt/google/chrome/chrome`.
- The full #5092 workflow (Cloudflare dashboard token edit + GSC re-validation) ran through the sandboxed browser without a crash.

Ref #5092.


deruelle:
Code review (2 finder agents + verification): no blocking findings.

- `channel: chrome` is **not** a behavior change — @playwright/mcp 0.0.76 already defaults channel to `chrome` when `browserName` is unset (verified in `validateBrowserConfig`); the pin freezes that default against upstream drift, same rationale as the existing `headless: false` pin. Live session pre-fix was already running `/opt/google/chrome/chrome`.
- `chromiumSandbox: true` residual risk is narrow: the config pins headed mode, so display-less containers/CI never load this path (cron-ux-audit writes its own `.mcp.json` without `--config`); only root-with-xvfb would regress, which is not a supported operator setup. An operator who overrides `--browser chromium` on a userns-restricted host would hit a sandbox launch error whose message names the fix.
- Pre-existing (not introduced here): `knowledge-base/project/learnings/2026-05-15-playwright-mcp-headed-and-persistent-profile.md` still describes the pre-`--config` wiring.

Issue #5145, opened 2026-06-11
fix(infra): verify_inngest_health cron-plan budget (30s) shorter than post-restart registry re-sync — healthy deploys record inngest_health_failed


deruelle:
## Problem

`ci-deploy.sh`'s `verify_inngest_health` cron-plan probe (`/v1/functions` must show a `"cron":` trigger) runs with `max_attempts=10, interval=3` (~30s budget). On the 2026-06-11 `deploy inngest v1.1.13` (PR #5131 second-pass quota remediation), the bootstrap succeeded and the registry re-synced — but only after the 30s window, so the deploy recorded `exit_code: 1, reason: inngest_health_failed` in deploy-state.

Observed timeline: deploy window 05:11:56–05:12:37Z; `/health` was live (heartbeat oneshot green); Sentry monitor `cron-egress-resolve` checked in at 05:18:57Z proving the registry synced and crons fire — i.e., a healthy deploy permanently recorded as failed.

Consequence: every inngest deploy whose SDK re-sync exceeds 30s writes a false-negative failure into `/hooks/deploy-status`, which the post-deploy runbook (and any agent reading the status) must then manually disprove via Sentry check-ins — exactly what happened this session.

## Proposed Fix

Widen the cron-plan half of `verify_inngest_health` (e.g. `max_attempts=30` for the `/v1/functions` loop, keeping the fast /health loop as-is), OR have the deploy path retry the verify once after a 60–120s delay before writing the terminal failure state. One file (`apps/web-platform/infra/ci-deploy.sh`), but it is a `deploy_pipeline_fix` trigger file — needs its own PR + the auto-apply on merge.

## Evidence

- deploy-status payload: `{"exit_code":1,"reason":"inngest_health_failed","services":{"inngest_server":"active","vector":"active"}}` while all services active
- `cron-egress-resolve` Sentry check-in 05:18:57Z (post-restart)
- ci-deploy.sh `verify_inngest_health` (~line 201) — `/v1/functions` loop shares the 10×3s budget with /health

Ref PR #5131 (the deploy that surfaced this). Found during the #5110 second-pass remediation session.

deruelle:
Fixed by https://github.com/jikig-ai/soleur/pull/5146 (merged as c2146e7a).

[...]

PR #5146, opened 2026-06-11
fix(infra): give verify_inngest_health's cron-plan loop its own 120s budget; widen restart workflow verify window

deruelle:
## Summary

- `verify_inngest_health()` in `apps/web-platform/infra/ci-deploy.sh` gave both its probes one shared ~30s budget, but inngest-server runs with `--poll-interval 60`: when the immediate post-restart SDK sync races the web-platform container, the registry only populates on the next 60s poll cycle — so healthy deploys recorded `inngest_health_failed` (observed on the 2026-06-11 v1.1.13 deploy, Ref PR #5131). The cron-plan loop now owns `cron_max_attempts=40` (120s nominal, two poll cycles); the `/health` loop stays fast.
- Client side, `.github/workflows/restart-inngest-server.yml` widens its verify window to `MAX_POLLS=140 × POLL_INTERVAL=5` (700s ≥ 400s verify tail + 180s TimeoutStopSec hung-stop + 60s margin) and adds a `start_ts` freshness guard so the first poll cannot honor a stale pre-trigger terminal state.
- `ci-deploy.test.sh` pins the budget statically (exact-value + seq-form + loop-ordering + `--max-time` probes) and adds a cross-file drift guard that extracts every operand by shape from all three source files — including `TimeoutStopSec` from the inngest-server unit heredoc — and asserts the client ≥ server inequality.
- `infra-validation.yml` now triggers on restart-workflow edits so the drift guard fires on client-side changes; the deploy-status runbook gains `inngest_health_failed` and `inngest_restart_failed` taxonomy rows plus corrections of stale 900s ceilings to 1800s (per the raise in PR 5061).

Ref #5145 — issue closure is gated on the post-merge live verification below (the prod fix lands via the auto-apply on merge, not the merge itself).

## User-Brand Impact

- threshold: none, reason: ci-deploy.sh is operator-only deploy plumbing — the diff changes retry-budget integers, log text, and CI poll windows; it never touches user data, auth, or any user-reachable surface; worst case is a slower failure report on an already-failing deploy.

### Web Platform (deploy tooling)

- `verify_inngest_health` cron-plan probe owns a separate 120s budget (two `--poll-interval 60` cycles), eliminating false-negative `inngest_health_failed` on healthy deploys whose registry re-sync exceeds 30s
- Restart workflow verify window widened 150s → 700s with a `start_ts` freshness guard against stale single-slot deploy-state reads
- Cross-file budget drift guard (client window ≥ server verify worst case) in `ci-deploy.test.sh`, enforced on client-side edits via the `infra-validation.yml` paths filter

### Plugin

- Deploy-status runbook: `inngest_health_failed` (with the #5145 slow-resync signature and a no-SSH host-delivery procedure) and `inngest_restart_failed` taxonomy rows; stale 900s wrapper/release ceilings corrected to 1800s
- Work skill: drift guards must extract every numeric operand by shape (learning from this PR's review cycle)

## Test plan

- [x] `bash apps/web-platform/infra/ci-deploy.test.sh` → 83/83 (two new static assertions; RED verified before GREEN)
- [x] `bash scripts/test-all.sh` full suite green
- [x] `actionlint` clean on both edited workflows (incl. resolving the pre-existing SC2034 warning in the rewritten block)
- [x] AC1–AC9 grep battery from the plan
- [x] Live discoverability probe: deploy-status endpoint returns the canonical JSON envelope (verified during preflight)
- [ ] Post-merge (automated, this session): `apply-deploy-pipeline-fix.yml` run green for the merge SHA AND infra-config-status per-file sha256 for `/usr/local/bin/ci-deploy.sh` matches the merged file (AC11)
- [ ] Post-merge (automated, this session): re-dispatch `restart-inngest-server.yml` after delivery is verified → green run proves the widened budget live; then close the issue with run URLs (AC12)

Note: the merge push itself dispatches `restart-inngest-server.yml` (its push trigger is path-scoped to itself), racing the auto-apply against the OLD host script — that first run may legitimately fail and is not a regression signal; the AC12 re-dispatch is the authoritative green.


deruelle:
[deploy_pipeline_fix-drift-gate] This PR touches a trigger file (apps/web-platform/infra/ci-deploy.sh). `apply-deploy-pipeline-fix.yml` will auto-apply on merge — no action required. To skip the auto-apply, add `[skip-deploy-fix-apply]` to a commit message. If the auto-apply fails (transient outage), run the workflow manually from the Actions tab, or as a last resort: `doppler run -p soleur -c prd_terraform -- terraform apply -target=terraform_data.deploy_pipeline_fix -input=true`.

PR #5445, opened 2026-06-16
fix(security): clear unpatched Dependabot alerts via overrides (ws/protobufjs/js-yaml/@babel/core) + markdown-it

deruelle:
## What

Closes the Dependabot alerts that had **no auto-PR path** because the package is either pinned by an existing override or is transitive:

| Pkg | Sev | Where | Why no Dependabot PR | Fix |
|-----|-----|-------|----------------------|-----|
| `ws` | HIGH | web-platform (direct) | pinned by `overrides.ws` (stale #4035 closed) | dep + override → `^8.21.0` |
| `js-yaml` | MED ×2 | root + web-platform | transitive | override → `^4.2.0` |
| `@babel/core` | LOW | web-platform | transitive | override → `^7.29.6` (resolves 7.29.7) |
| `markdown-it` | MED | root (direct) | Dependabot #5431 left frozen `bun.lock` unsynced → CI fail | bump `^14.2.0` + sync lockfile |

## Why this PR (vs Dependabot)

- The `ws` HIGH alert can't be fixed by Dependabot because `apps/web-platform/package.json` `overrides` pins `ws` — Dependabot's earlier #4035 was closed for this reason. Bumping the override is the only path.
- `js-yaml` / `@babel/core` are transitive; only an `overrides` entry forces the patched version.
- **Supersedes Dependabot #5431** (markdown-it): that PR updated `package-lock.json` but not the repo's frozen `bun.lock`, so `test-bun` failed with `lockfile had changes, but lockfile is frozen`. This PR syncs both lockfiles.

## Verification

- `bun install --frozen-lockfile` passes in root **and** web-platform (the CI gate that failed #5431).
- `tsc --noEmit` (web-platform typecheck) clean.
- All target versions are past the 3-day `minimumReleaseAge` cooldown, so CI's `bun install` will resolve them.
- Both `bun.lock` and `package-lock.json` updated in each package so Dependabot stops re-flagging.


deruelle:
Folded in **protobufjs** (HIGH+MED) after Dependabot #5381 self-closed as "no longer updatable" on rebase — same override-required constraint as ws. This PR now clears 7 alerts: ws, protobufjs×2, js-yaml×2, @babel/core, markdown-it.

PR #5653, opened 2026-06-25
perf(dashboard): unblock cold-start render from /api/kb/tree

### LINKED ISSUE #5654: perf(dashboard): unblock cold-start render from /api/kb/tree
## Summary

On cold start (notably after a computer restart) the dashboard shows a full-screen skeleton for several seconds. The whole page is gated behind `GET /api/kb/tree` (the `kbLoading` early-return at `dashboard/page.tsx:421`), but that data only drives foundation-card completion checkmarks — not the primary conversation list. The fetch is uncached (service worker skips `/api/*`) and sits behind a serial per-request auth waterfall in `middleware.ts` (getUser → revocation RPC → users SELECT), plus a token refresh after restart.

**Fix:** decouple the render from `/api/kb/tree`. Gate the page on the conversation-list load instead; let foundation-card checkmarks fill in async. Highest perceived-perf win, lowest risk, single-file change.

## Key safety nuance

While the KB tree is still loading, `visionExists` is unknown — the render must NOT flash the first-run "Tell your organization what you're building" empty state to an existing user.

## Artifacts

- Brainstorm: `knowledge-base/project/brainstorms/2026-06-25-dashboard-cold-start-render-brainstorm.md`
- Spec: `knowledge-base/project/specs/feat-dashboard-cold-start-render/spec.md`
- Branch: `feat-dashboard-cold-start-render`
- Draft PR: #5653

## Acceptance criteria

- Conversation list paints without waiting for `/api/kb/tree`.
- Existing users never flash the first-run empty state during the KB-tree load window.
- Foundation-card checkmarks appear once the tree resolves; no layout shift.
- Provisioning (503), 401-redirect, and conversation-error states unchanged.

## Deferred (separate work)

- Persistent client cache (revisits ADR-067 in-memory-only / CPO C1).
- Speeding up the middleware auth waterfall (touches the revocation gate).

Elvalio:
## Problem

On cold start — notably after a computer restart — the dashboard shows a full-screen loading skeleton for several seconds before any content appears.

Root cause: the entire page is gated behind `GET /api/kb/tree` (the `kbLoading` early-return in `dashboard/page.tsx`). That data only drives **foundation-card completion checkmarks**, but it's the slowest thing on a cold start:
- **Uncached** — the service worker explicitly skips `/api/*` (`public/sw.js`), so it always hits the network.
- Behind the **per-request middleware auth waterfall** (`getUser` → `check_my_revocation` RPC → `users` SELECT) plus a **disk tree-walk** (`buildTree`).
- After a restart, the first authenticated request also pays a **Supabase token refresh**.

The JS bundle itself is already service-worker-cached, so "store it in cache" was half the story — the real block was the per-user data waterfall the skeleton needlessly waited on.

## Fix

Gate the dashboard on the **conversation-list load** (the primary content), not the KB tree. The tree now resolves async; foundation-card checkmarks fill in when it arrives.

- Users **with conversations** never wait on `/api/kb/tree` — the inbox paints as soon as conversations load.
- **No-content users** still hold the skeleton briefly while the tree resolves, so first-run vs. empty is chosen without flashing the wrong state.
- A filter-triggered refetch keeps prior rows → the inbox's own row-level skeleton handles it (no full-page blank).
- The 401 → `/login` redirect hold is preserved.

Single-file render-flow change (`app/(dashboard)/dashboard/page.tsx`). No middleware, server, caching, or service-worker changes.

## Test plan

- [x] `tsc --noEmit` clean
- Existing user w/ conversations: inbox paints without waiting on the KB tree; checkmarks fill in after.
- Existing user w/ vision, no conversations: never flashes first-run.
- Brand-new user: first-run shown correctly once tree resolves (no empty-state flash).
- Filter change: row-level skeleton only, no full-page blank.
- 401 mid-fetch: holds through `/login` redirect.

## Deferred (separate work)

- Persistent client cache (revisits ADR-067 in-memory-only / CPO C1).
- Speeding up the middleware auth waterfall (touches the security-sensitive revocation gate).

Closes #5654


deruelle:
## Review verdict: do not merge on the current premise

Reviewed as part of an operator-confirmed open-PR drain. I resolved this branch against `origin/main` first (see the merge commit), then ran a multi-agent review. **The premise no longer holds.**

### The bottleneck this PR removes was already removed on main

The PR gates first paint off `/api/kb/tree` because that path ran a full recursive `buildTree()` walk. But `main`'s `dashboard/page.tsx` no longer fetches it — it fetches `/api/dashboard/foundation-status`, a targeted stat. Verified directly:

```
$ git show origin/main:'apps/web-platform/app/(dashboard)/dashboard/page.tsx' | grep -n 'api/kb/tree\|foundation-status'
60:// set of KB paths, so it fetches /api/dashboard/foundation-status (a targeted
61:// stat) instead of the whole-KB-tree walk (/api/kb/tree buildTree()) that used
66:  "/api/dashboard/foundation-status",
```

The spec, the brainstorm, and this PR's new in-code comment ("gated on the CONVERSATION-LIST load, NOT on /api/kb/tree") all argue against a fetch this file stopped making.

### Consequence: the change is a net regression

- **Cold start slower.** The replaced gate was 1 client round trip (`foundation-status`). The new clause (a) gates on `useConversations`, which is 3–4 serial round trips: `auth.getUser()` → `/api/workspace/active-repo` (also `/api/*`, also behind the same middleware waterfall) → `list_conversations_enriched`. The new gate *contains* the old one plus two extra trips.
- **Warm start regresses.** `useConversations` is `useState`/`useEffect`, not SWR — `loading` starts `true` and `conversations` starts `[]` on every mount, so clause (a) fires on every remount. `foundationData` was SWR-cached, so `kbLoading` was already `false` on a warm remount. This file's own comment asserts the invariant the change breaks: *"a warm remount keeps it defined, so the skeleton never re-shows"*.
- **Full-page skeleton on a zero-match filter.** Clause (a) implements "loading with no rows", not "first load". When the previous filter matched zero conversations, it blanks the page — intercepting the row-level skeleton and the "No conversations match your filters" state. That is precisely the scenario the comment promises won't blank the page.
- **FR3/AC3 unimplemented.** The spec requires a placeholder so foundation cards don't shift layout. No placeholder ships; `FoundationSection` is absent while loading and inserts above the conversation list on resolve (~380–450px desktop, ~560px mobile).
- **"The KB tree only drives foundation-card completion checkmarks" is false.** `foundationPaths` also feeds `visionExists`, which gates the entire first-run screen, whether `FoundationSection` mounts, the container layout class, and the h1 copy.

### What I did land

The merge resolution itself was independently audited and confirmed correct — including dropping the `emailItems.length === 0` clause from `showInitialSkeleton` (main's feat-inbox-attention-badge removed that surface; **porting the clause would have been the actual bug**, making clause (c) stronger than the gates it protects). `npx tsc --noEmit` exits 0.

### Recommendation

Re-baseline against what `main` does today and measure before/after. If the remaining cost is the middleware auth waterfall, that is this brainstorm's own deferred non-goal and a render-flow change cannot reach it. Leaving open rather than closing — the merge resolution is worth keeping if the work is revived.

*No findings filed as issues: the blocking finding is the PR's own premise, not a separable defect.*

deruelle:
Closing per the review verdict above.

The premise no longer holds: `main` already stopped fetching `/api/kb/tree` — it uses `/api/dashboard/foundation-status`, a targeted stat — so the whole-tree walk this PR exists to unblock was eliminated by a different change. As written the diff is a net regression on both cold start (the new gate is 3–4 serial round trips where the old one was 1) and warm start (`useConversations` is uncached, so every remount now shows a full-page skeleton).

Not filing a follow-up issue against this branch. Per the review, the remaining cold-start cost is the middleware auth waterfall — a different bottleneck that would not reuse this diff, and one the originating brainstorm already lists as a deferred non-goal. A future effort there should start from a fresh measurement against current `main`, not from this branch.

The conflict resolution against `main` (including correctly dropping the `emailItems` clause) is preserved in this branch's history if it is ever wanted.

Closed as part of an operator-confirmed open-PR drain (`/soleur:drain-prs`).

Issue #6060, opened 2026-07-05
web-2-recreate verify: deferred hardening (graceful web-1 swap / web-2 health probe / merge-freeze) — #6051 follow-ups


deruelle:
## Summary

Deferred hardening from the #6051/#6040 fix (web-2-recreate fan-out verify bounded degraded-retry + warm_standby dedup). The retry bounds the web-1 re-swap amplification and both dispatch paths now share `apps/web-platform/infra/scripts/deploy-status-fanout-verify.sh`, but three follow-on items were deliberately scoped out. Grouped into one tracker per the net-issue-flow discipline (all are deferred hardening, not discovered bugs). Ref #6051, Ref #6040 (run 28747333763).

## Deferred items

- [ ] **(a) Truly zero-downtime web-1 re-swap OR a web-2-only fan-out path (secondary finding).**
  The recreate verify's only off-host trigger is `/hooks/deploy` → web-1 → fan-out, so it re-swaps
  web-1 (the sole live origin) at its current tag. This PR *bounds* the amplification to ≤2 swaps
  (`DEGRADED_RETRY_MAX=1` + `FRESH_BOOT_WINDOW_S` gate) but does not make the swap perfectly graceful —
  run 28747333763 showed a transient single-probe `521` on `app.soleur.ai/health` that recovered in
  ~3s. Making the container swap truly zero-downtime is a server-side `ci-deploy.sh` change; skipping
  web-1's re-swap entirely needs a new off-host web-2-only webhook capability (`/hooks/deploy-peer` is
  private-net-deny). BetterStack's apex/origin probe already catches a *sustained* (not transient) 521.

- [ ] **(b) Private-net web-2 post-accept health probe for the GA cutover (user-impact FINDING 2).**
  The off-host verify proves web-2 ACCEPTED the fan-out (HTTP 202 + `:9000` bound) via `reason==ok`,
  NOT web-2's own post-accept canary/health (`cosign_verify_failed`, `canary_sandbox_failed`, etc.
  are soak-verified only over the PRIVATE net — the off-host CI runner is private-net-deny by the
  ADR-068 topology). web-2 is weight-0 / drained throughout, so an accepted-but-unhealthy web-2 serves
  no live traffic — but before the GA cutover routes live traffic to web-2, add a private-net health
  probe (or a GA-cutover step) so `reason==ok` is backed by web-2's own canary verdict.

[...]

PR #6082, opened 2026-07-06
fix(infra): tolerate web-1's :latest tag in the off-host fan-out verify baseline

deruelle:
## Summary

The off-host web-2 verify (used by `web_2_recreate` + `warm_standby`) reads web-1's deployed tag from `/hooks/deploy-status`, then re-deploys web-1 at that tag to trigger the deploy fan-out to web-2. The baseline **required a pinned `^v[0-9]` tag and aborted on anything else** — a deliberate guard (test AC4-tag) against redeploying a floating tag to the live origin. But web-1 currently reports `tag: "latest"` (the durable-`:latest` state), so the verify **aborts at the baseline and web-2 can never be confirmed** — the blocker behind the last several verify attempts.

**Change:** tolerate `latest` in addition to `^v[0-9]` (`^(v[0-9][A-Za-z0-9._-]*|latest)$`). Truly-invalid tags (empty/garbage) still abort (new `garbage-tag` fixture guards this; new `ok-latest-s300` fixture proves `latest` reaches a successful verify). Verify suite **27/27**.

## User-Brand Impact

- **Artifact / vector:** `hcloud_server.web["web-1"]` (sole live origin). The verify re-deploys web-1 at `latest` to trigger the fan-out.
- **Mitigation:** `latest` is the **newest** tag, so the re-swap can **never downgrade** web-1 (the exact risk the original `^v[0-9]`-only guard addressed). The verify only redeploys the **current GA image** — the same operation any release performs — and ci-deploy's canary (`/login`+`/dashboard`+`data-error-boundary` rejection + rollback) still protects web-1 from a bad image. The tag-downgrade race guard is unchanged (it only adopts a fresher *versioned* tag). Worst case vs. status quo: web-1 gets re-deployed the current GA it's already running (idempotent) instead of the verify aborting.
- **Brand-survival threshold:** `single-user incident` — the verify touches the live origin, so kept conservative.

### Web Platform (infra)
- The off-host web-2 fan-out verify now tolerates web-1's `:latest` tag at the baseline (was aborting), unblocking web-2 warm-standby confirmation. Truly-invalid tags still abort.

## Test plan

- [x] `deploy-status-fanout-verify.test.sh` 27/27 (new AC4-latest tolerated→exit 0; AC4-tag garbage→exit 1); shellcheck clean
- [x] architecture review of the web-1 re-swap safety (latest cannot downgrade; canary/rollback intact)
- [ ] Post-merge (this session): re-run `warm-standby` — the verify should now pass the baseline and confirm web-2's `:9000` (or surface a real web-2 issue instead of the tag abort).

Notes the cleaner long-term fixes in-code: a web-2-only fan-out that never re-swaps web-1 (Ref #6060), and resolving the durable-`:latest` at the deploy source.


deruelle:
Holding as draft — architecture review found this is **non-functional** (P1): `ci-deploy.sh:924` hard-rejects any non-`vX.Y.Z` tag as `tag_malformed`, so tolerating `latest` at the verify baseline is inconsistent with the deploy path — POSTing `deploy … latest` is rejected and the verify still fails (the passing test mocked the POST, a false positive). The real fix must resolve `latest` → a pinned GA `vX.Y.Z` before POSTing, or route through the web-2-only fan-out (#6060). Re-planning.

Issue #6416, opened 2026-07-14
infra: web-2 missing private-net IP → zot mirror push unreachable (ADR-096 zot-primary dead end-to-end)


deruelle:
## web-2 has no private-network IP → CF tunnel origin can't reach zot → CI zot dual-push mirror silently skips

**Found during #6400 recovery.** `soleur-web-2` (warm standby) has **no private IP** (Hetzner API `private_net` empty; `privIP=-`). The CF tunnel that serves `registry.soleur.ai` / `ssh.soleur.ai` currently routes to **web-2's cloudflared** — but web-2 is not on `10.0.1.0/24`, so it cannot reach the zot registry at `10.0.1.30:5000`. Result: the CI `Bridge to zot registry` step's true `outcome` is `failure` (masked green by `continue-on-error`), so `Mirror image GHCR→zot (crane)` **skips** on every release → **zot never receives new images** → deploys fall through to GHCR (the ADR-096 zot-primary path is effectively dead end-to-end even though the zot host itself is now healthy).

**Ask:**
1. Restore web-2's private-network attachment (it should be a `10.0.1.0/24` member like web-1) — likely a `web-2-recreate` / server_network fix.
2. Verify the CF tunnel ingress for `registry.*`/`ssh.*` is served by a private-net-attached origin (web-1), or that both web hosts are private-net members so any tunnel origin can reach `10.0.1.30`.
3. Consider failing the release loud (not `continue-on-error`-masked) when the zot mirror can't reach the registry, so "zot silently not backfilled" self-reports.

**Verification:** from the CF-SSH bridge, `hostname` returns `soleur-web-2` and `nc -z 10.0.1.30 5000` → UNREACHABLE; from web-1 (`10.0.1.10`, direct) it is REACHABLE.

Ref #6400

---

## Follow-through enrollment (added by the #6416 fix PR)

This issue stays **open** until zot is provably being backfilled again. It is not closed at
merge: the failure is statistical (one tunnel, N connector replicas), so a single green
release proves nothing. The sweeper closes this issue automatically once **≥5 consecutive**
releases run the zot mirror with no degraded signal — no human revisit needed.

**Measured baseline at fix time (16 most recent completed release runs that built an image):**

| bridge step `conclusion` | mirror step | count |
|---|---|---|
| `success` (masked by `continue-on-error`) | **skipped** | **15** |
| `success` | `success` | **1** |

The bridge's real error, from the job log:
[...]

Issue #5933, opened 2026-07-03
chore(infra): fresh web-2 boot observability prerequisites — per-host absence detector, A-record drain, egress enforcement probe, image digest pin (blocks #5887)


deruelle:
> **Status update — 2026-07-03 (brainstorm scoping):**
> - **Item 3 (egress enforcement probe) is DONE** — shipped in PR #5945 (`cron-egress-enforce-probe.sh`; `server.tf` cites "#5933 item 3"). Checked below.
> - **Blocker citation corrected:** this work is NOT blocked by #5887. **#5887 is a CLOSED Terraform `moved`-block CI fix**, not a provisioning cutover. The real web-2 provisioning cutover is **#5274 Phase 3.D** (OPEN), per `dns.tf:4`.
> - **Item 1 is UNBLOCKED.** ADR-082 deferred Item 1 only because `apply-web-platform-infra.yml` was RED (#5887); that pipeline is now **green** (post-#5887). web-1's per-host coverage is shippable now (`monitored`-gated; web-2 stays `monitored=false` until the #5274 cutover).
> - **Design is ADR-locked** in [ADR-082](knowledge-base/engineering/architecture/decisions/ADR-082-fresh-web2-boot-observability.md). Delivery: **two PRs, Item 1 first**. Item 4 verify extended to **both** the fresh-host (`cloud-init.yml`) and running-host (`ci-deploy.sh`) paths — an ADR-082 amendment.
> - Brainstorm: `knowledge-base/project/brainstorms/2026-07-03-web2-absence-detector-image-pin-brainstorm.md` · Spec: `knowledge-base/project/specs/feat-web2-absence-detector-image-pin/spec.md` · Branch: `feat-web2-absence-detector-image-pin` · Draft PR: #5970

## Problem

The fresh Hetzner **web-2** boot surface (introduced by #5921's bake-and-extract cloud-init path) is a **blind execution surface** — no SSH, CI can't reach the host. #5921 ships the SSH-free *error-path* signals (a discriminating Sentry event `{stage, failed_file, image_ref, host_id}` from the extraction/bootstrap failure traps) and the fail-closed guarantee (a failed extraction aborts runcmd before the app `docker run`, so the app never serves). BUT the **PRIMARY absence detector the design leans on does not yet exist as code**, and observability review (gating) surfaced concrete gaps that only bite once web-2 is actually provisioned (behind the #5274 Phase 3.D operator cutover). Filing them as a hard prerequisite of that cutover.

[...]

PR #6421, opened 2026-07-15
fix(infra): guard per-PR host creation + un-mask the silently-skipped zot mirror (ADR-114)

deruelle:
## What this fixes, and what it deliberately does not

**Ref #6416** — deliberately `Ref`, not `Closes`. **The user-facing harm is still live at merge.** This PR does not attach web-2 to the private network; that is a post-merge `warm-standby` dispatch. Until it runs and ≥5 consecutive releases mirror cleanly, zot stays un-backfilled and deploys keep falling back to GHCR. #6416 is enrolled in the follow-through sweeper (`earliest=2026-07-22`) and closes on evidence, not at merge.

### The operator's question, answered

> *"Why do we have a tunnel for each backend instead of a tunnel that allows to reach any backend?"*

**The premise is inverted.** We do not have a tunnel per backend — we have exactly **one** (measured: 1 `cloudflare_zero_trust_tunnel_cloudflared`, 3 ingress rules; the CF API reports 2 connectors / 8 QUIC conns). One tunnel is also *correct*. What is per-backend is the (hostname + CF Access app + service token + CI bridge) quadruple — deliberate least-privilege, which this PR keeps.

The real defect is narrower: **`localhost:` in a multi-replica tunnel is a category error.** It does not mean "this host"; it means "whichever replica answered". CF binds ingress to a *tunnel*, then load-balances across replicas — the contract, not a bug. So a per-hostname ingress **does not pin a connector**, and any "route to a specific host" fix that doesn't change the ingress *service* address changes nothing. Recorded as **ADR-114** (normative: I1 connector homogeneity, I2 origin-relative ingress).

### Measured

| bridge step `conclusion` | mirror step | count |
|---|---|---|
| `success` (forced by `continue-on-error`) | **skipped** | **15** |
| `success` | `success` | **1** |

≈94% failure across the 16 most recent completed releases that built an image. The bridge's real error: `Get "http://127.0.0.1:5000/v2/": context deadline exceeded` — the connector that answered had no route to `10.0.1.30:5000`.

Every layer that could have reported this was silent **by construction**: `continue-on-error` forces the bridge's `conclusion` to `success`; the mirror's `if: steps.zot_bridge.outcome == 'success'` then skipped it; a skipped step emits nothing, so `mirror_status` stayed unset and the Slack degraded line (which reads that output *by step id*) stayed inert.

> **Correction recorded, not smoothed over.** An earlier revision of this PR claimed "14/14, therefore the route is **pinned**". A later run succeeded, falsifying pinning outright. Review caught it by re-running a measurement I had asserted from a snapshot. Neither ~50% nor 100% is right; the skew's mechanism is unexplained and is an open question on #6441.

## Changes

| Phase | Change |
|---|---|
| **P2** | `host_creates` — a 7th destroy-guard counter + a HALT on the per-PR apply. A host birth was invisible to all three existing counters (no delete, no nested-block shrinkage, not an `["update"]`), which is how web-2 was born unattached. Evaluated **outside** the `destroy_count` sum, so `[ack-destroy]` cannot reach it. Per-type remediation names the right dispatch per host, plus `[skip-web-platform-apply]` as the unwedge. |
| **P3** | The zot mirror is gated on `docker_build` (not `zot_bridge`) and branches on the bridge **internally**, so a failed bridge reaches the same `degraded()` emitter and the same output id the Slack line reads. ADR-096 is **not** reversed — `continue-on-error` stays on both steps; the ask is satisfied by loudness, not blocking. |
| **P4** | **ADR-114**; ADR-008 → `superseded-in-part` (its claimed `app.soleur.ai → localhost:3000` route does not exist — measured 0); ADR-068 **extended** (it already stated the multi-connector fact and already rejected per-host tunnels; its "11 SSH provisioners" corrected to a measured 12). ADR-096 vindicated, unamended. C4 updated. |
| **P5** | Two dead `-target`s dropped (`cloudflare_record.web_host`, `betteruptime_monitor.web_host` — both resources retired with #5933). Soak probe enrolled. |
| **~~P2c~~** | **CUT at review** — see below. |

### Reviewer note: the T18 change is not a weakened test

T18 asserted a per-PR `hcloud_server.web["web-2"]` create passes the guard and named it *"a legit new host"*. Its `rupd=0` assertion is retained and still true — a create genuinely is not a reboot. What is withdrawn is the **conclusion** that it was therefore legitimate. That belief is the thing this PR overturns; T29 now owns the legitimacy call.

## Cut at review: the wrong-host tripwire (P2c)

Five agents converged, so it is deleted rather than patched (the plan's own rule, which already cut v1's P2b):

1. **Not fail-closed** — Terraform dials a new SSH session per provisioner block, so the assert certified a *different* session than the writes. It halved wrong-host writes and **certified** the survivors: worse than the silence it replaced.
2. **Keyed on a rejected primitive** — `ADR-082:196-199` rejected runtime `$(hostname)` for this exact host pair; `server.tf:236` calls it "not guaranteed"; **web-1's value was never measured** (the issue measured web-2's).
3. **Its guard didn't check what it claimed** — mutation-proven: moving it below the `file` provisioner stayed green.
4. **It couldn't run on this merge** (no `triggers_replace` input changed) — first execution would land on an unrelated future PR and taint 12 resources.

Moved to **#6441** with the better design review surfaced (same-session assert, TF-authoritative instance-id).

## Plan

`knowledge-base/project/plans/2026-07-15-fix-web2-private-net-attach-and-tunnel-connector-homogeneity-plan.md` — carries the `## User-Brand Impact` section (threshold: `single-user incident`), the `## Observability` block, and the v3 Implementation Record of every deviation and correction.

## Review

10 agents. Fixed inline: **4 P1** — the `host_creates` replace fail-open (a `-replace` births a host, read 0, and the destroy gate then *prints* "add `[ack-destroy]`" → #6416 reproducing **through** the guard); the HALT being **untested** (deleting it entirely left 33/33 green); the soak probe able to **PASS on the mirror step's absence**; and the tripwire (cut). Plus the falsified rate claim, a third exposed `deploy.` leg, an ADR-096 misquote, and the ~50% figure that six files had copied from a model the ADR refutes.

**Net issue flow: 0 closed / 4 filed (#6440 P1 audit, #6441 I2, #6442 unfirewalled first boot, #6443 drowned drift detector).** Each is a discovered defect in a different subsystem or the ADR's named follow-up — the gate forbids burying those in a shared tracker. #6416 itself closes via the soak.

## Post-merge (agent-run — not operator)

1. Dispatch `apply_target=warm-standby` (it already targets the missing `hcloud_server_network.web["web-2"]`). On RED, read the **attach-proof step first** — a red verify with a green apply means the attach landed; do **not** reflexively `web-2-recreate`, which would destroy a good attach.
2. Prove packets, not state: `private_net[0].ip` → `10.0.1.11` **and** ≥5 consecutive clean registry bridges.
3. The sweeper auto-resolves issue #6416 once the probe passes.

## Changelog

- Guard the per-PR terraform apply against birthing an unattached host (`host_creates`, no `[ack-destroy]` bypass).
- Un-mask the silently-skipped zot mirror: a failed registry bridge now reports degraded instead of skipping into silence.
- ADR-114: one tunnel, many connectors — tunnel ingress must be origin-relative.


deruelle:
## Red check `validate (apps/web-platform/infra)` is pre-existing — not from this PR

**Reproduced against `origin/main`:**

```
$ git show origin/main:apps/web-platform/infra/cloud-init.yml > /tmp/main-cloud-init.yml
$ cloud-init schema -c /tmp/main-cloud-init.yml
Error: Invalid schema: user-data
```

Identical failure on code already on main. This PR does not touch `cloud-init.yml` (0 hits in the diff).

**Why it surfaced here:** `infra-validation.yml` is path-filtered on `apps/web-platform/infra/**`. Its last run on `main` was 2026-06-11 (green). The `%{ if web_colocate_inngest ~}` directive that breaks the raw-YAML parse landed later in #6344, and no `main` push has touched the filtered path since — so the check never re-ran there. This PR's comment-only edit to `tunnel.tf` is the first to trigger it.

The step validates the **raw** Terraform `templatefile()`, which cannot parse as YAML by construction. The repo already documents the fix (`work/SKILL.md:622` — render via `terraform console`, then validate). Tracked in #6450.

Not required for merge, and not caused here — flagging so it is not misattributed to this PR.

Issue #6285, opened 2026-07-09
observability(infra): upgrade zot mirror-staleness alarm to sentry_metric_alert (aggregate) once a resolvable notify target exists


deruelle:
## Upgrade the zot mirror-staleness alarm from `sentry_issue_alert` to `sentry_metric_alert`

Follow-up from #6278 (PR provisioning the live fallback-rate alarm).

### Background

#6278 shipped `sentry_issue_alert.zot_mirror_fallback_rate` (`apps/web-platform/infra/sentry/issue-alerts.tf`) — an `event_frequency` count > 3/1h over `filter_match="any"` across the four runtime fallback/degrade signals. The `soleur:engineering:cto` agent ruled **against** the more sensitive `sentry_metric_alert` (true cross-signal aggregate) for that PR because:

- `sentry_metric_alert.trigger.action` requires a **concrete numeric `target_identifier`** (team-id or member-id) — there is **no** `IssueOwners → ActiveMembers` symbolic fallthrough for metric-alert actions.
- This Sentry Terraform root has **zero** resolvable numeric notify targets (no `sentry_team` data source, no `owner=`, no member variable; all 21 issue alerts use the symbolic fallthrough).
- The token to look one up (`SENTRY_IAC_AUTH_TOKEN`) is **CI-only** (ADR-031, not in Doppler), so a numeric target is unresolvable/unverifiable in an autonomous session → shipping a metric alert would risk **paging nobody**.

Full rationale + rejected alternatives: `knowledge-base/project/specs/feat-one-shot-6278-zot-mirror-staleness-alarm/decision-challenges.md`.

### Why upgrade later

The shipped issue-alert thresholds **per fingerprinted Sentry issue-group**, not on a true cross-signal aggregate. It reliably pages on the dominant shared-tag correlated outage (a rolling-deploy zot miss concentrates in one `ghcr-fallback (web:<tag>)` group), but a **fully-distributed thin spread** (< 3 fallbacks in every signal-group simultaneously) would not page. A `sentry_metric_alert` with `aggregate="count()"` over the OR-of-signals query would page on the aggregate, closing that tail. (See also the per-signal grouping note in the resource comment: `inngest_ghcr_fallback` shares the shared `soleur-boot-emit` message-group, so it is over-loud today — a metric alert would normalize all four signals.)

### Re-evaluate when (dependency trigger)

[...]

Issue #6122, opened 2026-07-06
feat(supply-chain): migrate container registry off GHCR to self-hosted zot (Hetzner, R2-backed, control-plane-minted OIDC)


deruelle:
## Migrate container registry off GHCR to self-hosted zot (Hetzner, R2-backed)

**Why:** A GitHub App installation token **cannot** `docker pull` private repo-linked GHCR packages — confirmed GitHub platform limitation ([community #171423](https://github.com/orgs/community/discussions/171423), GitHub staff: "GHCR does not yet accept GitHub App installation tokens for authentication"), no fix/ETA. GHCR therefore cannot deliver a zero-touch machine identity (the only pulling credentials are a browser-only user PAT or the workflow-only `GITHUB_TOKEN`). This resolves the #6073 escalation question (answer: no — migrate, don't wait for GitHub) and supersedes the #6031 minter approach.

**Decision:** Self-hosted [zot](https://zotregistry.dev) registry in-datacenter on Hetzner, Cloudflare R2 storage backend, zot's native OIDC bearer-token workload auth validating a **control-plane-minted JWT**. Reuses the existing (disabled) Inngest minter — swap the token endpoint. cosign chain preserved (signing identity is GitHub Actions OIDC, registry-agnostic).

**Why zot over alternatives:** A booting Hetzner VM has no native OIDC issuer, so pure OIDC-federation registries (ECR/GAR/CF-OIDC) would force building an IdP — OIDC's benefit collapses into the control-plane-mint pattern we already run. zot validates a minted bearer natively (no bespoke auth code, unlike an R2+Worker registry). In-datacenter placement satisfies the restricted-egress firewall (ADR-052/#5046) with zero egress. $0 new vendor.

**Artifacts:**
- Brainstorm: `knowledge-base/project/brainstorms/2026-07-06-registry-oidc-migration-brainstorm.md`
- Spec: `knowledge-base/project/specs/feat-registry-oidc-migration/spec.md`
- Branch: `feat-registry-oidc-migration` · Draft PR: #6120

**Acceptance criteria:** host pulls via short-lived minted credential (no user PAT) · minter re-pointed to zot · cosign verify passes from zot · firewall/egress satisfied · zero-gap cutover (interim PAT stays as break-glass until zot HA proven) · exposed PAT rotated.

**Top open risk:** zot boot-time SPOF — mitigate with HA / pull-through cache / PAT break-glass (plan-time).

[...]

PR #6424, opened 2026-07-15
fix(observability): zot mirror-fallback alarm threshold can never fire

deruelle:
## Summary

`sentry_issue_alert.zot_mirror_fallback_rate` **has never been able to fire.** This changes one integer — `event_frequency value = 3` → `value = 0` — and corrects the false comment that produced the unreachable threshold.

Ref #6285

### Why it could never fire (proven at source, not inferred)

`event_frequency` counts the whole Sentry **issue-group's** events — `filters_v2` gate whether an event *evaluates* the rule, they do not scope the count. And `ci-deploy.sh:607` embeds the unique deploy tag in the **message**, so Sentry mints a **fresh issue-group per deploy**; the per-group count is bounded by fleet size, not a rate.

Sentry OSS `src/sentry/rules/conditions/event_frequency.py`:

```python
# :174 — new-group short-circuit
# Assumes that the first event in a group will always be below the threshold.
if state.is_new and value > 1:
    return False
return current_value > value     # :195 — STRICT >
```

| `value` | New-group short-circuit | Fresh per-deploy group | Verdict |
|---|---|---|---|
| **3** (shipped) | `is_new and 3 > 1` → **returns False** | later: `2 > 3` false | **never fires** |
| 1 | proceeds | `1 > 1` false | fires only on the 2nd host |
| **0** | proceeds | `1 > 0` **true** | **fires on first event** ✅ |

`value = 0` is accepted server-side: `EventFrequencyForm.value` declares **no `min_value`** (`:93`), while the sibling `EventFrequencyPercentForm` declares one at `:854` — the omission is deliberate. It also matches `zot-soak-6122.sh`, which FAILs the Phase-5 gate on **>=1** fallback.

#6285 asked for a `sentry_metric_alert`. Rejected — at its own `alert_threshold = 3` it would be **equally dead**. The integer was the bug.

## User-Brand Impact

- **If this lands broken, the user experiences:** nothing directly — which is why the miss is invisible. The fleet keeps deploying successfully off the GHCR fallback; the first signal is the 7-day soak sweeper failing at day 7, resetting the Phase-5 cutover by a week.
- **If this lands *working but muted*, the user experiences a full deploy/recovery outage:** all four signals share one rule (`filter_match = "any"`). Muting the **rule** to escape `zot-gate-degraded` noise also kills `registry:ghcr-fallback` — the only no-SSH page gating the **irreversible** ADR-096 5.5 PAT rotate+revoke. Mitigated here: the `.tf` comment and the revert runbook both now say **mute the ISSUE, never the rule**, and explain why that is safe by construction.
- **If this leaks, the user's data is exposed via:** N/A — both emitters carry fixed literals + `image_kind` + a git tag + a 3-literal reason. No user content, no credential, no PII. `value` changes **notification**, not **capture**.
- **Brand-survival threshold:** `single-user incident`

### Web Platform (infra)
- `sentry_issue_alert.zot_mirror_fallback_rate`: `event_frequency value` 3 → 0 (fire-on-first). The alarm can now page on a zot→GHCR mirror fallback; previously it could not fire on its primary signal at any fleet size below four pullers.
- Replaced the comment block's false justification (*"a rolling-deploy zot miss drives many hosts onto the SAME group"* — false in both halves) with the **invariant** (no host count, so it cannot rot as the fleet changes), a contrast against `web_terminal_boot_fatal`'s `value = 1`, a change-trigger, and a **mute-the-issue-not-the-rule** note.
- Added retirement tripwires at all **three** ADR-096 task-5.3 sites (`ci-deploy.sh`, and both `cloud-init.yml` fallback branches) — comments only, no behavior.

### Docs
- `ADR-096`: corrected the alarm's window — it **opens at task 1.8** (`ZOT_REGISTRY_URL` set) for 3 of 4 signals, **not** at the `ZOT_ACTIVE=1` flip; **closes at 5.3** for 3 of 4 (`zot-gate-degraded` survives, so 5.3 must **narrow** `filters_v2`, never retire). Also corrected the `betteruptime_heartbeat.registry_prd` zot-liveness claim — that heartbeat has **zero consumers** and ships `paused = true`.
- `zot-registry-revert.md`: the alarm's only triage doc told the operator to **dismiss** the exact page this fix creates. Corrected to fire-on-first; removed the false "arm it at cutover" instruction.

### Plugin
- `review/SKILL.md`, `brainstorm/SKILL.md`: one bounded bullet each (route-to-definition from `/compound`).

## Test plan

- [x] `vitest run test/sentry-zot-mirror-fallback-alert-op-contract.test.ts` — **5 pass**. Genuine RED→GREEN: the suite *passed by pinning `value = 3`*, so flipping it to assert `value = 0` failed against the un-fixed `.tf`.
- [x] `ci-deploy.test.sh` — **137 PASS / 0 FAIL** (comment-only change).
- [x] `terraform fmt -check` + `terraform validate` — clean. The `sentry_issue_alert` deprecation warning is **expected** (ADR-031 NG9 forbids `sentry_alert` until provider GA).
- [x] No new resource type → the 4 destroy-guard artifacts untouched; an in-place attribute change is `["update"]` (element count 6→6), so no `[ack-destroy]`.
- [x] `-target=sentry_issue_alert.zot_mirror_fallback_rate` present at `apply-sentry-infra.yml:265` — the merge-apply will actually apply this.

Post-apply verification runs automatically (read-only Sentry API + a synthetic `zot-gate-degraded` probe that is invisible to `zot-soak-6122.sh`); #6285 closes only after both pass.

## Review

4 agents (security-sentinel, architecture-strategist, user-impact-reviewer, observability-coverage-reviewer). **14 findings, all 14 fixed inline, 0 scoped out.** They caught this PR committing its own failure mode — the first tripwire I wrote claimed *"darkens 3 of the 4 signals … retire that alarm"* when it darkens **1 of 4**, and retiring would have blinded 3 live signals.

## Related

Filed, not fixed here: **#6435** (`zot-soak-6122.sh` is blind to 2 of the alarm's 4 signals → can false-PASS the irreversible GHCR-retirement gate — **P1, higher value than this fix**) · **#6436** (Sentry unmodeled in C4) · **#6437** (ci-deploy Sentry-dark paths) · **#6429** (sibling `sandbox_startup_failure` may share the defect) · **#6445** (ship trailer-gate false-positives) · **#6446** (`validate`'s cloud-init schema step red on every infra PR since #6344, merging anyway) · **#6447** (`ghcr-minter-doppler-token.tf:53` cites `cloud-init.yml:554`; real target `:408`) · **#6427** retargeted onto the soak-retirement remainder.

### CI fix after first merge attempt (`6fb8dec27`)

The first merge poll exited on a required-check failure. Three checks were red; the diagnosis is worth recording because two of them were unrelated to each other:

| Check | Required? | Verdict |
| --- | --- | --- |
| `test` | **yes** | **Mine.** Shard aggregator — failed solely because `test-bun` did. |
| `test-bun` | no | **Mine.** See below. |
| `validate (apps/web-platform/infra)` | no | **Pre-existing** → #6446. Red on every infra PR since #6344 (2026-07-11); reproduced on an unmodified `origin/main` checkout. |

**The regression:** this PR's `cloud-init.yml` tripwires were described as *"comment-only, zero risk"* — verified only for template injection (no `${` / `%{`), never for size. `cloud-init.yml` renders into base64gzip'd Hetzner `user_data` under a sub-cap ratchet (`WEB_GZIP_BUDGET = 21_800`, `HETZNER_CAP = 32_768`); the 11 comment lines landed at **22,104 B — 272 B over**. Comments in that file ship over the wire to every host.

**The fix:** compressed to two one-line pointers into `ci-deploy.sh`'s `RETIREMENT TRIPWIRE`, which already carries the full enumeration for free (a CI script, not `user_data`) and names both cloud-init sites by *emit name* rather than line number. Lands at **~21,720 B with the budget untouched** — the ratchet's own rationale ends *"Comments trimmed first"*, so trimming to fit is the fix, not re-baselining.

**Also re-anchored line refs this PR's own diff staled** (its Learning B, third instance inside the PR that exists to fix comment rot): ADR-096's `cloud-init.yml:536`/`:695` + `ci-deploy.sh:857` → emit-name anchors (`:536` was already wrong — the branch is at `:535`); the plan's `ci-deploy.sh:1049`/`:1056` → the FATAL literals (my +14 lines moved `:1049`→`:1063`, and `:1056` was never a FATAL — it's `tmpenv=$(mktemp ...)`).

Green after the fix: **bun 2131/2131**, **vitest 5/5**, `terraform fmt` clean.

Plan: `knowledge-base/project/plans/2026-07-15-fix-zot-fallback-alarm-unreachable-threshold-plan.md`

---

**Soak-followthrough gate override (operator-approved).** #6285 has **no soak-gated close criterion** — it closes on AC10 (live rule reads `value: 0`) + AC11 (synthetic live-fire), both immediate post-apply. The 7-day soak this body references is **#6122's** Phase-5 gate, already enrolled via `scripts/followthroughs/zot-soak-6122.sh`. The gate keys on the prose token `soak` rather than on whether *this* tracker has a time-gated close — same false-positive class as #6445.

---

**Preflight Check 10 override (operator-approved).** Check 10 FAILs this PR's `discoverability_test.command` at Step 10.5 (`contains shell-active token; refusing to run`) because the probe carries `$SENTRY_IAC_AUTH_TOKEN` and a `| jq`. The probe is correct — SSH-free, read-only, and the canonical way to verify this change (it is AC10). The gate's row-7 SKIP exists for exactly this "auth-gated probe" state but is **unreachable**: it keys on `$DT_RC == 22`, which is assigned only after the Step 10.5 `exit 1`. Row 7's own suggested remediation ("add a Doppler-fetched probe variant") is rejected by the same regex on `$(`. So an auth-gated observability surface can only ever FAIL this check.

Per `2026-04-27-preflight-security-gates-skip-vs-fail-defaults.md`, the correct disposition is SKIP (invariant untested), not FAIL (invariant contradicted) — the gate's stated invariant is not contradicted here. Filed as **#6445** (Defect 2) with a proposed fix that keeps the blocklist intact and re-classifies the credential-only rejection. Not fixed inline: a semantic change to a gate in another skill is out-of-scope per `review/SKILL.md` §5.


deruelle:
### Preflight: 7 PASS / 3 SKIP / 1 FAIL — Check 10 overridden as a gate bug (operator-approved)



**Check 10 (Discoverability Test Execution) FAILs on a correct command.** It executes `discoverability_test.command` under `env -i` (all secrets scrubbed) and rejects any `$VAR` at Step 10.5 *before* running. Reading a Sentry alert rule requires `-H "Authorization: Bearer $SENTRY_IAC_AUTH_TOKEN"` — there is no unauthenticated equivalent. The check's own decision matrix **row 7** says an auth-gated probe should **SKIP**, but the Step 10.5 reject fires first, so row 7 is unreachable dead code. No authenticated probe is executable by this check, by design.

Filed as **#6445** (alongside the ship trailer-parse gate, same class).

**The probe was deliberately not weakened to pass.** Swapping in an unauthenticated `curl https://sentry.io/api/0/ → 401` would turn the gate green while proving the vendor is reachable and nothing about whether the alarm is armed — a proxy, not the invariant. That is the exact defect class this PR exists to fix, and the same one `kieran-rails-reviewer` caught in this PR's own plan (`terraform validate` offered as evidence one bullet after the `rules/preview` probe was discredited for the identical fallacy).

The command is correct and was **verified live** during this session — it is how the deployed `conditions[0].value = 3` was confirmed.

| Check | Result |
|---|---|
| Not-Bare-Repo | PASS |
| DB Migration Status | SKIP (no migrations) |
| Security Headers | PASS (4/4 critical live) |
| Lockfile Consistency | SKIP |
| Environment Isolation | PASS (dev ≠ prd) |
| Prod Bundle Supabase Host | SKIP |
| Brand-Survival Self-Review | PASS |
| Canary Probe Set | PASS |
| SW Cache Bump | SKIP |
| Node-Only Encodings | PASS |
| Discoverability Test | **FAIL → overridden (#6445)** |

Issue #6448, opened 2026-07-15
review: docker-daemon.json insecure-registries hardcodes 10.0.1.30:5000 — drift from local.registry_private_ip fails SILENT (self-referential probe)


deruelle:
Discovered during the multi-agent review of #6422 (private-NIC convergence, #6415) by `pattern-recognition-specialist`, verified independently by me and co-signed by `code-simplicity-reviewer`. **Pre-existing on `main`** — not introduced by #6422 (that PR *removed* a hardcoded copy).

## The defect

`local.registry_private_ip` (`apps/web-platform/infra/zot-registry.tf:40`) is the single source for the registry's private IP, and `local.registry_endpoint = "${local.registry_private_ip}:5000"` (`:44`) already derives from it — feeding `tunnel.tf:66` and the `ZOT_REGISTRY_URL` output (`:189`).

But the **`:5000`-suffixed endpoint is independently hardcoded** in three live places, with no drift guard:

| Site | Content |
| --- | --- |
| `apps/web-platform/infra/docker-daemon.json:7` | `"insecure-registries": ["10.0.1.30:5000"]` |
| `apps/web-platform/infra/cloud-init.yml:445` | same literal, **web-host** cloud-init |
| `apps/web-platform/infra/server.tf:608` | `docker info … \| grep -q '10.0.1.30:5000'` |
| `apps/web-platform/infra/registry-insecure-config.test.sh:63,70` | asserts `.30` in both of the above |

**The probe that looks like it guards this is self-referential.** `server.tf:571` is `sha256(file("${path.module}/docker-daemon.json"))` and `:589` is `provisioner "file" { source = "${path.module}/docker-daemon.json" }` — a **static file copy, not a `templatefile`**. So `:608` greps the literal `.30:5000` against the `docker-daemon.json` it just delivered, which hardcodes that same literal. It validates the file against a copy of its own content and **cannot detect drift from the local by construction**.

## Why it matters — silent, and invisible from the deploy pipeline

Set `local.registry_private_ip = "10.0.1.31"` (a subnet renumber, or a region move like #6288):

1. `network.tf` attaches at `.31`; cloud-init bakes `EXPECTED_IP=.31` → the #6415 NIC guard is healthy.
2. `local.registry_endpoint` → `.31:5000` → web hosts + CI pull from `.31:5000`.
[...]

Issue #6446, opened 2026-07-15
validate (apps/web-platform/infra): cloud-init schema step red on every infra PR since #6344 — validates the raw templatefile, and isn't a required check


deruelle:
## Summary

`validate (apps/web-platform/infra)` → the **Validate cloud-init schema** step has been **failing on every infra PR since #6344 merged (2026-07-11)**, and each of those PRs merged anyway because `validate` is **not a required status check**. The gate is dark: it cannot catch a real cloud-init schema regression today.

## Evidence

```
PR #6344 → validate (apps/web-platform/infra): failure   <- the PR that introduced the break
PR #6401 → validate (apps/web-platform/infra): failure
PR #6412 → validate (apps/web-platform/infra): failure
```

Reproduced locally against an **unmodified `origin/main`** checkout:

```
$ cloud-init schema -c main-ci.yml
expected alphabetic or numeric character, but found '{'
  in "<unicode string>", line 664, column 1:
    %{ if web_colocate_inngest ~}
     ^
Error: Invalid schema: user-data
```

## Root cause

`.github/workflows/infra-validation.yml:83-90` runs `cloud-init schema -c cloud-init.yml` against the **raw file**, with no render step:

```yaml
- name: Validate cloud-init schema
  run: |
    if [[ -f cloud-init.yml ]]; then
      sudo apt-get install -y -qq cloud-init 2>&1 | tail -1
      cloud-init schema -c cloud-init.yml
```

But `apps/web-platform/infra/cloud-init.yml` is a Terraform `templatefile()` source, not YAML. It survived for years because `${...}` interpolations sit inside values and parse as ordinary scalars. #6344 (`feat(infra): gate co-located inngest bootstrap behind web_colocate_inngest`) introduced the first **column-1** `%{ if ... ~}` directive — and `%` at the start of a YAML line is the reserved **directive indicator** (`%YAML`, `%TAG`), so the parser aborts before cloud-init ever sees it.

## Why it went unnoticed

`infra-validation.yml` triggers on `pull_request` only — it never runs on main pushes, so main's run history (last entry 2026-06-11) is stale `workflow_dispatch` results and looks green. Combined with `validate` not being a required check, a red result is invisible unless someone opens the run.

## Suggested fix

[...]

PR #6458, opened 2026-07-15
fix(ci): render cloud-init templatefiles before schema-checking them

### LINKED ISSUE #6454: ci: infra-validation 'validate' schema-checks the RAW cloud-init templatefile — red on every infra PR since #6344
`infra-validation.yml`'s `validate (apps/web-platform/infra)` job is **red on every PR that touches `apps/*/infra/**`**, and has been since #6344 merged. It is invisible on `main` because the workflow is `pull_request`-only, so nobody sees it unless they are shipping infra.

Discovered while shipping #6422 (private-NIC convergence) — that PR failed the check on a file it never touched.

## Reproduce (against `main`, no PR needed)

```
git show origin/main:apps/web-platform/infra/cloud-init.yml > /tmp/main-ci.yml
cloud-init schema -c /tmp/main-ci.yml; echo "EXIT=$?"
```

```
Error: Cloud config schema errors: format-l665.c1: File /tmp/main-ci.yml is not valid YAML.
while scanning a directive
  in "<unicode string>", line 665, column 1:
    %{ if web_colocate_inngest ~}
EXIT=1
```

## Root cause

The job schema-validates a **Terraform template as if it were a rendered file**:

```yaml
# .github/workflows/infra-validation.yml
- name: Validate cloud-init schema
  run: |
    if [[ -f cloud-init.yml ]]; then
      cloud-init schema -c cloud-init.yml      # <-- the RAW template
    fi
  working-directory: ${{ matrix.directory }}
```

`apps/web-platform/infra/cloud-init.yml` is consumed via `templatefile(...)`, so it legitimately contains Terraform directives. `%{ if web_colocate_inngest ~}` (added by **#6344**, `5fbf00f0e`) is a *template directive*, not YAML — the YAML scanner reads a leading `%` as a **directive indicator** and hard-fails. Every `${...}` interpolation is the same class of problem.

This is a known, already-documented trap. `plugins/soleur/skills/work/SKILL.md` §Infrastructure Validation says it outright:

> When the `cloud-init.yml` is a Terraform `templatefile()`, `cloud-init schema` on the RAW file **always** fails on the un-rendered `${...}` — validate the RENDERED output instead.

The workflow never got that memo. `cloud-init-registry.yml` and `cloud-init-inngest.yml` happen to pass only because they contain no `%{` and their `${...}` occurrences don't start a line.

## Why it matters (beyond noise)

A **permanently-red gate trains operators to ignore it** — the exact review-fatigue class. Right now the honest reading of `validate (apps/web-platform/infra)` is "always red, ignore," which means the day it goes red for a *real* reason (an actual malformed cloud-init, a broken `terraform validate`) nobody will look. The job also runs `terraform fmt -check` and `terraform validate` in the same matrix entry — **those are real gates and they are currently sharing a red light with a broken one.**

It is non-required, so it does not block merge — which is precisely why it has survived. #6422 merged with it red.

## Fix

Render before validating, mirroring what `/work` already prescribes and what `private-nic-guard.test.sh` does:

- [ ] In the `Validate cloud-init schema` step, detect a templatefile (`grep -qE '\$\{|%\{' cloud-init.yml`) and, when found, render it via `terraform console` + `templatefile(...)` with a stub var map, strip the `<<EOT … EOT` wrapper, then `cloud-init schema -c` the **rendered** output. Validate the raw file directly only when it carries no template syntax.
- [ ] Alternatively (cheaper, weaker): skip the schema step for templatefile-consumed cloud-inits and rely on each host's own `*.test.sh` render step. This loses coverage — prefer the render.
- [ ] Add a guard so the job cannot silently pass by skipping: assert it validated *something* (rendered or raw) for each `cloud-init*.yml` present.

**Acceptance:** `validate (apps/web-platform/infra)` is green on a PR that touches `apps/*/infra/**` with `main`'s current `cloud-init.yml`, AND a deliberately-malformed rendered cloud-init still fails it (prove the gate is not merely disabled).

## Scope-Out Justification

**Criterion: `pre-existing-unrelated`.** Verified mechanically — #6422's diff does not contain `apps/web-platform/infra/cloud-init.yml`; the failing line entered via #6344 (`5fbf00f0e`) and reproduces against `origin/main` with no PR involved. #6422 only *surfaced* it (its auto-sync pulled main's file in, and the workflow is PR-only so main never showed it).

Filed **standalone** rather than folded into #6438 per the review skill's rule that a **discovered defect in a different subsystem** must stay its own issue: this is CI-gate correctness for the **web-host** cloud-init, not private-NIC convergence.

**Re-eval by (dependency):** blocked until any PR touches `apps/web-platform/infra/cloud-init.yml` or `.github/workflows/infra-validation.yml` — whichever lands first should carry the fix, because until then every infra PR ships with a red gate the author must manually reason past.

Ref #6344 · surfaced by #6422 · context #6415

deruelle:
🐛 Closes #6454

Ref #6446 — the same bug, filed 69 minutes earlier and broader. Its second half ("isn't a required check") is out of scope here and tracked in #6480, so this PR deliberately does not auto-close it.

## The bug

`infra-validation.yml`'s `validate` job ran `cloud-init schema -c cloud-init.yml` against a **raw Terraform templatefile**. `apps/web-platform/infra/cloud-init.yml` carries `%{ if web_colocate_inngest ~}` at column 1 (introduced by #6344); YAML reads a leading `%` as a directive indicator and the scanner hard-fails before schema-checking begins. Red on **every** PR touching `apps/*/infra/**` for four days, invisible on `main` because the workflow is `pull_request`-only, and non-required — so every one of those PRs merged anyway. A permanently-red gate trains operators to ignore it, and it shared that red light with `terraform fmt` and `terraform validate`, which are real gates.

**Two failure classes, one root cause.** The step was also guarded by `if [[ -f cloud-init.yml ]]`, so it validated **1 of 5** templates. Three cloud-inits and `hooks.json.tmpl` were never validated at all.

## The fix: render, never skip

The issue's cheaper alternative — skip the schema step for templatefile-consumed cloud-inits — degrades the gate to a no-op (all four are templatefile-consumed). A `%{`-strip was rejected on executed evidence too: it emits a document matching **neither** directive arm, never validates the **default** production config (`web_colocate_inngest` defaults `false`), and cannot parse JSON templates.

**Discovery is structural, never a filename allowlist:** the union of (A) every `templatefile()` referent parsed from the dir's `*.tf` and (B) every `cloud-init*.yml`. A validator that only knows files by name is the same defect class as the bug it fixes. This covers #6448's `docker-daemon.json` the day it becomes a templatefile, with **zero** edits here (proven by simulation, fixture F9).

Terraform rules on grammar: keys come from the `.tf` call site, types from the body, then `terraform console` renders and its message is surfaced verbatim. The type derivation is still a guess — documented in the script header, with every narrowing failing loud rather than mis-validating.

**Verified on the real corpus:** 5/5 rendered+validated, both arms of `cloud-init.yml` (795/731) pass schema, `hooks.json.tmpl` validated for the first time. Credential-free: `terraform console` runs in an empty scratch dir with no init, no providers, no backend, every var stubbed `"x"`.

## What review changed (the fix had the bug's own defect class in it)

| Found | Fix |
| --- | --- |
| **Required check RED, gate never proven.** The fixture suite auto-globbed into `guard-script-fixture-tests` — REQUIRED, merge_group, path-filter-free — which is a bare-bash runner with no terraform/cloud-init. All 22 fixtures exited 6; the suite had never once run the render path in CI, and the check blocked every PR in the repo. | Renamed `fixtures-*` to leave the glob; runs from `deploy-script-tests` (has terraform). Adding apt to the required job was rejected: it would put a package-mirror dependency on the merge-queue critical path for every PR. |
| **Silent false-green.** The fmt-clean *wrapped* `templatefile(` call style yielded ZERO referents from a line-based grep — `hooks.json.tmpl` silently left the corpus, gate reported a happy `4/4`. The N/N counter is blind to this: both sides derive from the same discovery pass. | Line-independent discovery + attribution, plus an **independent** completeness floor (templatefile( call count vs parsed referents). |
| **Six false-reds on correct files** — one-line map, nested `})` at col 0, brace inside a string value, negated `%{ if !x }`, compound `%{ if a && b }`, left-strip `%{~ if`. Each reds a valid `.tf`, which *is* the #6454 dynamic. | Brace-depth + string-aware awk state machine; bool typing harvests the whole condition with a comparison carve-out. Each pinned by a fixture. |
| **A false claim that would have wedged the repo.** The aggregator comment said the ruleset flip was "a one-line admin follow-up" because `if: always()` always reports. It is JOB-level; this workflow is path-filtered with no `merge_group`, so the context would never post on a non-infra PR. The precedent I cited *dropped its path filter for exactly this reason*. | Comment rewritten; real scope tracked in **#6480**. |
| **My "inert" comment was not inert.** The `cloud-init.yml` note reddened `cloud-init-user-data-size.test.ts` (22,076 vs a 21,800 budget). I'd proved it inert by diffing the *rendered* arms and never considered **size** — that file is baked into a 32,768-byte-capped `user_data` with <300 bytes of headroom. | Moved to `server.tf`'s call site: no user_data cost, better home. `cloud-init.yml` is **untouched**. |
| exit 3 survived deletion with the suite green; F6a passed via a *different* exit-4 branch than it named; two `assert_not_out` checks were unfalsifiable. | Pinned by message, not just exit code. A removed guard also made the SUT *hang* (empty filename → awk reads stdin) — harness now runs under `timeout`. |

**41/41 fixtures** (was 30), real corpus 5/5, `run-all.sh` green with terraform absent, all three matrix roots green, shellcheck + actionlint clean, size budget green.

## Verification

- `validate (apps/web-platform/infra)`: **ran** (matrix non-empty via the `server.tf` touch) and **passed**, log carries `rendered+validated 5/5 file(s)`. A skipped-but-green job would be a FAIL of this task.
- `infra-validate-required`: green.
- Sibling suites unaffected: `registry-insecure-config`, `private-nic-guard`, `soleur-host-bootstrap-observability`, `cloud-init-inngest-bootstrap` (51/51).

## Known scope boundaries

- **`infra-validate-required` is advisory today** — #6480 makes it real. Gate correctness being advisory is the same property that let this bug live four days; stated plainly rather than papered over.
- **Coverage narrowing:** two bool arms cover 2 of 2^N. Complete today (N=1, verified). The script **emits the bool count** each run, so N≥2 surfaces in the log rather than being discovered by a miss.
- **Collision:** draft PR #6456 fixes the same bug differently inside an unrelated observability batch, editing the same file. A CTO review ruled for this design; see #6446 for the full disposition.


deruelle:
## `secret-scan-allow-rename` — attestation record

Applied on operator instruction to clear `rename-guard (allowlist destinations)`. Recording the evidence so the override is auditable rather than a bare checkbox.

**What tripped it** — the archive commit (`6815a16bf`) renames two KB artifacts:

| from | to |
|---|---|
| `knowledge-base/project/plans/2026-07-15-…-plan.md` | `knowledge-base/project/plans/archive/20260715-164601-…-plan.md` |
| `knowledge-base/project/specs/feat-one-shot-6454-…/session-state.md` | `knowledge-base/project/specs/archive/20260715-164601-…/session-state.md` |

**Why this is outside the guard's threat model.** `rename-guard` exists because gitleaks v8.24.2 evaluates path allowlists against the rename **destination** and does not re-scan content against the source path — so `git mv <scanned>/with-secret.ts <allowlisted>/now-allowed.ts` launders a secret past the gate. That attack requires the **source** to be scanned.

Here the allowlist regex is `knowledge-base/(?:plans|project/(?:plans|specs))/.*\.md$`, and **both** source and destination match it. Scan coverage is therefore unchanged by the rename: these paths were allowlisted before and after. This is an allowlisted→allowlisted move, not scanned→allowlisted.

**Content verified independently** (because allowlisted paths are never scanned in *either* location, the guard's firing is the only prompt to look — so we looked):

- `gitleaks dir` over both files with the allowlist **disabled** (no `.gitleaks.toml`): `scanned ~62.59 KB — no leaks found`, exit 0.
- Manual sweep for `sk-…`, `ghp_`, `github_pat_`, `AKIA…`, `-----BEGIN`, `eyJ…` (JWT), `dp.pt.…` (Doppler), and `password|token` assignments: no matches.
- Both files are our own prose (481 + 19 lines): the #6454 plan and its session-state.

**Not claimed:** that the guard is wrong in general. It is correct and stays. A narrowing — don't fire when the source path is *already* allowlisted, since coverage cannot change — looks worth doing but is out of scope here and would need its own smoke case in the `rename-guard-*` matrix.

<sub>Refs #6454. Override path documented in `knowledge-base/engineering/operations/secret-scanning.md` §Rename-laundering.</sub>

Issue #6497, opened 2026-07-15
P1: zot/GHCR docker-login gate cannot name its own failure — class=unclassified is >=4 modes (WEB-PLATFORM-5B)


deruelle:
**Sentry: `WEB-PLATFORM-5B`** — *"zot gate degraded (login_failed) — configured but inactive, using GHCR"*. 14 events, first `2026-07-14T21:11:28Z`, last `2026-07-15T15:10:07Z`, 100% `zot_gate_reason=login_failed`.

## This is not a regression. zot has never worked.

zot has served **zero pulls in 90 days** (Sentry, 90d: zero `registry:zot` AND zero `registry:ghcr-fallback` issues). And it is **not** credential drift: tfstate `random_password.zot_pull.result` == Doppler `soleur/prd` == Doppler `soleur-registry/prd` (sha256 `922cf93d…`, len 40); push tokens likewise agree. All three planes agree. Nothing degraded — the capability never existed.

## Root cause (structural, not inferred)

`/etc/zot/htpasswd` is baked **exactly once**, at boot, by `cloud-init-registry.yml`'s `runcmd`. `hcloud_server.registry`'s `templatefile()` passes only the two non-secret usernames and **zero** references to `random_password.*.result` — the token values are deliberately kept out of `user_data` (correct, and documented as such). `grep -n replace_triggered_by apps/web-platform/infra/zot-registry.tf` → **zero hits**.

So Terraform has **no data edge** from the password to the host and cannot know the bake is stale. `zot-registry.tf:78-80` claimed:

> `rotation via terraform apply -replace=random_password.zot_pull re-propagates htpasswd + Doppler in ONE apply`

**That claim is false, and that comment is the bug.** It documented a guarantee the code did not provide — which is why three consecutive fixes in this area (#6452, #6424, #6421) never looked here.

**The discriminator:** `crane`/`cosign` **push** authenticates against the same `/etc/zot/htpasswd` in the same window that `zot-pull` is rejected. One entry current, one stale — the shape only a per-entry divergence produces. A wholesale-stale or missing htpasswd would break both.

## Why it stayed invisible

The one datum that decides this was destroyed at the source: `docker login … >/dev/null 2>&1` (`ci-deploy.sh`) made `login_failed` a single undifferentiated bucket for bad-credential / authz-denial / transport / TLS. The beacon carried no `host_id`. The registry host is deny-all/no-SSH, so zot's own auth log is never shipped.

[...]

Issue #6463, opened 2026-07-15
infra: cx33 AND the entire ARM line are orderable NOWHERE in EU — web-2 AND the zot registry are both UN-RECREATABLE; only cpx32 (~4x cost) fits the 8GB floor


deruelle:
**Live probe, 2026-07-15** (found while planning #6453). `hcloud`/`GET /v1/datacenters` against the prd token.

## The trap

`soleur-web-2` is **cx33 @ fsn1** (`apps/web-platform/infra/variables.tf:86`, `server_type = optional(string, "cx33")`).

**cx33 is currently orderable in exactly ONE datacenter globally: `hel1-dc2`.**

```
fsn1-dc14 orderable NOW:  ccx33 ccx43 cpx32 ccx13 cpx22 cpx52 cpx62 ccx63 ccx53 cpx12 cpx42 ccx23
hel1-dc2  orderable NOW:  ccx33 cx23 ccx43 cpx32 ccx13 cpx22 cpx52 cpx62 cx33 ccx63 ccx53 cpx12 cpx42 ccx23
```

The **entire CX (Intel) line is absent from fsn1** right now. Each location maps to exactly one DC (`fsn1 -> fsn1-dc14`), so there is no sibling DC to fall back into.

**Consequence: dispatching `apply_target=web-2-recreate` today destroys web-2 and cannot re-place it** — `resource_unavailable` — stranding the fleet a host with no rollback. That is a verbatim repeat of #6393.

## The irony

#6393 relocated web-2 hel1→fsn1 *specifically because* a `-replace` during a **hel1** capacity shortage destroyed it and couldn't re-place it. `variables.tf:80-85` documents exactly that. The relocation traded a hel1 stock problem for an **fsn1** stock problem — and web-2 has needed a recreate twice in the last week (#6393, #6416).

## The dilemma

With cx33, web-2 cannot be **both** cross-DC **and** recreatable:

| option | cross-DC HA | recreatable | cost |
|---|---|---|---|
| cx33 @ fsn1 (**today**) | ✅ | ❌ | baseline |
| cx33 @ hel1 | ❌ (reverses #6393) | ✅ | baseline |
| cpx32 @ fsn1 (4 vCPU AMD, 8 GB — matches cx33's RAM) | ✅ | ✅ | higher, needs pricing |
| ccx13 @ fsn1 (dedicated) | ✅ | ✅ | higher, needs pricing |

[...]

Issue #6288, opened 2026-07-09
zot registry container restart-loops (~4/min) post-disk-fix — non-ENOSPC, likely OOM on cx23 large-store scan


deruelle:
## Symptom (live, prod)

After PR #6284 / #6247 definitively fixed the zot disk-full (`pcent` 100→58 on the grown 60 GB volume, verified), the zot **container is still restart-looping** — `SOLEUR_ZOT_DISK zot_restarts` (= `docker inspect -f '{{.RestartCount}}' zot`) climbs steadily on the freshly-replaced host:

| dt (UTC) | pcent | zot_restarts |
|---|---|---|
| 16:55 (fresh boot) | 53 | 0 |
| 17:00 | 57 | 21 |
| 17:05 | 58 | 46 |
| 17:10 | 58 | 68 |
| 17:15 | 58 | 88 |

~**4 restarts/min, steady**, with the disk at **58% (NOT ENOSPC)**. This is the operator's second stated symptom ("registry restart-looping 800+ times") — the disk-full is fixed, but the restart-loop is a **separate, disk-independent** cause.

## Why this is distinct from the disk-full fix (#6247)

- The pre-fix restart rate (~4/min, under ENOSPC) and the post-fix rate (~4/min, disk 58%) are **the same** — strongly suggesting a disk-independent crash cause that the disk-capacity fix legitimately does not address (the ENOSPC was a separate symptom that filled the disk).
- zot is still **functional** between restarts: the post-merge Web Platform Release's zot-mirror push (crane copy + cosign-sign) **succeeded**, the `soleur-registry-disk-prd` heartbeat is `up`, and prod `/health` is 200. So this is **GHCR-fallback-covered reliability degradation, not a current user-facing outage**.

## Hypothesis (unconfirmed — telemetry gap)

Most likely **OOM / memory pressure** on the small `cx23` host during zot's large-store startup scan: the fresh host re-attached the preserved ~35 GB store, and `gc` (gcInterval 1h) has **not run yet** (~17:53 UTC first pass), so zot scans/holds a large store in memory. The `docker run --name zot` has **no `--memory` limit** (`cloud-init-registry.yml`), so a host OOM-kill → `--restart unless-stopped` → restart is plausible.

[...]

PR #6508, opened 2026-07-15
fix(infra): derive zot's memory cap from the host instead of hardcoding 7168m

deruelle:
## The bug

zot's cgroup cap was the literal `7168m` in `cloud-init-registry.yml:640`, with **nothing tying it to `var.registry_server_type`**.

That is the same missing-data-edge shape as the htpasswd bug this host's `replace_triggered_by` exists to fix (#6497), and just as silent. Switching the var to a 4 GB type leaves a **7168m cap on a 4096m host** — the cgroup limit can never bind, so zot is free to take the host down instead of being contained. That is silently back to the *uncapped-on-cx23* condition that produced #6288's 261-restart loop. Nothing would report it; the cap would simply stop working.

Found while sizing the host for #6497, whose fix cannot land until the registry can be recreated (#6463: cx33 is orderable nowhere in EU).

## The fix

```hcl
data "hcloud_server_type" "registry" { name = var.registry_server_type }

registry_memory_cap_mb = data.hcloud_server_type.registry.memory * 1024 - local.registry_host_reserve_mb
```

Read from the live Hetzner catalog, not a hand-maintained type→RAM map that rots.

**Provable no-op on today's host.** Verified against the live API:

| type | memory | derived cap | |
|---|---|---|---|
| **cx33** (current default) | 8 GB | **7168m** | **byte-identical to the literal it replaces** |
| cx23 | 4 GB | 3072m | |
| cpx32 | 8 GB | 7168m | |

## Bonus: phantom-type tripwire

The data source resolves at **plan** time. #6288 set this var to `cx32` — a type that **does not exist** — and the apply **destroyed the registry host, then failed** `server type cx32 not found` on the create, stranding it (`variables.tf:140`). Now:

```
Error: Resource not found
Resource (server type) was not found: name="cx32"     ← at plan, rc=1, nothing destroyed
```

## The cap was replicated — deriving it in one place was not enough

The boot-guard's own comment said so: *"cloud-init source of truth + probe default + docs"*.

**`zot-restart-plateau-6288.sh:52` held its own hardcoded `CAP_MB=7168`.** Against a 3072m container it would compare `zot_anon_mb` to `7168−1024=6144` — a ceiling the kernel kills the container long before — and therefore **PASS a starved host**. A gate that does not pin the thing it names, exactly per the merged learning of that name.

So the host now **reports the cap it is actually running under**, read from the container's live cgroup `memory.max` (authoritative — survives an operator `ZOT_MEMORY_CAP` override, unlike a templated copy), emitted as `zot_memory_cap_mb` in `SOLEUR_ZOT_DISK`. The gate consumes that, falls back to its constant only when the host reports nothing, **and says which it did**:

```
... zot_anon_mb below the 7168m cap [assumed (host reported none)]. #6288 remediation holds.
```

`scheduled-zot-restart-loop.yml`'s operator decode table also told operators the cap is 7168m; it now points at the reported field.

## Tests

`registry-boot-guard.test.sh`: **70 → 74** assertions. The old pin asserted the `7168m` literal and **correctly failed on this change** — it now pins the derivation, the cgroup read, and the telemetry field, plus a no-literal-in-executable-shell guard (comments still cite 7168m deliberately; they explain what it was and why it's gone).

**Mutation-verified:** reintroducing the literal fails 2 assertions. `terraform fmt` + `validate` clean. Probe re-verified against live telemetry (PASS; fallback branch exercised).

## Ordering

Independent of #6507 (different files), but the two together are what make the pending cx23 sizing trial *observable* rather than blind. This PR does not change the server type — that is a separate decision.

Refs #6288
Refs #6497
Refs #6463


deruelle:
## Verification: rendered the template through Terraform rather than reasoning about the escaping

`terraform validate` does **not** catch a `templatefile()` var mismatch (that's evaluated at plan), and this file's `$${...}` vs `${...}` escaping is easy to get subtly wrong — a bad escape breaks a prod host's boot. So I rendered it both ways and diffed the bytes.

### 1. The escaping is correct

Rendered with `zot_memory_cap_mb = 7168`, the line the host's shell actually executes:

```sh
ZOT_MEMORY_CAP="${ZOT_MEMORY_CAP:-7168m}"
```

`$${...}` correctly collapsed to shell-level `${...}`; `${zot_memory_cap_mb}` interpolated to `7168`. Render exit 0 (so every one of the template's 11 vars is supplied), and **zero unrendered `${...}` leaked** into the output.

### 2. The no-op claim, proven byte-for-byte

Rendering `origin/main`'s template and this PR's template with the same inputs:

```
OLD:     ZOT_MEMORY_CAP="${ZOT_MEMORY_CAP:-7168m}"
NEW:     ZOT_MEMORY_CAP="${ZOT_MEMORY_CAP:-7168m}"
>>> IDENTICAL
```

On cx33 the cap the host boots with is **character-for-character unchanged**. The derivation replaces the literal without moving the value.

### 3. The complete rendered diff — only the three intended changes

```diff
-      ZOT_ANON_MB=-1; ZOT_OOM_KILLS=-1
+      ZOT_ANON_MB=-1; ZOT_OOM_KILLS=-1; ZOT_MEMORY_CAP_MB=-1

+        cap_bytes=$(cat "$cg/memory.max" 2>/dev/null)
+        case "$cap_bytes" in
+          ''|max|*[!0-9]*) : ;;
+          *) ZOT_MEMORY_CAP_MB=$(( cap_bytes / 1048576 )) ;;
+        esac

-      LINE="SOLEUR_ZOT_DISK ... zot_anon_mb=$ZOT_ANON_MB zot_oom_kills=...
+      LINE="SOLEUR_ZOT_DISK ... zot_anon_mb=$ZOT_ANON_MB zot_memory_cap_mb=$ZOT_MEMORY_CAP_MB zot_oom_kills=...
```

The cap assignment does not appear in the diff at all. `zot_last_err` stays last (it is free-text and would corrupt key=value parsing of anything after it).

### 4. Two risks I went looking for in my own change — both clear

**The `-1` sentinel could have caused a false FAIL.** If `zot_nonsentinel_values` did not filter negatives, a window where the container is absent would report `zot_memory_cap_mb=-1` for every sample → `effective_cap=-1` → `threshold = -1 - 1024 = -1025` → every anon sample "exceeds" it → **FAIL on a healthy host**. It does filter (`scripts/lib/zot-telemetry-parse.sh:48`):

```sh
grep -oE "$1=-?[0-9]+" | cut -d= -f2 | grep -vE '^-' || true
```

`-1` is stripped, `reported_cap` goes empty, the fallback engages. Correct, and consistent with the sibling `ZOT_ANON_MB=-1` / `ZOT_OOM_KILLS=-1` sentinels in the same block.

**`sort -n | tail -1` takes the MAX cap** — the more permissive value, the wrong direction for a safety gate if two caps could coexist. They cannot: `SCOPED` is scoped to a single `NEWEST_BOOT` (`:80`, `:86`), and within one boot the cap is constant (the container is started once with `--memory`; `--restart unless-stopped` reuses it). A host replacement mints a new `boot_id`, so a 7168 sample and a 3072 sample can never share a scope. Max and min are the same value here.

### 5. Live re-verification

Probe run against merged main's query script (#6507), no override:

```
PASS: ... over 21338s / 75 events — restarts flat, no exit_code=137, zot_oom_kills=0,
oom_kills_5m=0, zot_anon_mb below the 7168m cap [assumed (host reported none)].
```

The fallback branch is exercised (the live host predates this change, so it reports no cap) and the gate now **says** it is assuming instead of silently asserting 7168.

Issue #6698, opened 2026-07-19
cron-gh-pages-cert-reissue: toggle succeeds but ACME cert never validates in the 14-min window + zero step-level telemetry


deruelle:
Follow-up from #6657. After the permission fix (#6687 + #6694, deployed v0.226.4), the reissue routine now executes the full cname toggle successfully — `PUT /pages` returns 2xx and DNS flips to DNS-only. But the GitHub Pages cert still does not reissue.

## Symptom
Two manual fires ~2.5h apart (07:45Z and 10:23Z on 2026-07-19), both:
- Preconditions pass; DNS flips apex+www to `proxied=false`; cname toggles (null → re-set) with **no** `reissue_failed` terminal (so the PUT succeeded).
- Cert stays `bad_authz` through the ~14-min poll window; DNS restores to `proxied=true`; cert never reaches `issued`/`approved`.
- Cert description: "The ACME authorization is in a bad state. We need to start over."

### 1. Validation doesn't complete in the DNS-only window
Hypotheses (not yet isolated — see gap #2, which blocks diagnosis):
- The 14-min `POLL_MAX_MS` window is too short for GitHub/Let's Encrypt to complete a fresh authorization after the toggle.
- Let's Encrypt failed-validation rate-limiting from repeated attempts (the earlier permission-403 fires also toggled the cert).
- The `proxied=false` DNS-only flip may not propagate to public resolvers fast enough within the window (observed public DNS still returning Cloudflare IPs `188.114.x` around the window; needs a controlled mid-window `dig @1.1.1.1/@8.8.8.8` to confirm the flip reaches `185.199.x`).

Candidate fixes: lengthen the DNS-only poll window; add a mid-window public-DNS-propagation assertion before starting the cert poll; back off on repeated attempts to respect LE limits.

### 2. Zero step-level telemetry (blocks diagnosis) — hr-observability-as-plan-quality-gate
[...]

Issue #6604, opened 2026-07-17
P1(security): #6588 PR 2 — the /workspaces LUKS cutover (the half that moves sole-copy user data)


deruelle:
**PR 1 (#6593) declared the encrypted volume. It encrypts nothing.** The volume is not born until this
job runs. This issue is the cutover — the half that actually moves user data.

`Ref #6588` · ADR-118 (`status: adopting` → `accepted` on soak-pass) · plan:
`knowledge-base/project/plans/2026-07-17-fix-6588-luks-encrypt-workspaces-volume-plan.md`

## What PR 1 already did

- `workspaces-luks.tf` — additive volume (no `format`, so the raw-device discriminator exists), key in
  the dedicated `prd_workspaces_luks` config, scoped read-only token, web-1 singleton.
- `workspaces-luks.test.sh` — 20 mutation-tested assertions, registered in `infra-validation.yml`.
- `OPERATOR_APPLIED_EXCLUSIONS` — load-bearing for mergeability (verified by counterfactual).
- ADR-118 + the `model.c4` "host-local NVMe" correction.

## The stakes, stated plainly

The volume holds **sole-copy** user data. `refs/checkpoints/*` is pushed by no refspec,
`session-sync.ts` autocommits only `knowledge-base/**`, and signup-provisioned workspaces have **no git
remote at all** — so ADR-068 §1's *"GitHub remains the durable rehydration source"* does not hold. There
is no second copy anywhere. This fix also **creates** a terminal failure mode that does not exist today:
passphrase or header loss ⇒ unreadable forever. Today's worst case is someone else reads the user's
code; post-LUKS it is that the user cannot.

## Scope

- [ ] `workspaces-luks-cutover` dispatch job (template: `git_data_host_replace`), incl. the ephemeral
      SSH keygen step (HCL evaluates `file()` at plan time regardless of `-target`).
- [ ] Add to the `apply_target` **`type: choice` options** + the description — `choice` rejects any value
      not in `options`, so the job is otherwise **undispatchable**.
- [ ] Job-level `concurrency: group: web-1-swap` — `warm_standby` and `web_2_recreate` each declare their
      own; a job that stops the container and rewrites fstab **must** serialize against them.
- [ ] `tests/scripts/lib/workspaces-luks-cutover-gate.sh` + `tests/scripts/test-...-gate.sh` (CAP-COUPLING).
[...]

Issue #6462, opened 2026-07-15
observability(infra): zot soak has no denominator — fresh-boot GHCR service is invisible (blocks GHCR retirement)


deruelle:
Found while fixing #6435 (the soak's 2-of-4 signal blindness). **This blocks GHCR retirement (ADR-096 5.3–5.5) and outranks the blindness fix it came from.**

## The gap

#6435 added the two missing numerator terms to `scripts/followthroughs/zot-soak-6122.sh`. **Adding numerator terms cannot close a denominator gap**, and this probe has no denominator: `FALLBACKS == 0` means *"no GHCR service was reported"*, not *"none occurred"*.

The sibling probe already has the right shape — `zot-mirror-connector-6416.sh` enumerates workflow runs and turns any run it **cannot account for** into TRANSIENT. The soak has no equivalent.

Three concrete holes, all on the fresh-boot (web) path:

1. **Probe-miss emits nothing — and it is the DOMINANT path.** `cloud-init.yml`: if the `/v2/` probe misses, `REF` stays the GHCR ref, the pull succeeds on the first try, and the emit's guard (`REF != IMAGE_REF`) never fires. So the fleet is GHCR-served and Sentry sees **nothing at all**. At the time of filing, `probe_unreachable` was the reason on **34 of 38** live `zot-gate-degraded` events — i.e. the majority branch is the silent one. `stage:"app_ghcr_fallback"` catches only the *minority* branch.
2. **No fresh-boot liveness beacon.** There is no `app_zot` counterpart to the inngest one. So "0 fresh-boot fallbacks" is indistinguishable from "no fresh boot happened".
3. **`_emit` tags cannot satisfy the sample arm.** The boot emitter writes only `{stage,image_ref,host_id,detail}` — no `feature`/`op`/`registry` — so fresh boots can never contribute to the `registry:"zot"` sample that is supposed to prove the flip was exercised.

## Why it matters

5.3–5.5 is **irreversible**: it rotates *and revokes* the PAT. After it, a fleet that still needs GHCR can pull from neither registry, with no rollback. A soak that cannot distinguish "healthy" from "never observed" is not sufficient evidence to authorize that.

ADR-096 has been amended (in the #6435 PR) to state the soak is **necessary but not sufficient** for 5.3–5.5, naming this gap. That amendment is a disclosure, not a fix — this issue is the fix.

## Minimum closure

[...]

PR #6700, opened 2026-07-19
fix(web-platform): make the GitHub Pages cert reissue diagnosable and gate its DNS-only window

deruelle:
## Summary

`cron-gh-pages-cert-reissue` completed its full toggle but the ACME cert never validated, and the failure could not be isolated because the routine was **observationally dark on its success path** — two gates stacked: inngest's `ProxyLogger` is disabled outside an executing step, and Vector drops pino INFO. So the only diagnosis available was hand-run App-JWT probes.

This makes the routine self-reporting, stops it wasting its own outage window, and closes the hole that let a follow-through tracker be closed while still unrecovered.

- **Step markers** — a `SOLEUR_CERT_REISSUE` pino-WARN marker per phase, including inside `emitTerminal`. That last one is the headline fix: benign outcomes route through `logger.info`, so without it the `issued` success path stays dark even with every other marker added.
- **DNS-propagation gate** — confirms from public resolvers (1.1.1.1 / 8.8.8.8) that the flip actually propagated before the cert poll begins. A poll that starts early is guaranteed-wasted outage window.
- **Probe-only mode** — measures propagation without consuming a Let's Encrypt validation attempt, and is the **default**; remediation needs an explicit `{"probeOnly": false}`. LE's failure limits are hourly and compounding, and a blind fire deepens them while discriminating nothing.
- **Sweeper reopen path** — the follow-through sweeper listed `--state open` only, so a tracker closed while still unrecovered was invisible to it forever. #6657 was closed `COMPLETED` while its cert was (and is) `bad_authz`.

Ref #6698. Ref #6657.

Plan: `knowledge-base/project/plans/2026-07-19-fix-gh-pages-cert-reissue-telemetry-and-dns-propagation-gate-plan.md` (carries the `## User-Brand Impact` and `## Observability` sections).

## AC1 — Phase 0.1 blocking gate: H-W4 refuted

The plan made everything conditional on one read-only Cloudflare query, because a live AAAA on the apex would mean the routine **could never succeed at any window length** (Let's Encrypt prefers IPv6 and does not fall back from a proxied AAAA that answers 200 with the wrong content). Verbatim output:

```json
{ "success": true, "errors": [], "count": 0, "records": [] }
```

Zero AAAA records in the zone. The AAAA public resolvers return is Cloudflare's synthetic proxy answer, which `proxied=false` removes. **H-W4 refuted** — proceeded to the full plan. The gate is retained in code as an in-flight detector: a confirmed AAAA is terminal, not retryable.

### Web platform

- Add `server/cert-reissue-marker.ts` — module-scope pino WARN marker (mirrors `claude-cost-marker.ts`), with field names that avoid the eight keys `pii_scrub_drop_userdata` deletes, enforced by a test that derives the list from `vector.toml` rather than hand-copying it.
- Emit a step marker for every phase, including the terminal one, and from both branches of the `onFailure` handler.
- Add a DNS-propagation gate between the toggle and the poll. Raw observations live on the dep; the verdict is a pure exported function.
- Add probe-only mode (`probe_only_complete`) and `dns_propagation_failed` to `ReissueOutcome`.
- Widen `REISSUE_ALLOWED_STATES` to the documented terminal failure states — `errored` and `authorization_revoked` were previously declined as `not_stuck` and never remediated.
- Restructure the post-toggle tail to exactly one return site after restore, so "no post-toggle exit skips restore" holds by construction (`onFailure` does not fire on a clean early `return`).
- Derive `ok`/`errorSummary` so `runLogMiddleware` can tell a probe, a success and a paging failure apart in `routine_runs`.

### Ops

- Sweeper evaluates recently-closed `COMPLETED` follow-throughs and reopens any whose verification still fails, bounded by a stateless reopen cap, a `NOT_PLANNED` exclusion, and a no-op on PASS/TRANSIENT.
- Register `sweep-followthroughs.test.sh` in `test-all.sh` — it was an orphan suite gating nothing.

### Docs

- Amend ADR-125 (gate, probe-only, widened allowlist, corrected window budget) and correct the "only suspension point" claim in both the ADR and the file docstring.
- Model Let's Encrypt and the public resolvers in C4; fix a duplicate edge and a stale `administration:write`-alone claim.

## What review changed

Multi-agent review found seven defects that 192/192 green suites, `tsc`, semgrep and shellcheck all missed — every one in a seam a test cannot reach by construction. The two most consequential were plan-prescribed designs that were themselves wrong:

| Defect | Consequence if shipped |
| --- | --- |
| Wall-clock `break` in the replayed Inngest body | On resume it read a memoized `retry`, broke, and terminated `dns_propagation_failed` — discarding a memoized `propagated` and destroying the poll trajectory, on exactly the slow run this exists to diagnose. Banned by ADR-077. |
| Gate budget (2 min) shorter than Cloudflare's fixed 300 s proxied TTL | The paging outcome becomes the **default** result of a correct remediation. |
| Sweeper bypassing the `earliest` gate | Soak probes exit 1 for "still soaking", so #6604, #6416 and #6462 — all legitimately closed — would have been reopened at the next sweep. The bypass was also unnecessary: a 14-day lookback already covers #6657's soak. |
| `onFailure` reading the envelope, not the wrapped event | An operator's explicit `probeOnly:false` remediation recorded as a **probe**, with markers un-joinable to the run they describe. |
| `ReissueResult` missing `ok`/`errorSummary` | A probe, a success and a paging failure all write identical `completed` WORM rows. |
| `resolve6` coalescing every error to "no AAAA" | A resolver timeout green-lights exactly the state the gate exists to catch. |
| `SWEEPER_PASS_HEADING` read but never written | Reword the producer and the guard silently stops recognising its own output. |

One review agent recommended adding the wall-clock construct to the poll loop as well; that was declined because it would have replicated the first defect. Each fix is mutation-verified — including a 4-space-indent `return` that bypassed the post-toggle restore and passed 94/94 before the anchor was widened.

## Test plan

- [x] `scripts/test-all.sh` — full suite green
- [x] `tsc --noEmit` clean; semgrep 0 findings (2 files, 0 errors); shellcheck clean
- [x] Marker + orchestration suites (105) and sweeper suite (49) green
- [x] Mutation-verified: each new guard goes RED when its defect is reintroduced
- [x] Live DNS probe against the real zone, fed through the pure verdict function — confirms the `Server` header is the discriminator (both Cloudflare and GitHub answer 404 on the ACME path)
- [x] C4 model regenerated; freshness gate green
- [ ] ⏳ Post-deploy: fire probe-only, confirm step markers reach Better Stack across ≥3 phases

Verification of the last item is enrolled with the follow-through sweeper rather than left to memory — `scripts/followthroughs/cert-reissue-markers-6698.sh` asserts markers are actually readable, since "the code merged" is not the same claim as "the routine is diagnosable".


deruelle:
## rename-guard: `secret-scan-allow-rename` applied with evidence

The two renames are `/soleur:compound`'s standard end-of-feature archival — `session-state.md` and `tasks.md` moved from `knowledge-base/project/specs/feat-…/` into `knowledge-base/project/specs/archive/…/` by `archive-kb.sh` via `git mv`.

**Why this does not meet the gate's threat model.** rename-guard exists to stop a previously-scanned file being smuggled *into* a gitleaks-allowlisted path where it would stop being scanned. Here the **source path is already in the same allowlist** as the destination — both match `/knowledge-base/(?:plans|project/(?:plans|specs))/.*\.md$/`. Nothing moves from scanned to unscanned; the scanning posture is byte-identical before and after.

**Verified rather than asserted** (all three run against the two archived files):

| Check | Result |
| --- | --- |
| Canonical secret-shape scan (PEM, `ghp_`, `ghs_`, `github_pat_`, `sk-ant-api03-`, `sk_live/test_`, `sbp_`, `xoxb-`, `AKIA`, `dp.pt/st/sa/ct.`) | clean (rc=1) |
| Credential-shaped URLs (`scheme://user:pass@host`) | clean |
| `gitleaks detect --no-git` over the archive tree | `no leaks found` (1.69 MB scanned) |

The label is the sanctioned mechanism per the guard's own message; the alternative (`Rename-Allowed-By:` trailer) would require rewording an already-pushed merge commit for no additional safety.

Ref #6698

Issue #5097, opened 2026-06-10
compound: document rename-guard label/trailer for knowledge-base archival renames


deruelle:
## Problem

PR #5084's first auto-merge attempt failed CI on the required check `rename-guard (allowlist destinations)`. The compound phase's standard archival flow (`archive-kb.sh` + plan archival) renames `knowledge-base/project/{plans,specs}/**.md` into `archive/` subdirectories, and those destinations match the gitleaks-allowlisted path regex `/knowledge-base/(?:plans|project/(?:plans|specs))/.*\.md$/` — so the guard demands either the `secret-scan-allow-rename` PR label or a `Rename-Allowed-By: <name>` commit trailer.

Neither requirement is documented anywhere an agent reads: `grep -rln "secret-scan-allow-rename|Rename-Allowed-By" plugins/soleur/skills/ .claude/hooks/ AGENTS*.md knowledge-base/project/learnings/` returns zero hits. Every one-shot/ship pipeline that archives plan/spec artifacts at compound time will hit the same red CI round and burn a fix-push + full re-run cycle (~15 min wall clock per PR).

## Proposed Fix

Add the convention at the point of friction (≤10 lines, 2 files):

1. `plugins/soleur/skills/compound/SKILL.md` (Automatic Consolidation & Archival section): after archiving, note that the eventual PR needs the `secret-scan-allow-rename` label OR a `Rename-Allowed-By:` trailer on a commit in the range.
2. `plugins/soleur/skills/ship/SKILL.md` (Phase 6, semver/label step): when `git diff --name-status origin/main...HEAD` shows `R` entries into `knowledge-base/project/{plans,specs}/**`, apply `gh pr edit <N> --add-label secret-scan-allow-rename` proactively.

Recovery used in #5084 (for reference): `gh pr edit 5084 --add-label secret-scan-allow-rename` + an empty commit carrying `Rename-Allowed-By: Jean Deruelle` as a contiguous trailer, then re-queue auto-merge.

## Scope-Out Justification

Criterion: pre-existing-unrelated — the guard and its undocumented label/trailer contract predate PR #5084; the gap lives in the ship/compound skill docs (a different subsystem from the merged storage-retry fix), and the feature branch is already merged with its worktree reaped, so no inline fix site exists.

Re-eval by: event-grep — close when `grep -rln "secret-scan-allow-rename" plugins/soleur/skills/` returns ≥1 hit.

deruelle:
**Automated Triage**

[...]

PR #6709, opened 2026-07-19
fix(community): spool GitHub collector payloads and make collector failures self-report

deruelle:
## Summary

The 2026-07-19 community digest reported Stars/Forks/Watchers as `10 / 1 / 10 (stale — last confirmed 2026-06-08)` and claimed to cover a 41-day period. Neither number came from that run. This fixes the three collector defects underneath it, and — more importantly — makes the next failure report itself instead of surfacing as a footnote a human has to notice.

Ref #6695 (a scheduled digest issue; its Discord/X/Bluesky/LinkedIn/HN content is unrelated, so this references rather than closes it).

## The three defects

| # | Defect | Mechanism |
|---|---|---|
| RC1 | `github activity` and `contributors` died with `Argument list too long` | jq `--argjson` put the whole payload in **one** `execve` argument. The binding limit is **`MAX_ARG_STRLEN` = 131,072 B per argument**, not `ARG_MAX` (2 MB) — bisected: 131,071 passes, 131,072 fails. That is why it fired on as few as **10 commits** and recurred every single run. |
| RC2 | `github repo-stats` stargazer parse error | The fetch piped `2>&1` into `jq -s`, merging stderr into the JSON stream. One warning byte made the payload unparseable. |
| RC3 | The digest printed stale stats and an invented period | No code authorized either. The script failed loudly; the LLM filled a mandatory table from a six-week-old digest. |

## What changed

**Collector** (`github-community.sh`) — payloads spool to files and reach jq through a file descriptor. Every slurped binding uses one dereference shape, `($x | add // [])`, which flattens single-array and paginated bodies identically. The stargazer fetch captures stdout and stderr separately, keeping the diagnostic without corrupting the data. New guards turn an exit-0 error body or an empty body into a named `GITHUB_COLLECTOR_CAUSE=`, instead of an opaque jq indexing error.

**Observability** — this is the part that matters. Every failure channel the collectors had terminated in the spawned agent's context window: the scripts run as Bash tool calls *inside* `claude`, so their stderr never reaches `spawnResult.stderrTail`. `resolveOutputAwareOk` could not close the gap either — it checks whether a labelled issue was updated in the run window, so a fabricated digest and an honest one both resolve GREEN. The collector now appends a JSONL status record per dispatch and the handler reads it directly from `spawnCwd`, with no LLM in the path. A non-zero record pages and turns the monitor red; a truncation `warn` is reported without paging; an absent sidecar is reported as its own distinct condition, so "no signal" and "all green" cannot collapse into the same value.

**Prompt** — the failure clause amends the standing `log the error and continue` permission rather than appending to it, since the vaguer instruction otherwise wins.

## Six plan claims that did not survive execution

The plan was authoritative for intent, not for facts. Recorded in full in the plan's *Work-Phase Corrections* table; the two that changed the design:

- **The "plausible 0" path does not exist.** An exit-0 object error body makes jq hard-error (`Cannot index string with "starred_at"`, exit 5) — the script already failed loudly. The shape assertion's real value is a *named cause*, not closing a fabrication hole.
- **Cap detection fired on every run.** The `pulls` endpoint over-fetches a fixed page and filters by date client-side, so a full raw page is its steady state — it returned exactly 100 on the live repo every time. Measuring the cap post-filter for that endpoint keeps the signal worth reading. A detector that cries wolf daily is worse than none.

## What the review panel changed

Ten agents ran; nine returned findings and six converged on the same P1. Eleven were fixed in-PR. The three worth reading:

**The fix had a hole in its own thesis.** Lowering `heartbeatOk` on a collector failure was the try's *last statement*. A throw from `safe-commit-pr` jumps to the catch — which deliberately keeps `heartbeatOk` true for a trailing-step failure — so a collector failure would have paged **GREEN** on exactly the compound-failure run. That is the class this PR exists to close, reproduced inside the fix. It now applies after the inner catch closes.

**A converged P1 I'd have shipped.** `check_cap` was handed a *file path* where the refactored signature takes a count. `[[ /tmp/tmp.X -eq 100 ]]` is arithmetic evaluation, so it raised a bash syntax error, returned non-zero inside an `if`, and left truncation detection permanently dead on `contributors` — plus stderr noise on every run. My `replace(…, 1)` had converted only the first of two identical call sites.

**The suite was unverified where it mattered most.** Mutating `add // []` → `.[0] // []` at all six sites left it 36/36 green: every fixture was a single JSON array, so `--paginate`'s multi-array output — the entire reason the flattening exists — was never exercised. A two-page fixture gives 150 vs 100: a silent 33% undercount, the same plausible-wrong-number class as the bug being fixed.

Also fixed: the untracked sidecar made `safe-commit-paths-dropped` fire every run, drowning the control that detects a bot writing outside its allowlist; `warn` was written, typed, and read by nothing; the stargazer cap fired on a *complete* set (`--paginate` exhausts pages); and `cmd_fetch_interactions` ran in production with zero coverage while collapsing a lost fetch into `[]` at exit 0.

**Cut on review:** `digestFabricatesRepoStats`. It could not change any outcome — it required `repo-stats` to have failed, which already reds the monitor — and its regex failed *silently* on a bold-list rendering. The plan called it "the only deterministic control on RC3"; that was wrong.

**My sibling sweep was wrong.** The plan declared the scope "verified" while under-reporting its own grep by two sites — one paragraph after citing the learning that says *"X is unaffected" is a hypothesis*. `scripts/compound-promote.sh` was **already broken** at 8.2× the ceiling (1,073,302 B across 1,972 learning files); fixed here as a two-line `--slurpfile` swap with its 22-case suite still green. `domain-model-drift.sh` sits at ~55% and is filed as #6720. The false claim is corrected in the plan rather than quietly dropped.

One design change came from review pressure rather than the plan: `heartbeatOk` gates *both* the Sentry page and `safeCommitAndPr`, so lowering it on a collector failure would have discarded the honest digest — leaving the operator a red monitor and nothing to read. The gate now raises a flag applied after persistence. An earlier attempt renamed the persistence gate's variable instead, which broke `cron-safe-commit-parity` — that test asserts one literal gate shape across 9 migrated crons, and the rename would have quietly weakened a cohort-wide invariant to suit one handler.

## Plan

`knowledge-base/project/plans/2026-07-19-fix-github-collector-data-gaps-plan.md` — carries the corrected sibling sweep, the six work-phase corrections, and the eleven review-phase corrections.

## User-Brand Impact

**If this lands broken, the user experiences:** a community digest stating confident, wrong growth numbers — stars carried forward from a six-week-old run and presented as current, under a period the digest invented. The operator makes positioning and prioritization calls on fabricated traction data.

**If this leaks, the user's data is exposed via:** no new exposure. The collector reads only public repository metadata through already-granted `gh` auth. No new data category, destination, or retention.

- **Brand-survival threshold:** `none` — internal operator reporting, no external surface, no customer data.

**Sensitive-path scope-out:** `threshold: none, reason: the sensitive-path files are apps/web-platform/server/inngest/functions/cron-community-monitor.ts and _cron-safe-commit.ts, and the edits are confined to LLM prompt prose, a failure-signal read, and one structural-exclusion prefix — no auth, credential, payment, schema, or user-data path is read or written.`

## Verification

- **Live**, against `jikig-ai/soleur` at `days=1`: all **five** subcommands exit 0 with real data and clean stderr (`activity` 50 issues / 14 PRs, `contributors` 12 commits, `repo-stats` 11 stars, `fetch-interactions`, `discussions`). Three of them failed before.
- **Regression suite** — 54 assertions across 11 cases, auto-globbed by `test-all.sh` (verified present in the run log, not assumed). Plus 12 handler unit tests and the pre-existing 2,444-test inngest suite.
- **Mutation-tested**: 8/8 script and 4/4 handler mutations turn the suite RED — after **three** were caught vacuous and fixed. Two were digit-free fixtures whose guard never ran; the third was an ordering assertion whose bare `indexOf("} catch (err) {")` anchored on the first of **three** matches, in a different function, making it trivially true. A green mutation battery only measures the mutations its author imagined — three of this PR's holes were found by mutating something I hadn't.
- Full `TEST_GROUP=scripts` suite green; `tsc` clean; semgrep 77 rules / 0 findings (non-vacuous); shellcheck clean.

## Issue flow

Closing 0, filing 3 — net +3. Each is a discovered defect in a *different* subsystem, which the repo's own rule says must stay its own issue rather than be buried in a consolidated tracker:

- **#6713** — `workspaces-luks-freeze.test.sh` leaks `mktemp` files and filled `/tmp` *during this session*, aborting a test run and producing a misattributed failure in an unrelated suite. A discovered defect in a different subsystem; the repo's own rule is that these stay their own issue rather than being buried.
- **#6714** — why no digest was committed for 41 days. An availability question, distinct from collector correctness, and pre-declared out of scope in the plan.
- **#6720** — `domain-model-drift.sh` binds an unbounded accumulator on argv (~55% of the ceiling). Surfaced by correcting my own sweep. Not inlined because, unlike `compound-promote.sh`, it has no existing tempfile to slurp: the fix is a spool restructure plus a cap decision, not a binding swap.

## Changelog

- Fixed: GitHub community collectors no longer fail with `Argument list too long`; the limit is per-argument (`MAX_ARG_STRLEN`), which is why it fired on 10 commits.
- Fixed: `repo-stats` no longer corrupts its own payload by merging stderr into the JSON stream.
- Added: a collector-status sidecar so a collector failure reaches Sentry deterministically instead of dying in the agent's context window.
- Added: the digest must now report `collection failed: <reason>` instead of carrying a stale number forward, and must derive its period from the collector rather than inferring one.
- Fixed: `scripts/compound-promote.sh` hit the same argv ceiling at 8.2x and could not run at all.
- Corrected: the 2026-03-28 learning attributed the limit to `ARG_MAX` and concluded the sibling call sites were safe — the error that kept this bug alive for months.


deruelle:
**`secret-scan-allow-rename` override — rationale**

The `rename-guard` check failed on two `git mv`s performed by `/compound`'s standard KB archival:

```
knowledge-base/project/specs/feat-one-shot-6695-.../session-state.md
  -> knowledge-base/project/specs/archive/20260719-220502-.../session-state.md
knowledge-base/project/specs/feat-one-shot-6695-.../tasks.md
  -> knowledge-base/project/specs/archive/20260719-220502-.../tasks.md
```

The label was applied on evidence, not convenience:

1. **Both files scanned clean** against the canonical secret-shape set (PEM headers, `ghp_`/`ghs_`/`github_pat_`, `sk-ant-api03-`, `sk_test`/`sk_live`, `sbp_`, `xoxb-`, `AKIA`, Doppler `dp.*`, JWT) — zero matches.
2. **The source path was already allowlisted.** `knowledge-base/project/specs/…` matches the same `/knowledge-base/(?:plans|project/(?:plans|specs))/.*\.md$/` rule as the destination, so this is an allowlisted→allowlisted move and is *structurally incapable* of the laundering the guard exists to block — gitleaks' behaviour is identical before and after. The guard flags it because it evaluates the destination alone, which is the correct conservative default.

Recurring class, already tracked: #5097 (`compound: document rename-guard label/trailer for knowledge-base archival renames`) — every `/compound` archival on a `feat-*` branch trips this.

PR #6863, opened 2026-07-23
fix(preflight): stop skill body from auto-attaching a real secret file

deruelle:
## Summary

**Security fix.** The preflight skill's Check-10 denylist prose quoted an exfiltration example whose `curl --data-binary` argument was a literal `@` immediately followed by the real path `~/.doppler/.doppler.yaml`. When a skill body loads it is delivered as **user-turn content**, and Claude Code's `@file`-mention auto-attach resolves that path and reads the operator's **live Doppler token file** into model context — and into on-disk session transcripts — on **every** preflight run. It surfaced as a fake-looking `<system-reminder>` claiming a `Read` nobody initiated.

Root cause was confirmed via session-transcript forensics: the payload was an `attachment` of `type:"file"` (`filename: ~/.doppler/.doppler.yaml`) tied to the skill-load user turn — **not** any hook. Every configured hook (project/global/plugin) was verified clean and reproduced only benign output; the quoted path was the sole `@`-anchored real-file mention in the repo.

## Changes

- **preflight/SKILL.md** — rephrase the example so the real path sits in a plain code-span with the `@` **detached** (meaning preserved; nothing resolves), plus a blockquote warning against the pattern.
- **`plugins/soleur/scripts/lint-at-mention-secret-paths.sh`** — repo guard that fails on any `@`-mention resolving to a home/absolute real path (`@~/`, `@$HOME`, `@/home|Users|root|etc|…/`, or `@`+a path with `.doppler`/`.ssh`/`.aws`/`.netrc`/`.env`/`credentials`) in a skill/agent/command markdown body. Next.js `@/` import aliases (`@/server`, `@/lib`) and package scopes (`@types/…`) are intentionally **not** flagged — they resolve to nonexistent absolute paths.
- **`plugins/soleur/test/at-mention-secret-path-guard.test.ts`** — auto-discovered by `bun test plugins/soleur/`; regression against the live repo + positive/negative fixtures. Dangerous strings are built by concatenation so no literal `@`-adjacent real path exists in the test source.
- **security learning** documenting the footgun and the forensic method.

## Changelog

- Fixed: preflight skill no longer causes Claude Code to auto-attach the operator's real `~/.doppler/.doppler.yaml` (and other local secret files) into model context when the skill body loads. Added a CI guard + test to prevent regression across all skill/agent/command bodies.

## User-Brand Impact

- **Artifact:** the operator's local secret files (Doppler token; by extension any `@`-referenced home/absolute path) read into model context and persisted in session transcripts.
- **Vector:** a skill/agent/command body (or any user-turn content) containing a literal `@` before a real path → Claude Code `@file` auto-attach.
- **Brand-survival threshold:** single-user incident (operator secret exposure).

## Observability

```yaml
discoverability_test:
  command: bash plugins/soleur/scripts/lint-at-mention-secret-paths.sh
  expected_output: "OK: no dangerous @-real-path mentions"
  # Unauthenticated, local, no ssh / no gh / no doppler / no credentialed CLI:
  # the guard scans tracked markdown and exits 0 clean, non-zero on a violation.
```

## Notes

- Delivered as a **controlled, hand-built PR** (not autonomous one-shot) on purpose: quoting the vulnerable `@`-path in skill args re-triggers the auto-attach, so every delivered string here is kept defused.
- Separate operator action (out of scope for this PR): rotate the exposed credentials — verify the Doppler token, and the Discord/Hetzner tokens found in plaintext inside `.claude/settings.local.json` permission-allow rules.
- Independent of PR #6831 (action-required SLA contract); does not touch that branch.

### DIFF
--- a/knowledge-base/project/learnings/security-issues/2026-07-23-at-mention-in-skill-body-auto-attaches-real-secret-files.md
+++ b/knowledge-base/project/learnings/security-issues/2026-07-23-at-mention-in-skill-body-auto-attaches-real-secret-files.md
@@ -1,0 +1,87 @@
+---
+title: An `@`-path in a skill body makes Claude Code auto-attach the real local file (secret leak)
+date: 2026-07-23
+category: security
+tags: [claude-code, prompt-injection, secrets, skills, preflight, at-mention, footgun]
+type: bug-fix
+severity: high
+---
+
+# An `@`-path in a skill body makes Claude Code auto-attach the real local file
+
+## Symptom
+
+During `soleur:preflight`, a `<system-reminder>` appeared that looked like a
+prompt-injection: it claimed a `Read` of `~/.doppler/.doppler.yaml` (the Doppler
+CLI token file) that neither the operator nor the agent had initiated, and it
+surfaced the token value. It recurred deterministically on every preflight run,
+across 9+ session transcripts (byte-identical, because the Doppler CLI's
+`version-check` timestamp was frozen — so each live read returned the same bytes).
+
+## Root cause (not an attacker — a self-inflicted footgun)
+
+`plugins/soleur/skills/preflight/SKILL.md` documented, in the Check-10 denylist
+prose, an exfiltration example using the curl `@file` upload form whose argument
+was a literal `@` immediately followed by the real path `~/.doppler/.doppler.yaml`.
+
+When a skill loads, its SKILL.md body is delivered to the model **as user-turn
+content**. Claude Code's `@file`-mention auto-attach scans user-turn content for
+`@<path>` tokens and **reads the referenced file into model context** (recorded
+in the transcript as an `attachment` of `type:"file"`). So the security
+documentation, by quoting an `@`-prefixed real path, caused CC to read the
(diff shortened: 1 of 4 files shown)


deruelle:
Superseded by the parallel-shipped credfile-footgun fix already on main (`feat-one-shot-neutralize-credfile-paths-in-docs`, merged 2026-07-23).

Main already has, more comprehensively:
- `preflight/SKILL.md` line 843 neutralized to `@<doppler-config>` (identical fix)
- `.github/scripts/test/test-no-at-mention-credfile-footgun.sh` — a broader guard (covers skills/agents/commands + AGENTS.md/CLAUDE.md + hook dirs) wired into the **required** `guard-script-fixture-tests` job
- learning `2026-07-23-at-mention-auto-attach-credfile-exfil-footgun.md`

This PR's only non-overlapping detector branch (`@`+non-root-path+sensitive-basename, e.g. `@relative/.doppler`) is a non-threat — such a token resolves to a nonexistent/relative path and attaches nothing. No unique value to port. Closing to keep the backlog clean.

Issue #6575, opened 2026-07-16
chore(infra): sweep the dead web-2 dispatch surface after the #6538 retire — warm-standby + web-2-recreate target a host that will not exist, and two runbooks tell operators to run them


deruelle:
Follow-up to the #6538 web-2 retire (operator-chosen: land the retire first, sweep next — deleting this surface inside a prod-destroy PR inverts the risk budget). Surfaced by the #6538 plan-review panel (spec-flow P0-3, CTO P0-3, code-simplicity P0, DHH). Every claim below was verified against the worktree.

**These dispatches do not go dead — they go BROKEN.** They hard-`-target` addresses that will not exist post-retire.

### Dispatch jobs (both `-target` web-2 addresses directly)
- `apply-web-platform-infra.yml` `warm_standby` job — `-target='hcloud_server_network.web["web-2"]'`, `-target='hcloud_volume.workspaces["web-2"]'`, `-target='hcloud_volume_attachment.workspaces["web-2"]'` (4 of its 6 targets are web-2).
- `apply-web-platform-infra.yml` `web_2_recreate` job (~490 lines) — `-replace='hcloud_server.web["web-2"]'` + 3 `-target`s.

### Runbooks that instruct operators to run them (actively wrong post-retire)
- The `host_creates` HALT `::error::` text: *"hcloud_server.web[\"web-2\"] / its volume → dispatch 'gh workflow run apply-web-platform-infra.yml -f apply_target=warm-standby'"* — a dead runbook baked into the guard.
- `scheduled-inngest-health.yml` remediation: *"recreate the offending host via `apply-web-platform-infra.yml` (`apply_target=web-2-recreate`)"*.
- `tests/scripts/lib/stock-preflight-gate.sh` — a `web-2 → dispatch warm-standby` remediation tine keyed on `hcloud_server.web["web-2"]`.

### Gates / scripts / fixtures
- `tests/scripts/lib/web2-recreate-gate.sh`
- `apps/web-platform/infra/scripts/web2-recreate-preflight.sh` + `tests/scripts/test-web2-recreate-preflight.sh`
- `web2_allow` / `web2_out_of_scope_changes` / `web2_server_replaced` clauses in `tests/scripts/lib/destroy-guard-filter-web-platform.jq`
- 8 `tfplan-web2-recreate-*.json` fixtures

### Parity guards that will FAIL when the jobs are removed (must be updated in the same PR)
[...]

PR #6913, opened 2026-07-24
fix(ci): enroll the #6854 workspaces-luks-recut job in the web-1-swap parity guard (main red)

deruelle:
## Problem (main is RED for all deploy-script PRs)

`web-1-swap-concurrency-parity.test.sh` expected **exactly 3** job-level `group: web-1-swap` members, but **#6854** (gated `workspaces-luks-recut` apply_target) added a **4th legitimate member**: the `workspaces_luks_recut` job in `apply-web-platform-infra.yml` destroys+recreates web-1's LUKS `/mnt/data` volume, so it correctly shares the `web-1-swap` serializer (its own comment says it "shares the cross-pipeline web-1-swap group … and the workspaces_luks_cutover job above"). #6854 didn't enroll it in the guard's allow-list.

Result: `deploy-script-tests` has failed on **main** for every PR since #6854 — `expected exactly 3 'group: web-1-swap' occurrences, found 4`. This is the documented **all-members drift-guard reddens main when a sibling adds a member** class.

## Fix

Instrument the new member (the correct fix — do **not** loosen the guard):
- Add `assert_member "$APPLY_INFRA_WF" "workspaces_luks_recut" "workspaces-luks-recut"`.
- Bump the exact count `-eq 3` → `-eq 4`, with the history comment updated (5 → 3 after the #6575 web-2 sweep → 4 after #6854).

Guard now **15 passed / 0 failed** (was 12/1). shellcheck clean.

Found while this failure blocked an unrelated inngest cutover fix (#6910); fixing it here unblocks main for every deploy-script PR.


deruelle:
Folded the second pre-existing main-red drift into this PR. Both failing assertions live in the same required `deploy-script-tests` check, so neither this PR (web-1-swap parity) nor #6920 (model.c4 cron-monitor count) could go green alone — a circular deadlock. This branch now carries both fixes plus the regenerated `model.likec4.json` (so the `test-scripts` c4-freshness gate passes too). Merging this greens `deploy-script-tests` + `test-scripts` for all PRs. Supersedes #6920.

PR #6920, opened 2026-07-24
fix(ci): refresh model.c4 cron-monitor count 50→51 (main red)

deruelle:
## Problem

`deploy-script-tests` is red on `main` (and therefore on every PR branched from it, incl. #6913). The failing assertion is `model.c4 monitor count` in `apps/web-platform/infra/supabase-advisor/scan-workflow.test.sh`:

```
FAIL model.c4 monitor count: model.c4 says '50' but cron-monitors.tf declares 51 — refresh the github -> sentry description
```

## Root cause

`apps/web-platform/infra/sentry/cron-monitors.tf` declares **51** `sentry_cron_monitor` resources. The 51st, `cron-action-required-sla`, landed with #6831, but `model.c4`'s `github -> sentry` edge still read *"Of 50 cron monitors, 7 check in from here and 43 from webapp."* The scan-workflow drift guard asserts that count equals the live `.tf` count, so main went red.

## Fix

`cron-action-required-sla` is a **webapp** Inngest self-check-in (`apps/web-platform/server/inngest/functions/cron-action-required-sla.ts`, absent from every `.github/workflows/` heartbeat), so the webapp tally goes 43→44 and the CI-side "7 check in from here" is unchanged (7 + 44 = 51). Refreshed the c4 edge to *"Of 51 cron monitors, 7 check in from here and 44 from webapp."*

Docs-only reconciliation. `scan-workflow.test.sh` passes locally after the change (`model.c4 says 51 cron monitors; cron-monitors.tf declares 51 — all checks passed`).

## Impact

Unblocks `deploy-script-tests` for **all** open PRs, not just #6913.


deruelle:
Superseded by #6913, which now carries this model.c4 cron-monitor count fix (50→51 / 43→44) **plus** the regenerated model.likec4.json **plus** the web-1-swap parity fix — all three are needed together to break the deploy-script-tests circular deadlock and green main. Closing to consolidate.

Issue #6780, opened 2026-07-20
chore(infra): dedicated Inngest host has no in-place redelivery channel — any host-script change costs a full host replace (deferred C5.8)


deruelle:
## Summary

The dedicated Inngest host (`soleur-inngest-prd`, 10.0.1.40) has **no in-place redelivery channel**. Any change to a script that runs on it — even a one-line log marker — can only be delivered by baking a new image and force-replacing the host via `apply-web-platform-infra.yml -f apply_target=inngest-host-replace`.

The web hosts have an in-place channel (`ci-deploy.sh:2758-2891`, plus the infra-config push that delivers the probe scripts). The dedicated host does not.

## Why this is filed separately

This was Phase C5.8 of the #6617 plan (`knowledge-base/project/specs/feat-one-shot-6617-inngest-liveness-marker-registry-probe/tasks.md`). PR C is HELD per the operator ruling dated 2026-07-20, so C5.8 would never fire and this debt would go unrecorded — while remaining true regardless of what is decided about PR C.

## Why it matters

The delivery cost is structurally disproportionate to the change size, and the replace is not a cheap operation:

- it force-replaces the **sole production Inngest scheduler**
- `hcloud_volume.inngest_redis` must survive and re-attach
- four pin sites must move together (`cloud-init-inngest.yml:390` tag+digest, `cloud-init.yml:699` `IREF`, `cloud-init.yml:705` `ZIREF`, `inngest-bootstrap.sh:492` comment literal), and the cross-file assertion that would catch a partial bump does not yet exist — the existing AC6b binds to `cloud-init.yml` only
- the tag push reds `main` until all four pins land

The practical consequence is that observability improvements to this host are priced out. The #6617 work wanted three discriminator fields on an existing log line; two turned out to be obtainable off-host by other means, and the third (`sdk_url`, from `systemctl show -p ExecStart`) is still unobtained specifically because shipping it costs a host replace.

## Possible directions (not a prescription)

- Extend the infra-config push to target the dedicated host, so `apps/web-platform/infra/*.sh` lands there the way it lands on web hosts
- A pull-based config refresh on the host (timer + signed fetch), avoiding an inbound channel into a deny-all-public host
- Accept the constraint explicitly and batch host-script changes into scheduled replace windows — but record that as a deliberate decision rather than an emergent one

## Acceptance

- [ ] A change to a host script under `apps/web-platform/infra/` can reach the dedicated host without `apply_target=inngest-host-replace`
[...]

Issue #6178, opened 2026-07-07
arch: extract inngest to its own HA host — co-location on web backends couples cold-boot + pollutes the deploy tunnel


deruelle:
## Problem

Inngest currently runs **co-located on every web backend** (web-1, web-2) — there is no dedicated inngest host (terraform has only `hcloud_server.web["web-1"|"web-2"]`). This co-location is a coupling that is actively breaking cold-provisioning / disaster-recovery:

- `webhook.service` (the deploy control-plane) has a **hard mount-namespace dependency on `/var/lib/inngest`** (in its `ReadWritePaths`, no optional `-` prefix). If inngest's bootstrap doesn't create `/var/lib/inngest` on a fresh host, `webhook.service` fails with **226/NAMESPACE**, `:9000` never binds, the deploy fan-out can't reach the host, and the backend is **un-provisionable**. Observed on the #6090 web-2 recreate (run 28864603877): inngest bootstrap's GHCR login was skipped → inngest image pull failed → `/var/lib/inngest` missing → webhook 226/NAMESPACE.
- Inngest's cloudflared also co-inhabits the **shared web deploy tunnel** (`6410c1ec`, `deploy.soleur.ai`): a half-provisioned web-2 (inngest up, app dead) registered on the tunnel and Cloudflare round-robined `deploy.soleur.ai` to the dead backend → **502s on web-1's deploy control-plane** (had to manually stop web-2's cloudflared to clear it).
- Running one inngest server **per web backend** also raises a **split-brain / HA question** — are web-1 and web-2 both meant to process inngest events?

## Proposed direction

Extract inngest onto its **own HA host** (mirroring the existing `git-data.tf` separate-host precedent), decoupling it from the agent backends. Benefits:

1. Removes the `/var/lib/inngest` dependency from the web `webhook.service` → **eliminates the 226/NAMESPACE cold-boot failure class** (fresh web hosts no longer need to bootstrap inngest to boot).
2. Stops inngest from polluting the web deploy tunnel + deploy-status.
3. Decouples inngest's lifecycle from web-backend deploys/recreates.
4. Gives inngest real HA (single authoritative service) rather than one-per-backend with split-brain risk.

Note: there are already `cutover-inngest.yml` / `deploy-inngest-image.yml` workflows, so an inngest-separation direction may be partly scaffolded / previously considered — this issue should reconcile with that existing work + any ADR.

## Context

[...]

PR #6933, opened 2026-07-24
fix(6178): op=verify — anchor the doublefire scan window on the flip-FSM transition + null-safe bucketing

deruelle:
`op=verify` step 2.6 is the last open gate on the #6178 dedicated-host cutover, and across
eleven dispatches it **never produced a verdict**. This fixes two defects sitting one behind
the other, both on the critical path.

Ref #6178 — deliberately **not** an auto-close. The probe is delivered *by* this merge, so the
verifying run cannot precede it; #6178 closes only on a non-vacuous AC-V4.

## Defect 1 — the window could not be exhausted

The scan opened at `cutover − 200d`. Phase 0 measured the density nobody had measured
(run `30121678305`): **728 runs/day**, so 200 days is ~145,600 runs ≈ **1,456 pages** against a
~18-page budget. The probe is **fail-loud on non-exhaustion** — `_pf_abort` exits 1 and emits
**no run set** — so the run died on `reason=deadline` with no verdict, every time.

The deadline lever is dead: the SUM bound `DEADLINE + PAGE_MIN ≤ outer_curl` caps it at 112 s,
buying ~14 pages against the ~1,456 needed. The operative variables are **window width vs.
budget**.

```
anchor  = CUTOVER_ANCHOR_FROM (operator override)  # anchor_source=override
        | earliest flip-FSM TRANSITION row          # anchor_source=fsm
        | CUTOVER_WINDOW_FROM                        # anchor_source=var
        | (none) -> FAIL CLOSED, do not scan

DF_FROM = min( bucket_floor(anchor) − 2×cron_period , now − 1d )
```

### The anchor is a TRANSITION row, not "the earliest flip row"

This is a correction made at implementation time against the plan's literal predicate, and it
is the whole correctness of the anchor.

`inngest-cutover-flip` runs on a **~30-second on-host timer** and re-emits
`flag:"done" reason:"noop-done"` on **every tick** — ~2,880 rows/day. Measured against
production: a 400-row query over a 5-day `--since` spanned only **four hours** and was **100%
`noop-done`**. Since `betterstack-query.sh`'s `--limit` takes the *newest* N rows, "earliest row
returned" resolves to a few hours ago no matter how far back the query reaches — producing a
window **narrower** than the coexistence region. That is the unsafe direction, and precisely the
vacuous "exactly-once VERIFIED" that AC-V3 exists to reject.

Production holds exactly **one** transition row (`done/flip-complete` @ `2026-07-24 10:20:51Z`)
against thousands of heartbeats. Reasons are matched in **quoted** form, because
`noop-rolled-back` *contains* the substring `rolled-back`.

Fixed inline rather than routed as an architecture fork: the plan's *intent* — anchor on the
instant stamped by `10.0.1.40`'s clock, the same clock that stamps `startedAt` — is preserved
exactly. Only its row-selection predicate was wrong.

### Other decisions of record

- **`min()` is a skew clamp, not a safety net.** A 7-day floor is ~5,100 runs ≈ 214 s — not
  exhaustible. Safety and liveness are in direct tension and are resolved by making the anchor
  *correct*, not the fallback *wide*.
- **No wide fallback.** An underivable anchor **fails closed**; scanning `now − 7d` would trade a
  deadline abort for a deadline abort while *looking* safer.
- **Per-arm fallback passed as `$1`.** `op=doublefire-probe` (the pre-cutover dark-host detector)
  keeps 200 d and makes no Better Stack call; only `op=verify` narrows.
- **Truncation refuses.** A full `--limit` page means the earliest transition may be hidden, so
  the deriver refuses rather than derive an under-covering anchor.
- **`page-1` feasibility gate.** The query had *always* requested `totalCount` and nothing parsed
  it. Now an unaffordable window aborts in ~2 s with a **computed** latest-viable anchor instead
  of burning 112 s to say nothing.
- **Fail-closed `DF_FROM` shape guard.** `date -u -d ''` **succeeds** (today's midnight), so
  emptiness is not self-announcing — an empty value would build `?from=` and silently restore the
  probe's own 365-day default.
- **Dead remediation retired — twice.** All three abort surfaces advised enlarging the deadline,
  which the SUM bound caps at 112 s. Removed. Review then caught that the *replacement* was also
  dead (it named a variable the normal path ignores); see the review section below.

## Defect 2 — the bucketing `jq` died on a null `startedAt`

The probe projects `{functionID, startedAt}` from **every** returned node, and a run that is
queued, running, or cancelled-before-start carries `startedAt: null`. `fromdateiso8601` throws on
it (`strptime/1 requires string inputs`), and jq's exit 5 propagates through `set -euo pipefail`.

**Narrowing the window alone would have moved the failure from `reason=deadline` to a jq crash**,
and AC-V4 would have recorded "the fix did not work". It was invisible only because the scan had
never once completed far enough to reach the bucketing step.

Fixed at **all three** sites (both probe arms + the missed-tick `OBSERVED` expression), with the
dropped count surfaced — a silent discard is the false-clean shape this gate exists to prevent.

## Review: 7 agents, 0 scope-outs

Every finding was fixed inline — each is ≤100 lines / ≤4 files, so the cost-of-filing gate says
fix, not file. **Filed as scope-out: 0.**

**Five agents independently landed on the same defect:** the verdict was not gated on non-vacuity
*in code*. The probe emitted `total_count` precisely so `op=verify` could distinguish "no
duplicates found" from "nothing was looked at" — and the workflow read it **zero times**, printing
`exactly-once VERIFIED` unconditionally, including at `RUN_COUNT=0`. Two real inputs reach
`{runs:[]}` on an HTTP 200. Since a green verdict authorizes closing the issue and **deleting the
rollback snapshot**, the empty scan now hard-fails, and a clean result is *qualified* when the
population was scoped, the window overridden, or the anchor weaker than `fsm`.

**Four agents found the anchor set was incomplete.** `unexpected-exit(from=…)` is the ERR-trap
terminal transition and the only row emitted when `start_server` **succeeds** — coexistence begins
— but the following `flag_set` (a Doppler network write) fails. Omitting it returns a *later* row:
a window narrower than the coexistence region. Its interpolated suffix is exactly what hid it from
my original reason-vocabulary grep. Fixed, plus an **emitter-parity test** that derives the expected
set from `inngest-cutover-flip.sh`, so this class cannot recur silently.

**Three agents found my replacement remediation was dead advice — the same class this PR set out to
remove.** The page-1 gate told the operator to set `CUTOVER_WINDOW_FROM`, but that variable is
consulted *only* when the fsm anchor is absent, so on the normal path it changed the window by
nothing and the next dispatch aborted identically. Added `CUTOVER_ANCHOR_FROM`, which outranks the
fsm row; kept distinct because `CUTOVER_WINDOW_FROM` also means "quiesce→register gap start" to the
missed-tick enumeration, and quiesce precedes the cutover.

One recommendation was **declined on its reasoning**: security-sentinel proposed restricting the
anchor set to `start_server`-bearing reasons. `earliest(A ∪ B) ≤ earliest(A)`, so a larger set can
only move the anchor *earlier* — restricting it trades safety for width. The real problem it
surfaced was the *deadlock* (extra width trips the gate, whose remediation was inert), fixed above.

Also fixed: four surviving comments still asserting the retired "never narrowed" invariant; the
runbook's `reason=deadline` triage, which told operators to raise the dead lever and **explicitly
forbade** the correct one; `anchor_source` overwriting `fsm`/`var` with a bare `floor` exactly when
AC-V3 needs the distinction; four silent `return 1` branches in the anchor deriver; a fail-closed
message asserting a cause it never established; and a missing terminal anchor on the `DF_FROM`
shape guard that let the two-field emission ship a two-token value into `?from=`.

### Test vacuity: 16 surviving mutations my own battery missed

The sharpest, in my own new assertions:

- `grep -qE 'doublefire_from 1 fsm'` was **file-scoped** — it cannot tell a call from a comment. A
  one-line comment restored green after reverting `op=verify` to the 200-day window, i.e. **the
  defect this PR exists to remove was reinstatable at 291/291**. Now arm-scoped and comment-stripped.
- the extracted jq programs were executed but only asserted to *exit 0*, so replacing the null guard
  with `(.startedAt // "1970-01-01T00:00:00Z")` — which buckets every queued run into a phantom 1970
  tick and reports a **false double-fire** — survived. Now asserts output values in both directions.
- the `total_count` extraction test re-typed the expression instead of extracting it from the arm —
  a tautology that stayed green when the workflow dropped its `// "absent"` fallback.
- AC9's literal command returned **0**; my earlier verification used a masking case-insensitive
  variant, so I had verified a weaker claim than the AC states.

## Verification

| Suite | Result |
|---|---|
| `inngest-doublefire-probe.test.sh` | **88 passed, 0 failed** |
| `cutover-inngest-workflow.test.sh` | **291 passed, 0 failed** |
| `inngest-registry-probe.test.sh` | 75 passed, 0 failed |
| `infra-config-install.test.sh` | 29 passed, 0 failed |

### AC6 — the harness is non-vacuous

`doublefire_from()` is **extracted and executed**, not grepped. Mutating it to return an empty
string:

```
=== Results: 278 passed, 13 failed ===
  FAIL: fsm anchor, period=1200 -> bucket_floor - 2*period, source=fsm
  FAIL: var anchor, period=1200, source=var
  ...
```

Reverted: `=== Results: 291 passed, 0 failed ===`

### A mutation battery found two guards this PR adds were unpinned

Worth stating plainly: the first battery over the new guards had **two survivors** — the
`--limit` truncation guard (covered only by a static grep for `"limit"`, which survived deleting
the guard outright) and the fail-closed `DF_FROM` shape guard (unasserted at both call sites).
Both are now pinned, the truncation guard by **executing** `_flip_transition_dt` against a
PATH-stubbed `doppler`. All previously-surviving mutations now go RED.

## ADR ordinal

The plan reserved **ADR-142** as provisional. It is now **taken** on `main` by
`ADR-142-inngest-redis-aof-zero-data-loss-luks-migration` (landed `d4c31605c`). Renumbered to
**ADR-143** and swept across the plan, `tasks.md`, and every AC naming the ordinal.

## Delivery note

`inngest-doublefire-probe.sh` is a **web-host** script delivered in-place by
`apply-deploy-pipeline-fix` on merge — not on the dedicated host. No host replace; #6780 N/A.

## Follow-ups

Net issue flow: **closing 0, filing 2, net +2.**

Per-issue justification for the net-positive override:

- **Closing 0 is deliberate, not an omission.** #6178 must stay open pending a non-vacuous
  AC-V4 — the probe is delivered *by* this merge, so the verifying run cannot precede it. The
  body uses `Ref`, not `Closes`. Auto-closing it here is precisely the false-completion this PR
  exists to prevent.
- **#6939 cannot be inlined and must not be consolidated.** It is a discovered *defect* in a
  different subsystem (missed-tick enumeration), and re-firing one of the lines it emits causes
  the double-fire this cutover prevents. Scope discipline keeps a different-subsystem defect in
  its own issue and PR; the review rules additionally forbid burying a possible-P1 bug in a
  consolidated tracker.
- **#6940 is already a consolidation, not an addition.** It bundles three deferrals
  (registry-sourced discovery, `CUTOVER_*` env mapping, `PAGE_SIZE` measurement) that would
  otherwise be three issues — so it reduces net flow by 2. Two of the three are gated on AC-V4
  being green by construction: mapping `CUTOVER_REGISTRY_BASELINE` activates a dormant `exit 1`
  **upstream** of the doublefire check this PR exists to reach, so inlining it here would
  re-block the gate being unblocked.

- **#6939** (`p1-high`) — discovered defect: the missed-tick enumeration emits a command that
  does not exist (`--function-id`/`--missed-tick` are not `soleur:trigger-cron` flags, and no
  UUID→name mapping exists) *and* emits re-fire lines for crons that were never due. Re-firing
  one causes the double-fire this cutover prevents. Kept **separate** so a possible-P1 is not
  buried in a tracker.
- **#6940** (`p2-medium`) — three deferrals consolidated: registry-sourced discovery,
  `CUTOVER_REGISTRY_BASELINE`/`CUTOVER_QUIESCE_PROBES` env mapping, and the unmeasured
  `INNGEST_GQL_PAGE_SIZE=500`.
- **#6178 triage corrected inline** from `p2-medium`/`Post-MVP / Later` to `p1-high`/`Phase 4` —
  the old triage predated the cutover actually running in production.

## After merge

Both steps are automated `gh` invocations run by the pipeline; neither is an operator task.

1. **AC-V1** — `gh run watch` the merge-triggered `apply-deploy-pipeline-fix`; assert
   `files_written == files_total` and that the delivered sha256 for
   `/usr/local/bin/inngest-doublefire-probe.sh` matches `git show main:… | sha256sum`.
2. **AC-V2** — `gh workflow run cutover-inngest.yml -f op=verify -f cron_period_seconds=1200`,
   **only after AC-V1 passes**. The two workflows use different concurrency groups and nothing
   serializes them, so a premature dispatch lets the *old* probe answer and fail identically.
   `cron_period_seconds=1200` is load-bearing: `cron-ghcr-token-minter` runs `*/20 * * * *`, so
   the 3600 default would collapse three legitimate runs into one bucket and report a phantom
   double-fire.
3. **AC-V3/V4** — a clean verdict counts only if `total_count > 0` and `anchor_source` is
   non-`floor` (or a floor window demonstrably covers the quiesce instant). A detected
   double-fire **opens an incident** rather than closing issue #6178. Hetzner image `411798619`
   is retained until AC-V4 is satisfied.


deruelle:
[deploy_pipeline_fix-drift-gate] This PR edits `apps/web-platform/infra/inngest-doublefire-probe.sh`, a `terraform_data.deploy_pipeline_fix` trigger file. `apply-deploy-pipeline-fix.yml` auto-applies on merge — no operator action required.

This delivery **is AC-V1**: the probe is a web-host script delivered in place, and `op=verify` (AC-V2) must not be dispatched until it reports `files_written == files_total` and the delivered sha256 matches `git show main:apps/web-platform/infra/inngest-doublefire-probe.sh | sha256sum`. The two workflows use different concurrency groups and nothing serializes them, so a premature dispatch lets the OLD probe answer and fail identically.

Kill switch: add `[skip-deploy-fix-apply]` to a commit message.

Issue #6969, opened 2026-07-26
infra: web-2 booted DARK at cloud-init stage doppler_download — first real use of the web-host birth path


deruelle:
## Symptom

The **first real use** of the `web-host-create` birth path (#6730, ADR-145) created the host and then
booted it **DARK**. `soleur-web-2` exists and is `running` at the Hetzner level, but cloud-init died
at stage `doppler_download` and emitted a `fatal`. Per the path's own verdict:

```
##[error] web-2 booted DARK: a fatal boot-stage event was surfaced
(last-reached stage: doppler_download). runcmd is once-per-instance, so this host
CANNOT be repaired by a reboot — it must be replaced. Do NOT put it into service.
```

Run: https://github.com/jikig-ai/soleur/actions/runs/30211101264 (dispatch, approved 2026-07-26)

**The birth path behaved correctly.** `Terraform apply (web-host birth)` succeeded, and the very next
step detected the dark boot and failed the run loudly. This issue is about the boot failure, not the
gate — the gate is the reason we know.

## No user impact

`web-2` is at serving-weight 0 and was never put into service: the `dns.tf` app record stays
web-1-only, the `#6575` `lb-weight-gate.sh` anti-pooling gate fail-closes, and
`web_tunnel_connector = each.key == "web-1"` (`server.tf:238`) means a fresh web-2 registers no
cloudflared connector. web-1 is untouched (the merge-path plan showed it as a strict `no-op`).

## Evidence — attribution is solid

The boot trail, pulled from Sentry (`jikigai-eu/web-platform`):

| time (UTC) | stage | level |
|---|---|---|
| 16:50:03 | `bootcmd_start` | info |
| 16:51:42 | `bootstrap_complete` | info |
| 16:51:52 | `webhook_bound` | info |
| **16:51:58** | **`doppler_download`** | **fatal** |

The `fatal` is on the `doppler_download` event **itself**, and that event carries
`host_id=155488316`, which the Hetzner API resolves to `soleur-web-2` (`cpx22`, created
2026-07-26T16:49:35Z). So this is **not** a shared-project mismatch — worth stating explicitly,
because the Sentry project is shared with web-1 traffic and app errors, and the gate's `LAST_STAGE`
and its `level=fatal` test are computed over *separate* events. Here they coincide on one event.

## The failing code

`apps/web-platform/infra/cloud-init.yml` (search: `stage=doppler_download`):

```bash
stage=doppler_download
[...]

Issue #6730, opened 2026-07-19
infra: web-1 has no executable birth path — every automated route to creating the sole web host now HALTs


deruelle:
## Summary

**web-1 — the sole web host, and a hard-pinned singleton A record for `app.soleur.ai` — cannot be born by any automated path.** Every route that could create it now terminates in a `host_creates > 0` HALT, and the one dispatch the HALT's own remediation used to name (`warm-standby`) both targets a retired host and now carries the same HALT.

This is a **deliberate, correct** state for today — an unguarded birth is worse than no birth (see "Why the HALTs are right" below). It is filed because the *absence of a control* has now become the *absence of a capability*, and that violates a hard rule.

## Rule violated

`hr-fresh-host-provisioning-reachable-from-terraform-apply` — *"Every prod service in `apps/<app>/infra/` MUST come up on `terraform apply` from empty state, zero operator actions."*

web-1 does not satisfy this. Scoped precisely: this is about **web** hosts. `inngest_host` and the registry dispatches are separate and out of scope here.

## How we got here (each step individually right)

1. **#6416** added the `host_creates > 0` HALT to the per-PR `apply` job. Correct: that path can reach `hcloud_server.web` but *not* its `hcloud_server_network` attachment, so a birth there yields a host with no private IP and, transiently, no firewall.
2. **#6538** retired web-2 (2026-07-17). `var.web_hosts` now holds only `web-1`.
3. **#6718 / PR #6725** wired the same HALT into `warm_standby`, which reaches `hcloud_server.web["web-1"]` transitively via `-target` and passes **no `-var image_name`** — so a birth there boots the mutable `:latest` tag while cloud-init re-verifies `host_scripts_content_hash` under `set -e`.

The composition none of the three steps owned: the `apply` HALT's remediation text routed web-host births to `warm-standby`. After (2) that dispatch targets a host that no longer exists; after (3) it HALTs too. **The last named exit closed.**

## Why the HALTs are right (do not "fix" this by weakening them)

A web-1 birth on an incoherent image aborts the entire cloud-init `runcmd` at `stage=verify` — no cloudflared connector, no deploy webhook, no monitors, no egress firewall. `runcmd` is once-per-instance, so **no reboot repairs it**, and there is no failover partner (web-2 retired; #6459 unbuilt). The correct resolution is a *birth path that is safe*, not a removed tripwire.

## What "done" looks like

[...]

PR #6970, opened 2026-07-26
fix(6969): make the cloud-init doppler_download fatal name its own cause

deruelle:
## Summary

`soleur-web-2` booted **DARK** at cloud-init stage `doppler_download` — the first real use of the
`web-host-create` birth path (#6730, ADR-145). The birth path behaved correctly and failed its own
run loudly. The problem is what the failure *said*: the Sentry `fatal` named only the **stage**,
never the **cause**, because the failing call discarded the Doppler CLI's stderr and exit code at
the source — and Vector's shipping token comes from the very fetch that failed, so the host had no
second channel to fall back on.

Every root-cause hypothesis is still `UNKNOWN`, and no further reasoning can close them: the
discriminating datum was never captured. So this PR ships **the component's own error channel
first**, ahead of any attempted fix, per the `2026-07-11` webhook-202/E2BIG learning.

Ref #6969
Tracks #6971

### What changes

- **`soleur-boot-emit`** gains a per-stage `detail` tag (read from `/run/soleur-stage-detail.d/<stage>`,
  with **no** legacy fallback — a fallback would let one stage ship another stage's captured output)
  and a `host_name` tag. Credential redaction lives **in the emitter**, the single choke point every
  producer flows through, rather than in each producer.
- **New baked helper `soleur-doppler-download`**: `timeout -k 5 20`, 3 attempts, 5s/10s backoff,
  stderr-only capture, scrub, per-attempt `doppler_retry` breadcrumbs (bounded emits), and the final
  summary the fatal trap reads. The failing call was the **only unbounded Doppler invocation** in
  `cloud-init.yml` (measured: 11 bounded siblings, 1 unbounded — and the unbounded one is the one
  that failed), so the channel was structurally blind to a hang.
- **Fail-closed call site**, twice over: `&& rc=0 || rc=$?` **plus** `[ "$rc" = 0 ] || exit "$rc"`.
  Both are mandatory; either alone regresses.
- **Birth-path gate** (`apply-web-platform-infra.yml`): its `::error::` now names the cause, is
  **host-scoped** (a missing `host_name` still matches, so the filter can only ever exclude events
  provably from another host), and emits a `::warning::` on a *green* birth that nonetheless
  retried — a retry that succeeded still names a real fault.
- **ADR-147** — boot-stage diagnostics live in baked host-scripts, not `user_data`. This is the
  load-bearing constraint: comments in `cloud-init.yml` cost `user_data` bytes (the render failed at
  23,860B once, on a comment), so rationale belongs in the baked ADR, where it is free.
- **A committed mutation harness** (30 mutations) alongside the 88-assertion suite.
- **PIR** for the incident, recording an UNDETERMINED root cause deliberately — the finding is that
  the evidence was discarded, not that the write-up is incomplete.

### Scope deliberately excluded

`docker_run` stderr capture was **removed on purpose**. `docker run --env-file` holds the whole prd
secret set and quotes offending lines back verbatim on a parse error; no regex closes that. Measured:
prd is 129 keys / 129 lines, so the leak is latent today and would go live with the first multi-line
secret. Tracked in #6971 — please do not re-add it without classification.

### Why `Ref`, not a closing keyword

#6969 also carries an unresolved open decision — what to do with the dark `soleur-web-2`, which is
still occupying a fleet slot and is tracked nowhere else. This PR scopes the destroy-and-rebirth
out. The recommended sequence is to land this diagnosability change first, **then** do the
destroy-and-rebirth in one window (destroying re-arms the `host_creates` HALT, because the host is
declared in `var.web_hosts`), and read the named cause off the fresh birth.

### This change reaches no running host

`hcloud_server.web` pins `user_data` under `lifecycle.ignore_changes`, and `runcmd` is
once-per-instance. The new tags therefore take effect **only at a fresh create** — a `terraform plan`
showing no diff is expected, not evidence of no effect. It cannot repair `soleur-web-2`.

Recorded per ADR-084. Neither was applied — the stated direction was the default in both cases.
Also filed as #6972 so it surfaces in the digest.

- **DC-1 — cut the bounded retry?** Three independent signals argued for shipping the error channel
  alone (`code-simplicity-reviewer`: *"a real defect created by the retry the plan should not be
  shipping"*; the `2026-07-16` invisible-discriminator learning; the `2026-07-11` error-channel-first
  learning). Retained because #6969 explicitly scopes it, because it is not a fix for a *diagnosed*
  cause (all hypotheses are `UNKNOWN`), and because it was made evidence-**preserving** — each failed
  attempt emits its own breadcrumb before sleeping. Written as a separable block, so cutting it stays
  a clean, small descope.
- **DC-2 — `host_name` narrowed.** The stated direction was "the whole boot-stage emitter"; this
  narrows to the **baked** emitter only, deferring the `bootcmd` beacon and inline `_emit` to #6971.
  Folding a three-site attribution change into a P0 error-channel change makes rollback expensive,
  and the inline `_emit` costs `user_data` bytes against a budget with ~300B of headroom.

### Web Platform (infra)

- Boot-stage Sentry events now carry a redacted, ASCII-only, 180-byte-capped `detail` tag and a
  `host_name` tag, so a dark-boot fatal names its own cause instead of only its stage.
- The cloud-init Doppler secrets download is now bounded (`timeout -k 5 20`), retried 3× with 5s/10s
  backoff, and fail-closed after the retries, with per-attempt breadcrumbs.
- The fresh-host birth path's boot-trail gate is host-scoped, surfaces the captured cause in its
  `::error::`, and warns when a host booted green only after retrying a stage.
- Sentry issue-alert rules document the new tag payload and its redaction contract.

### Knowledge base

- ADR-147 (boot-stage diagnostics live in baked host-scripts, not `user_data`).
- Post-incident report for the `soleur-web-2` dark boot.
- Learning: an existence assertion that ran before the file existed bricked every boot.

## Test plan

Every number below is a re-run result from this branch, not a recollection.

- [x] `bash apps/web-platform/infra/run-registered-suites.sh` — **72/72 passed** (the authoritative
      gate for this diff; `scripts/test-all.sh` does **not** cover `apps/web-platform/infra/`, where
      nearly all of this change lives)
- [x] `bash apps/web-platform/infra/doppler-download-error-channel.test.sh` — 88 passed, 0 failed
- [x] `python3 apps/web-platform/infra/doppler-download-error-channel.mutation.py` — 30/30 mutations
      detected
- [x] `bash apps/web-platform/infra/soleur-host-bootstrap-observability.test.sh` — 95 passed, 0 failed
- [x] `bash scripts/test-all.sh` — 226/226 suites, exit 0
- [x] `cd apps/web-platform && ./node_modules/.bin/vitest run` — 1043 files / 12,622 tests, 0 failed
- [x] `bun test plugins/soleur/test/cloud-init-user-data-size.test.ts` — 30 pass; render 23,400B of a
      23,700B budget
- [x] `bash .github/scripts/validate-infra-templates.sh apps/web-platform/infra` — 7/7 rendered and
      validated
- [x] `python3 scripts/lint-encryption-posture.py --repo-sweep` — 14 stores, 3 connections, 0
      unledgered
- [x] `bash scripts/check-adr-ordinals.sh` — passed (ADR-147 free at time of merge)
- [x] Sentry ingest transport reachable without SSH: `curl -sS -o /dev/null -w "%{http_code}"
      --max-time 10 https://o4511404939345920.ingest.de.sentry.io/api/4511404943671376/store/` → `401`
      (auth-gated but reachable — the healthy signal)

Known pre-existing and **not** caused by this branch: `workspaces-luks-loopback.test.sh` exits 2
(`LOOPBACK_UNAVAILABLE`, needs root/sudo). Reproduced identically on pristine `origin/main`.

Verified at the next fresh host birth on a post-#6969 image: the gate must either reach
`cloud_init_complete`, or fail with an `::error::` naming `stage=doppler_download` **together with**
the CLI's own error line and exit code.

Override justification — #6972 is an ADR-084 model-dissent disclosure record, mandated by `/ship`
Phase 6 Step 2.5 for any headless-auto-decided dissent, and filed purely so the dissent reaches the
operator digest (which harvests `action-required` issues, not PR bodies). It is an informational
artifact, not deferred engineering scope, and it carries no work item of its own — the one real
follow-up it names (#6971) was already open before this PR.

Override justification — the soak signal is a false positive on the single word "Soak" in the plan,
on a line that reads *"Soak follow-through (§2.9.1): not engaged — no AC is time-gated"*. No
acceptance criterion in this change is time-gated, so there is no soak to enroll in the sweeper.


deruelle:
rename-guard override — why this is safe (recorded, not silent)

`rename-guard (allowlist destinations)` failed on three renames made by `/compound`'s
archive step in `ddce4def5`:

```
knowledge-base/project/specs/feat-one-shot-6969-.../{decision-challenges,session-state,tasks}.md
  -> knowledge-base/project/specs/archive/20260726-213425-feat-one-shot-6969-.../
```

Applied the sanctioned `secret-scan-allow-rename` label. The threat model the guard
protects against — `git mv` a file holding a real secret from a *scanned* path into a
*gitleaks-allowlisted* path, where gitleaks evaluates only the destination — does not
apply here, on three independent grounds:

1. **This is an allowlisted -> allowlisted move.** The matching regex is
   `/knowledge-base/(?:plans|project/(?:plans|specs))/.*\.md$/`, which matches the
   **source** path as well as the destination. Nothing moves from a scanned path into a
   relaxed one, so there is no scanning coverage to launder past.
2. **All three are `R100`** — 100% similarity, pure moves with byte-identical content. No
   content was introduced or altered under cover of the rename.
3. **Both paths were created on this branch**, so no pre-existing blob could be relocated.
   The files were also grepped for credential shapes (Doppler `dp.st.`/`dp.pt.`, GitHub
   `ghp_`/`github_pat_`, `sk-`, PEM headers, `Bearer`, AWS `AKIA`, URI userinfo) — no
   matches.

This is a benign false-positive class for the routine `/compound` archive step, not an
exception granted to this change's own content.

PR #7071, opened 2026-07-29
fix(infra): make the zot mirror fail-closed, and stop the pipeline asserting a dead GHCR fallback

deruelle:
## Summary

On 2026-07-29 the web-platform release for **v0.244.1** built, pushed, and published **green**. The
deploy then died `image_pull_failed` and production sat undeployable for ~5h.

The release was green because the CI zot-mirror step is warn-only — a correct decision *while GHCR
was a working break-glass fallback*. That premise is no longer true: `GHCR_READ_TOKEN` is a revoked
classic PAT and `GHCR_MINTER_DISABLED=true`. So the pipeline was publishing releases whose image
production had no way to pull, and then asserting in annotations, step summaries, Slack, runbooks and
two ADRs that a fallback existed to cover it.

This PR makes the mirror fail-closed, makes the gate actually reach the deploy, and removes the false
fallback claims wherever they are written down.

## Changelog

- **fix(ci):** the zot mirror is now release-blocking. `continue-on-error` removed, `degraded()`
  exits non-zero, and a positive post-copy `crane digest` assertion proves the manifest is in zot
  **at the expected digest** before `mirror_status=ok` is emitted. Six degrade call sites route
  through the one emitter, so every failure arm carries a reason label.
- **fix(ci):** `migrate` and `deploy` now gate on `needs.release.result == 'success'`. Both `if:`
  conditions led with `always() &&`, which discards GitHub's implicit skip-on-failed-`needs`, and
  `deploy` gated only on the `docker_pushed` output — written ten steps before the new assertion. So
  the gate could have protected nothing while leaving prod on new schema + old code.
- **fix(ci):** `mirror_reason=<bridge|crane_install|copy_v|copy_sha|copy_latest|sign|verify>` on
  every degrade path, plus rewritten operator messages that name a cause, a remedy, and the
  pipeline-bypass hatch. The copy-arm message no longer blames Sigstore.
- **fix(ci):** `failure()` ops-email on the release and deploy jobs; `mirror_verified` surfaced to
  the deploy summary; a dispatch-only, reason-required override.
- **fix(infra):** `ci-deploy.sh` now logs a bounded single-line tail of the zot-pull stderr it was
  discarding.
- **fix(infra):** the Cloudflare token-drift detector now covers Access **service tokens** and is
  actually invoked. Its enumeration regex `CF_API_TOKEN[A-Z0-9_]*` could not match
  `REGISTRY_PUSH_ACCESS_TOKEN_*` — the first case its own header cited — and nothing called the
  script at all.
- **fix(secret-scan):** narrow `generic-api-key` carve-out for the new synthesized fixture.
- **docs:** truthfulness sweep across `zot-registry-revert.md`, ADR-096, ADR-088, the
  principles-register AP-016 row, and three `model.c4` descriptions.

25 files, +3680/-133.

## Root cause, measured rather than asserted

The plan's own v1 was reviewed by 7 reviewers and returned 5 P0s, three of which invalidated v1's
design or premises. Two of three Deliverable-B premises were **falsified** and one "Confirmed"
premise was **retracted**. What survived measurement:

| Premise | Evidence |
|---|---|
| GHCR is dead | `GHCR_READ_TOKEN` is a 40-char `ghp_` classic PAT. `GET api.github.com/user` → **401**. Registry pull-token mint → **403** `{"errors":[{"code":"DENIED"}]}`. `GHCR_MINTER_DISABLED=true`. |
| The bridge is healthy | `cloudflared access tcp --hostname registry.soleur.ai --url 127.0.0.1:15000`, then `GET /v2/` → **HTTP 401 from zot itself** (`Www-Authenticate: Basic realm=…`, zot-shaped JSON). That is the origin answering. |
| Deploy failed `image_pull_failed` | Job 90634334826: `ci-deploy.sh exited 1 (reason=image_pull_failed, tag=v0.244.1)`. |
| Mirror was warn-only | `degraded()` → `mirror_status=degraded`, `exit 0`, `continue-on-error: true`. Run 30468080168 reproduces the annotation verbatim. |
| Host image verify is **warn**, not enforce | `IMAGE_VERIFY_MODE="${IMAGE_VERIFY_MODE:-warn}"`, with two tests pinning the default. An unsigned-but-present copy deploys today. |

**Falsified F1.** The `HTTP 200` + empty body from `https://registry.soleur.ai/v2/` is *correct
behaviour*, not a symptom. The ingress is `service: tcp://10.0.1.30:5000` — TCP-mode, consumable
only via `cloudflared access tcp`. A plain HTTPS GET is not a WebSocket upgrade for that stream.

**Falsified F3.** The real cause was a CF Access service-token rotation that never propagated. From
the teardown step of run 30468080168:
`failed to connect to origin error="websocket: bad handshake"` — the *client* failing at the edge, so
CF Access refused the upgrade. A missing origin route would instead handshake and fail origin-side.

**Retracted P0-A.** v1 claimed a failed `release` job blocks the deploy, citing
`deploy: needs: [release, …]`. `needs.release.result` appeared **0 times** in the workflow. That is
the same class of error this PR exists to fix: reading `needs:` and asserting a mechanism without
reading the `if:`.

## Second review pass — every mechanism this PR added was itself defective

Two commits here fix the PR's own additions, which is worth stating plainly rather than folding into
the summary:

- `9b3b69d5e` — the P0/P1 findings on the new mechanisms.
- `7b1d91e6d` — the new "Enforce token-drift result" step sat immediately after the token-drift
  check, so its `exit 1` skipped the **four** steps below it: the infra-drift label, the drift issue,
  the drift email body, and the drift email. Each carries a plain-expression `if:`, which Actions
  gives an implicit `success()`. A day with **both** a stale token and real Terraform drift would
  have reported **neither** — the step added to make token drift loud was darkening the channel it
  was added to, and combined with the enumeration defect that made it fire on every run, permanently.
  Now last in the job under `always() && …`. Verified by enumerating the rendered step list: index
  14 of 15, after the `always()` Sentry check-in.
  The same commit also corrects this job's claim to be "the workflow that CAUSES the drift" — it runs
  `terraform plan -detailed-exitcode` and never applies, so it **detects**.

## Disclosures

Two blast radii that are not obvious from the diff:

- **The `failure()` ops-email on the `release` job also fires for plugin releases.**
  `reusable-release.yml` has two consumers; `version-bump-and-release.yml` is the other. A failed
  plugin release will now email ops too.
- **The dispatch-only override also bypasses `await-ci`.** It is gated on
  `github.event_name == 'workflow_dispatch'` and requires a reason, but a run that uses it needs the
  **same `bump_type`** as the run it is recovering, or the version does not self-heal. No
  `environment:` reviewer clause was added — `jobs.release` has no `environment:`.

  The override does **not** launder the failure: `mirror_status` and `mirror_reason` keep their
  failure values and `mirror_verified` stays false, so an override-published release stays
  distinguishable from a verified one. A **digest mismatch is not overridable at all** — every other
  stage answers "did the copy land", but that one answers "is the thing in zot the thing this release
  built", so it can never be a misfire, and overriding it would publish a release whose image does not
  match its own notes. The reason string is also control-byte-stripped before it reaches the
  workflow-command and markdown sinks, so it cannot forge an `::error::`/`::add-mask::` line and
  rewrite the audit record it *is*.

## What the new gate does NOT prove

The assertion reads the manifest with the **push** credential over the tunnel. It does not prove the
**host** can pull: that path uses `ZOT_PULL_*` over the private NIC and is a different credential and
a different route. `web-zot-consumer-probe.sh` is the complement. This is stated in the gate's own
comment block so the next reader does not over-trust it.

## Secret-scan carve-out

`gitleaks` `generic-api-key` fires on the synthesized 64-hex Access service-token secret the new
`scripts/check-cloudflare-token-drift.test.sh` introduces. The fixture is invented and its shape is
deliberate (`cq-test-fixtures-synthesized-only`) so a future format check exercises the real
Access-secret shape, so weakening the fixture was not an option. A line-level `# gitleaks:allow`
cannot clear it either: the finding is frozen in branch history and the scan reads the historical
blob, which carries no waiver.

Shipped as a second `[[rules.allowlists]]` block with `condition = "AND"`, so path **and** secret
must both match. Measured against gitleaks 8.24.2:

| Case | Result |
|---|---|
| this value, this path | quiet |
| same value, any other path | still **FIRES** |
| any other secret, this same file | still **FIRES** |
| carve-out deleted | branch scan returns 1 leak again |

Pinned by T11 (four fixture rows, one per collapse direction) and T11b (arity + anchor guards, which
run with no gitleaks binary present). Five mutants of the block all go RED on the rows that name
them. Two harness defects were found while proving this and are recorded in the commit message: an
absolute scan root silently defeats every `^`-anchored `paths` entry, and a sequence counter
incremented inside `$( )` was discarded by the subshell, so three must-FIRE rows were passing off
each other's fixtures while staying green against the very mutants they existed to catch.

## Decisions — resolved by the operator before implementation

`decision-challenges.md` records three Taste/User-Challenge decisions. All three were **decided on
2026-07-30 at the start of `/work`**, before any implementation, since two of them change what gets
built. They are settled, not open, and no `action-required` issue is owed.

| # | Decision | Outcome |
|---|---|---|
| UC-1 | Drop the new bridge `/v2/` probe? | **Keep it cut** (the recommendation). It is the only change with zero runner-side evidence, and it composes into the gate — one false negative would block the release train. The `nc -z` false gate folds into the tracker that also carries the live #6416 defect. |
| UC-2 | Move the token-drift detector fix to a follow-up? | **Full scope, 5.1–5.4** — broader than the recommendation, including the release-preflight arm. A twice-daily detector can be hours stale relative to the release that trips over it. DHH's "finish the detector separately" framing is recorded as the rejected alternative. |
| UC-3 | Split gating from non-gating work? | **Single PR.** UC-1 resolving as "cut" was the precondition the recommendation named. The hard constraint is satisfied by construction: the `zot-registry-revert.md` correction ships in the same PR as the gate, so no window exists where the runbook prescribes a recovery that now fails. |

## Issue linkage

**No `Closes:` ref, deliberately.** This work originates from the 2026-07-29 v0.244.1 incident, not
from a filed issue, and filing one retroactively just to close it would be backlog theatre. Net issue
flow is **closing 0, filing 3, net +3**, each justified in the plan's Deferred table:

- **#7077** — extend the fail-closed invariant to the inngest image, and fix the live #6416 skip.
  A discovered defect in another subsystem, so it stays its own issue rather than being folded into a
  tracker. Also absorbs UC-1's deferred `nc -z` fix, which the panel wanted done together with it.
- **#7078** — follow-through: verify the gate on the first real release. Genuinely un-automatable
  pre-merge; the faithful test is a production release.
- **#7079** — the consolidation. Absorbs three plan items: the two sibling false `/v2/` gates in
  `cloud-init.yml` (measured correction to the plan: **two** occurrences, not one — and not inlinable,
  since `hcloud_server.web` carries `ignore_changes = [user_data]`), the `zot-entry-gate.sh`
  wire-or-delete decision, and the orphan-draft / stale-release-notes leak.

A status comment went on the still-open **#6416** rather than a duplicate issue.

## Post-merge

- Verification of the gate on the first real release. **Tracks #7078** — the merge itself is the
  first execution of the new gate, since the diff matches `on.push.paths: apps/web-platform/**`.

## Test plan

- `plugins/soleur/test/reusable-release-zot-mirror-retry.test.sh` — passes with only the two
  expected-rc literals changed (`"0 degraded warn 3"` → blocking, `"0 degraded warn 0"`).
- `apps/web-platform/infra/ci-deploy.test.sh` — asserts the zot-arm log line carries the pull reason.
- `plugins/soleur/test/gitleaks-rules.test.sh` — 44/44, including T11/T11b.
- `actionlint` on `reusable-release.yml` and `web-platform-release.yml`; extracted `run:` snippets
  checked with `bash -c`, never `bash -n` on the YAML.
- Both exit-gate runners: `scripts/test-all.sh` and
  `apps/web-platform/infra/run-registered-suites.sh` (the diff touches
  `apps/web-platform/infra/`, which `test-all.sh` does not cover — it says so in its own preamble).
- AC1–AC18 walked presence-first with `-F` and a missing-anchor guard.

11 scenarios are enumerated in the plan's Test Scenarios section, covering each `mirror_reason` arm,
the digest-mismatch case, the plugin-release skip path, the empty-reason override, and both detector
verdicts.

## Third review round — what it found in the first two rounds' output

A focused 5-agent pass over the commits that had had no independent eyes. It found
defects in the fixes, which is the point of running it:

- **The release preflight could not report.** `set -uo pipefail` does not clear `-e`, and
  Actions runs `run:` under `bash -eo pipefail {0}`, so the bare script invocation aborted
  the step and `rc=$?` never ran — taking the whole three-valued `case` with it. Measured:
  the step printed nothing and exited 1, and `continue-on-error` hid the abort. The step
  whose stated purpose is naming the cause up front was silent on both paths it exists for.
- **The token-drift email asserted a cause on runs where nothing was measured.** One static
  body claimed a token "is no longer accepted by Cloudflare" and said to set a live value —
  false, and the wrong action, for every UNVERIFIABLE-only and every could-not-run outcome.
  The detector's own report block condemns exactly this. Now three verdict-specific bodies,
  fed by a new `--json-file` so one scan serves both the log and the caller.
- **ADR-096 cited two CLOSED-NOT_PLANNED issues as the tracker** for the open architectural
  debt, one of them unrelated to pull credentials. The debt has no tracker; it now says so.
- **ADR-096 misidentified the revoked credential** as tied to a *human* account when it was a
  machine account — which inflated the relaxation condition in clause (d)(1).
- **A blanket "the rest of this bullet remains current"** recertified ~110 lines on the
  strength of two, one paragraph after the text warning about exactly that move.
- **The orphaned-tail class recurred a third time**, including on two claims that authorize
  irreversible destroys, and `model.c4`'s `zotRegistry` node still advertised a GHCR fallback
  while its sibling `ghcr` node was marked DEAD in the same enumeration.
- **The `.gitleaks.toml` `$`-anchor claim was falsified.** The anchor bounds the token
  *gitleaks captured*, not the credential on the line, so material appended after a
  non-`[A-Za-z0-9_-]` delimiter is still silenced. `regexTarget = "line"` closes it and was
  rejected after measuring why: `Line` is null in `gitleaks git` diff mode, which is the mode
  the PR range scan runs, so it would disable the carve-out entirely. Recorded as a measured
  residual gap rather than papered over.
- **Nothing pinned the step order** — and the defect this PR fixed was an ordering defect.
  Added `terraform-drift-step-order.test.sh`, mutation-proven on all three assertions.

Also fixed: the reporting steps were still shadowing *each other* (neither issue-filer carried
`continue-on-error`, so an API failure skipped the drift email by implicit `success()`) in both
jobs; and a surviving twin of a comment whose sibling copy was deleted.

**Two things this PR marks but does not fix**, because the premise it retracts is what makes
them dangerous:

- `scripts/registry-pull-path-health.sh` gates an irreversible destroy of the zot store on a
  premise that is now false, and one of its three ABORT operands can no longer fire. Header
  and PASS line now carry the retraction so a green PASS is not read as authorization.
- A fresh web host that misses the zot probe dies at `stage=pull`, which matches no Sentry
  alert rule. Fixing that is a terraform apply this PR does not perform.

**Net-issue-flow override — closing 0, filing 3 (net +3).** Each filing is justified
individually below; none is inlinable, and the alternative to filing is losing the item.

- **#7077 — inngest image.** A discovered defect in a *different subsystem*
  (`cloud-init-inngest.yml` hard-pins GHCR with no zot path, so its dedicated host is
  un-bootable now). The rule is explicit that a discovered defect in another subsystem stays
  its own issue and its own PR rather than being bundled into an unrelated branch.
- **#7078 — verify the gate on the first real release.** Genuinely un-automatable pre-merge:
  the faithful test *is* a production release. Enrolled as a follow-through so it cannot rot.
- **#7079 — consolidated residuals.** This is already the consolidation (it absorbs three
  separate plan items). Its largest component cannot be inlined at all: `hcloud_server.web`
  carries `lifecycle.ignore_changes = [user_data]`, so a `cloud-init.yml` edit is unverifiable
  without a host reprovision this PR does not perform.

Worth stating plainly rather than hiding behind the override: the queue growth this gate
exists to catch is real, and the deepest item here — **production has one registry and no
fallback** — is *unowned*. ADR-096 clause (g) claimed two trackers for the restoration paths;
both were closed NOT_PLANNED three weeks ago, and this PR corrects that claim rather than
papering over it. The debt is now recorded as unowned, which is the honest state.


deruelle:
## Overnight status — plan phase complete, implementation not started

Ran `plan` (+ its mandated review panel). **`deepen-plan`, `work` and `review` have NOT run.** No
product code was written; the diff is planning artifacts only. The stop was chosen, not a bypass —
the lifecycle skills were invoked directly rather than via `/one-shot` precisely so no merge to prod
could land unattended.

**Why the plan phase absorbed the budget:** the investigation overturned two of the three stated
premises, and the review panel returned **5 P0s** — three of which invalidated the plan's own design
or evidence. Consolidating that was worth more than starting code on a wrong design.

### The headline finding: the empty-200 was a red herring

`https://registry.soleur.ai/v2/` returning HTTP 200 with an empty body is **correct behaviour**, not
a broken origin. That hostname's tunnel ingress is `service: tcp://10.0.1.30:5000` — **TCP-mode** —
so a plain HTTPS GET is not a WebSocket upgrade for that stream and nothing is proxied. The repo
already documents this in the bridge action's own header.

The private-net route is present and healthy. Verified end-to-end, read-only:
`cloudflared access tcp --hostname registry.soleur.ai --url 127.0.0.1:15000`, then `GET /v2/` →
**HTTP 401 from zot itself** (`Www-Authenticate: Basic`, zot-shaped JSON). That is the origin
answering. `model.c4:428` already records the #6416 route hypothesis as closed.

**The actual cause**, from the bridge log recovered out of run `30468080168`'s teardown step:

```
2026-07-29T16:01:14Z ERR failed to connect to origin error="websocket: bad handshake"
```

`github-actions-registry-push` was **created at 15:57:52Z** (live CF API) — recreated by the Terraform
apply in #7065's own release. The bridge ran 3 minutes later with the stale Doppler value, so CF
Access refused the upgrade. `nc -z 127.0.0.1 5000` passed throughout, because cloudflared opens its
local listener before the far end is proven reachable — a false gate. Your #7067 commit already names
`REGISTRY_PUSH_ACCESS_TOKEN_*` as exactly this class. The token in Doppler works now.

**Consequence: Deliverable B needs no Terraform, no Cloudflare write, and no Doppler write.** Nothing
is staged for you to apply. It became a detection-and-truthfulness fix.

### What the panel caught that I had wrong

1. **The deploy was never gated on the release.** I listed "a failed release job blocks deploy" as
   *confirmed*, citing `needs: [release]`. Wrong — `deploy`'s condition leads with `always() &&`,
   which discards skip-on-failed-needs, and then gates on the `docker_pushed` output set ten steps
   before the assertion. `needs.release.result` appears **0 times**. Unfixed, the gate could have
   protected nothing while migrations still applied, leaving prod on new schema + old code.
2. **My separate verify step broke on the exact incident it targeted.** On a bridge failure
   `degraded()` exits before `install_crane`, so the step would find no `crane` and no zot login — and
   my error table would have told you *"CI regression, not a registry problem."* The assertion now
   lives inside the mirror step; the cost of doing it honestly was two expected-rc literals in a test.
3. **My truthfulness sweep missed the document you'd actually read.** `zot-registry-revert.md` tells
   you, during an outage, to fall back to GHCR and calls it *"always warm and current."* That's false
   now, and it's instructions rather than a comment. ADR-096's own "Instant revert" bullet and the
   AP-016 register row say the same thing.
4. One of my error labels named a cause it could not measure, and would have sent you to wait out a
   Sigstore outage that wasn't happening — the same defect this work exists to remove.

### Three decisions waiting on you

In `knowledge-base/project/specs/feat-one-shot-zot-mirror-fail-closed/decision-challenges.md`. All
three cut or move scope you asked for, so I did not decide them:

- **UC-1** — drop the new bridge probe from this PR? Three reviewers said yes: it's the only change
  with zero runner-side evidence, and the bridge action is shared by **three** workflows, one of which
  would get *quieter*.
- **UC-2** — move the token-drift detector fix to a follow-up ("finish #7067")?
- **UC-3** — split gating from non-gating into two PRs?

### Verification

```
git log --oneline origin/main..HEAD
gh pr view 7071 --json isDraft
curl -s https://app.soleur.ai/health | jq '{status,version}'   # must still be v0.244.0
```

Prod untouched: no deploy, no apply, no secret write, no merge. PR stays in draft.

PR #7127, opened 2026-07-31
fix(infra): the drift detector graded the CI-SSH token DEAD forever, and 200 was unreachable

deruelle:
## What was wrong

`scripts/check-cloudflare-token-drift.sh` graded every Cloudflare Access service token LIVE **only on HTTP 200**.

That rule fits `registry.<base>` — its tunnel ingress is `tcp://10.0.1.30:5000` fronting zot, which speaks HTTP over the raw TCP forward.

It cannot fit `ssh.<base>`. That ingress is `ssh://<web-1-ip>:22` (`apps/web-platform/infra/tunnel.tf`), and sshd answers an SSH version banner, not HTTP. A request Access **admits** is handed to sshd, and cloudflared returns 5xx.

So **200 is unreachable for a perfectly healthy credential**, and `CI_SSH_ACCESS_TOKEN_*` graded DEAD on every run regardless of its real state — a permanent false positive driving a twice-daily *"[TOKEN DRIFT] rotation did not propagate"* email whose prescribed remediation is to overwrite a healthy secret.

## How it was found

While diagnosing the genuinely-dead CI-SSH token, the detector's `31 live, 1 dead` output was being cited as evidence. It isn't evidence — that row is structurally guaranteed for this token family whatever the credential's state. The real discriminator was an independent `curl` returning 403 with real header values.

## The fix

`access_origin_kind_for()` declares what each mapped origin speaks. A non-HTTP origin is graded on **Access admission** — the strongest claim that is true of it:

| code | verdict | why |
|---|---|---|
| `403` | DEAD | only Access can emit it; an ssh origin produces no HTTP status at all |
| `5xx` | LIVE | Access admitted; cloudflared then failed to speak HTTP to sshd — healthy |
| `000` | UNVERIFIABLE | no definite answer (timeout/DNS/TLS) — learned nothing |

Fail-closed is **preserved, not traded away**: `000` renders UNVERIFIABLE and still exits non-zero. It is deliberately kept out of `DEAD_ROWS`, because "measured nothing" and "Cloudflare rejected this" have opposite remedies — and the DEAD heading tells the operator to overwrite the secret.

The HTTP arm is byte-identical.

## Verification

- 3 new cases, written **RED first** (T15/T17 failed as predicted; T16 passed, proving the rejection signal already worked and must not regress).
- **Mutation-tested both directions**: reverting the opaque arm reds T15/T17; grading 403 as LIVE reds T16.
- Suite: 19/19.
- **Real production run**: still reports the CI-SSH token DEAD — it genuinely is (403). This does not turn a real failure green; it makes a *repaired* token reportable, which it previously could not be.
- `shellcheck` clean (the one SC2016 info is pre-existing on `main`).

## Scope

Deliberately standalone. The CI-SSH token is still dead and its `-replace` repair path is separate work on `feat-one-shot-ci-ssh-token-replace`; this PR only stops the detector from lying about it.

### DIFF
--- a/scripts/check-cloudflare-token-drift.sh
+++ b/scripts/check-cloudflare-token-drift.sh
@@ -282,7 +282,13 @@ UNVERIFIABLE_ROWS=()
 declare -A PAIR_VERDICT=()   # "host|id|secret" -> LIVE|DEAD (verify each distinct pair once)
 
 verify_access_pair() {
-  local host="$1" id="$2" secret="$3" kind="${4:-http}" cache_key="$1|$2|$3"
+  # `kind` is part of the cache key: the same (host,id,secret) triple graded under two
+  # different origin kinds is two different questions with two different answers, and
+  # omitting it would let whichever base was scanned first (bases are iterated in `sort`
+  # order) silently decide the verdict for the other. No two bases map to one host today —
+  # this keeps that from becoming a silent miscarriage the day one does.
+  local host="$1" id="$2" secret="$3" kind="${4:-http}"
+  local cache_key="$1|$2|$3|$kind"
   if [[ -z "${PAIR_VERDICT[$cache_key]:-}" ]]; then
     local code
     code=$(curl -s -o /dev/null -w '%{http_code}' --max-time 20 \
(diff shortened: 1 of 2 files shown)


deruelle:
## ⚠️ HOLD — the premise of this PR is unverified, and new evidence undercuts it

Adversarial review + a history check found the load-bearing claim here is not established.

**1. The "permanent false positive" framing is wrong.** The entire Access service-token arm — enumeration *and* the `ssh.soleur.ai` mapping — was added by `106c671f6` (#7071), merged **2026-07-30T16:30**. The drift check first went red at **18:00 the same day**: the first run after that arm shipped. There is no history of this token grading LIVE, because the probe did not exist. The PR body's "twice-daily email over a permanent false positive" is unsupported.

**2. The `5xx => LIVE` arm has never been observed.** No 5xx from `ssh.soleur.ai` has ever been seen — the production run reports 403, and the tests only replay a hardcoded `MOCK_HTTP_CODE`. The arm rests on a prediction about what cloudflared returns for an HTTP GET to an `ssh://` ingress.

**3. The same file argues the opposite.** `check-cloudflare-token-drift.sh` documents the sibling `tcp://` registry ingress: *"a plain HTTPS GET is not the WebSocket upgrade that stream needs and nothing is proxied. Cloudflare answers on its own"* — with an empty **200**. If `ssh://` behaves the same, 200-only was correct, there is no bug, and this change would grade a healthy token UNVERIFIABLE.

**4. `5xx` is not proof of admission anyway.** Cloudflare's WAF / rate-limiting / challenges run *ahead* of Access and have historically been served as **503** — a rejection landing in the `5??` arm and grading LIVE, silently. That is the exact failure class the second commit was written to prevent.

### Also confirmed (real, but separable)

- `verify_access_pair` is called inside `$( )`, so every `PAIR_VERDICT[...]` write is discarded — the cache has never worked. Pre-existing; my `kind`-in-cache-key change protects nothing.
- `curl ... || printf '000'` yields `000000` (curl writes the `-w` template *and* exits non-zero), so the documented `000` case cannot occur.
- The UNVERIFIABLE report header and operator email say "no probe configured … do NOT rotate", which is wrong for an *inconclusive probe* — it would tell an operator not to rotate a genuinely revoked token.
- T18 is the weakest assertion in the new block despite being the flagship false-LIVE guard; it passes on `rc=2` and on empty output.

### Disposition

**Not merging.** The premise is only testable once a working credential exists, so the sound order is the reverse of what I proposed: repair the token first, then observe what an admitted request actually returns on this host, then decide whether the detector needs any change at all.

Kept as a draft for the finding trail and the separable defects above.

PR #7151, opened 2026-08-01
WIP: feat-one-shot-ugrep-oom-guards

deruelle:
Draft PR created automatically. Content will be added as work progresses.


deruelle:
Closing unmerged. Both guards in this PR were broken, and an 8-agent review established it after the branch was green everywhere.

**State at close:** 74/74 and 26/26 suites, two mutation batteries at 10/10 and 7/7 killed, shellcheck clean, 243/243 full suite, `[OK] B_ALWAYS=43189`. None of that was evidence the guards worked.

## Guard 1 — the cost model was refuted three times

Each model was derived from measurement, which is exactly why each felt settled. Each was measured against fixtures its own author chose.

| # | Model | Refuted by |
|---|---|---|
| 1 | bounded repeat AND alternation | Wrong on both sides (caught at plan time) |
| 2 | ∏(upper+1) ≥ 500 | Denies UUIDs, ISO timestamps, SHA-256 pairs and the conflict-marker regex *inside `guardrails.sh` itself*; allows a real blowup at product 289 — including this plan's own "highest observed-safe 441" datapoint, which does not reproduce |
| 3 | ∏(upper+1) ≥ 150 over a *wide* atom | Positive classes of any width blow up when the literal between the repeats is reachable |

The third ends the approach. Same pattern, same bound product; only the corpus differs:

| Pattern | Literal present? | Peak RSS |
|---|---|---|
| `[0-9]{0,80}x[0-9]{0,120}` | no | 7.3 MB |
| `[0-9]{0,80}5[0-9]{0,120}` | **yes** | **1.67 GB — killed** |
| `[a-z]{0,80}q[a-z]{0,120}` | yes | **killed** |
| `[[:print:]]{0,80}q[[:print:]]{0,120}` | yes | **killed** |

The first row is fixture G21 — the evidence used to prove narrow classes are cheap. It measured cheap only because `x` never occurs in the subject, so ugrep prefilters on the literal and never builds the DFA. Every "cheap" datapoint in both calibration ladders shares that confound.

Cost is class size × bound magnitude × literal reachability against a file the hook has not read. That is a missing variable, not a threshold to retune.

Guard 1 additionally left seven forms ungated, all of which reach the shim: `X=$(grep …)`, backtick substitution, `(grep …)` subshell, `G=grep; $G …`, `P='…'; grep "$P"`, `eval "grep …"`, `grep -f patternfile`. Command substitution is the most common agent idiom there is.

## Guard 2 — never executed, once

Committed `100644`; `settings.json` invokes the path directly, so every SessionStart died at `exec` with 126. Both sibling SessionStart hooks are `100755`.

Nothing could catch it: the suite runs `bash "$HOOK"`, which ignores the mode bit, and so does `test-all.sh`. The live AC7a verification — a real cgroup created, `memory.max` read back at 12 GiB — went through that same path, one production never uses.

When run manually it was wrong four more ways: it escaped the terminal's kill switch (sibling scope, so closing the terminal no longer reaps a runaway), capped at 12 GB when the desktop froze at **9.5 GB**, left `memory.swap.max` unbounded (swap exhaustion being the incident's observed symptom), and had no aggregate bound across sessions.

## Disposition

- **#7163** — lands the diagnosis chain and every measurement, including the ones that refuted the author's own models. That is the part worth keeping.
- **#7165** — the design that should have been chosen: `updatedInput` rewriting `grep` → `command grep`, which sidesteps classification entirely.
- **#7166** — the memory backstop done properly, on `StartTransientUnit` + a shared slice.
- **#7164** — a confirmed pre-existing RCE in the hook input parsing, surfaced by this review and independent of it.

## Operator exposure

**Unchanged.** No guard is in place and none was ever in place — Guard 2 never ran, and Guard 1 never merged. The live cgroup created during verification has been removed and the agent process returned to the terminal's scope.

PR #7158, opened 2026-08-02
WIP: feat-one-shot-resend-inbound-webhook-500

deruelle:
> # ⛔ DO NOT MERGE
>
> Review finding 2 is **2 of 4 steps** complete. The merge block stands until it finishes.
> Full status: `knowledge-base/project/specs/feat-one-shot-resend-inbound-webhook-500/review-findings.md`.
>
> This PR additionally carries a gate that **has never been observed passing** (the zot
> consumer-probe extension, below). Merging it today would re-provision the live web-1 host
> with a probe whose success condition is unverified. Steps 3–4 are blocked on **#7242**
> (zot bridge down since 2026-08-03).

Fixes the Resend inbound-webhook 500s / Inngest dispatch outage. Incident record: #7228.

## Status

Seven of eight review findings are applied. Finding 2 (the paused operating point is not
reproducible from IaC) is in progress at 2 of 4 steps:

1. ✅ Budget freed for the digest pin.
2. ✅ Mirror made digest-preserving — split out and merged as #7203.
3. ⛔ Backfill `v1.1.24` and prove parity — blocked on #7242.
4. ⛔ Pin both refs, then flip `web_colocate_inngest` — invalid until step 3 is green.

## Step-4 prerequisites (2026-08-04)

A green `mirror_only` run does **not** license step 4: it proves zot serves the digest to the
**push** credential over the tunnel from a runner, not that the **host** can pull it
(`ZOT_PULL_*`, private NIC), and zot `accessControl` can grant push-read without pull-read.
Since the GHCR pull PAT was revoked 2026-07-30, zot is the sole pull path, and `cloud-init.yml`'s
web arm runs under `set -e` with a failing `trap … EXIT` — so a digest zot cannot serve is a fatal
fresh boot on the only live web origin, not a graceful fallback.

**(a) Host-side pull proof — AUTHORED, NOT PROVEN.** `web-probe.tf`'s `zot_probe_repo` derived
from `var.image_name`, so the probe covered `soleur-web-platform` only; nothing verified the
bootstrap repo. `ZOT_PROBE_REPO` now takes a comma-separated list, and aggregation is
**all-must-serve** so a servable web repo cannot mask an unservable bootstrap repo (the #6400
class). `zot-entry-gate.sh` was deliberately not resurrected — it probes over the tunnel with
Doppler prd creds, so it cannot demonstrate host-side pullability regardless of its known
false-gate defect; that decision is tracked in #7258.

[...]

Issue #6274, opened 2026-07-09
infra(registry): GHCR→zot mirror step fails (connection reset), reds every release after a successful deploy


deruelle:
## Symptom

The `web-platform-release` → `release / release` job's **"Mirror image GHCR→zot (crane) + cosign-sign the zot digest"** step fails:

```
Error: Patch "http://127.0.0.1:5000/v2/jikig-ai/soleur-***/blobs/uploads/<uuid>":
  readfrom tcp 127.0.0.1:43078->127.0.0.1:5000: write: connection reset by peer
##[error]Process completed with exit code 1.
```

Run: https://github.com/jikig-ai/soleur/actions/runs/29015959409 (release/release job 86110882010).

## Impact

**Not a user-facing outage.** The primary deploy is unaffected — `deploy` + `live-verify` jobs SUCCEEDED (prod pulled the image from GHCR; `/health` = 200; the newly-shipped feature is live). Only the *secondary* GHCR→zot mirror copy failed. But the `release / release` job concludes `failure`, so **every release run reds** until the zot registry side is healthy.

## Likely cause

`connection reset by peer` mid blob-upload to the self-hosted zot registry (127.0.0.1:5000 via the cloudflared CF-Access tunnel). Consistent with the recent zot-disk-full work (#6238 registry-host-replace, #6240 zot-disk-full-observability) — the zot host may be disk-full again or the tunnel/token flaked.

## Ask

Infra triage: confirm zot host disk + cloudflared `registry_push` token/tunnel health; make the mirror step's failure non-release-blocking (or self-heal) so a transient zot hiccup doesn't red every release after a successful deploy.

Discovered post-merge during PR #6239 (feat-beta-crm-ui) ship — orthogonal to that feature.

PR #7203, opened 2026-08-03
fix(ci): the one image we exec as root was the only mirror not preserving digests

deruelle:
## Why

`build-inngest-bootstrap-image.yml` mirrored GHCR→zot with `docker tag` + `docker push`. Docker re-uploads and re-creates the manifest, so **zot served a different digest than GHCR for the same tag** — silently, because nothing compared them.

That is load-bearing rather than cosmetic, because `cloud-init.yml`'s web-host arm substitutes the zot ref *into* `$IREF` and then execs that payload **as root**:

```sh
ZIREF="$ZURL/jikig-ai/soleur-inngest-bootstrap:v1.1.24"
if docker pull "$ZIREF"; then IREF="$ZIREF"; ... fi
docker pull "$IREF"
```

So the root-executed path resolved through a mutable tag, and pinning only the GHCR literal would not close it — the zot branch overwrites the very variable being pinned. ADR-155 named this as the blocker for its Alternative C, and #7144 finding 2 inherits it as the precondition on flipping `web_colocate_inngest`.

The sibling mirrors were already correct: `reusable-release.yml` (app image) and `build-inngest-config-bundle.yml` (config bundle) both use digest-preserving `crane copy`, on the same SHA-pinned crane `v0.20.2` adopted here. **The only mirror still on `docker tag + push` was the one whose payload runs as root on the scheduler.**

## What changed

1. **`crane copy` instead of `docker tag` + `docker push`.** Registry-to-registry, so the manifest bytes are preserved and one `@sha256` resolves on both registries.

2. **Digest parity is asserted, not assumed.** The step compares `crane digest` on both refs. An empty source digest counts as failure — the capture uses `|| true`, so an auth fault yields `""`, and `"" == ""` would otherwise read as a match and certify nothing.

3. **A `mirror_only` dispatch input.** Because crane reads the published manifest, a backfill no longer needs a rebuild. This matters more than it sounds: the existing `workflow_dispatch` path **rebuilds and re-pushes the same tag, so the tag's GHCR digest moves** — and `cloud-init-inngest.yml` pins the dedicated host's payload by that digest. Using the existing dispatch to "re-mirror" `v1.1.24` would have invalidated a live pin. `mirror_only` skips the build entirely.

4. **The degrade contract inverts under `mirror_only`.** On the build path the mirror is a secondary copy and must never red a release. In a backfill the mirror *is* the run, so `degraded()` exits 1 and `continue-on-error` is off. An exit 0 there would report success for a backfill that did not happen, and the caller would then pin a digest zot cannot serve.

## Verification

- `actionlint` clean on the changed steps. The one `SC2015` note is pre-existing on `origin/main`, in the `set +e` teardown.
- `cutover-inngest-workflow.test.sh` 334/334, `inngest-host.test.sh` 41/41, `cloud-init-inngest-bootstrap.test.sh` 80/80.
- GHCR digest for `v1.1.24` independently confirmed as `sha256:6cdaa63d1496642e681898a831234b712f75d3b09bd0844bcabec3de74b0a0f8` from two sources (GitHub packages API, `crane digest`), matching the pin already in `cloud-init-inngest.yml`.
- Zot's current digest could not be read from a workstation — it is dark behind the CF Tunnel (`/v2/` → HTTP 000). The new parity assertion measures it in CI, which is the right place for it. If the digests already agree, the backfill is a no-op and the assertion passes.

## After this merges

Dispatch the backfill from `main`, then verify parity:

```
gh workflow run build-inngest-bootstrap-image.yml \
  -f ref=vinngest-v1.1.24 -f mirror_only=true
```

Only once that is green can `cloud-init.yml` pin both refs and `web_colocate_inngest` flip to `true` (#7144 finding 2, tracked on the feature branch).

## Changelog

Fixed: the inngest-bootstrap zot mirror now preserves image digests (`crane copy`), so the root-executed bootstrap payload can be pinned by `@sha256` on both registries. Adds a rebuild-free `mirror_only` backfill mode and a digest-parity assertion.

Ref #7144

---

**net-issue-flow override.** #7228 was not filed by this PR. It is the incident-tracking record for the sibling branch `feat-one-shot-resend-inbound-webhook-500` (PR #7158), filed from that branch's `tasks.md` task 0.1, and it references this PR only to record a blocking dependency ("finding 2 at 2 of 4 steps, blocked on #7203"). The gate attributes it here because it was created after this PR and bare-references the number; that is a contextual-citation false positive, not a filing. Option (d) does not apply — no `[mandates-filing]` rule forced it, and guessing a rule id would be worse than saying so.

### DIFF
--- a/.github/workflows/build-inngest-bootstrap-image.yml
+++ b/.github/workflows/build-inngest-bootstrap-image.yml
@@ -19,12 +19,23 @@ on:
   # for a CVE base-image rebuild of an already-released version). It takes the
   # tag name as `ref` — NOT a free-form version — so it can never mint a
   # tagless image that blinds the cloud-init drift-guard (#4692 / #4676 AC6).
+  #
+  # CAUTION on the default (rebuild) path: it re-runs `docker build` and pushes the SAME
+  # tag, so the tag's GHCR digest MOVES. Anything pinning that tag by @sha256 —
+  # cloud-init-inngest.yml pins the dedicated host's payload — is pinned to bytes the tag
+  # no longer points at. Use `mirror_only` when the goal is only to get an already-published
+  # tag onto zot; it cannot move a digest because it never builds.
   workflow_dispatch:
     inputs:
       ref:
         description: 'Existing vinngest-vX.Y.Z tag to (re)publish (e.g., vinngest-v1.1.11)'
         required: true
         type: string
+      mirror_only:
+        description: 'Backfill only: crane-copy the EXISTING GHCR manifest to zot, no rebuild (digest-preserving)'
+        required: false
+        default: false
+        type: boolean
 
 permissions:
   contents: read
@@ -144,7 +155,12 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # Skipped entirely under `mirror_only`. Not an optimisation: a rebuild re-pushes the
+      # same tag and MOVES its GHCR digest, which is the one thing a digest backfill must
+      # not do. `crane copy` reads the published manifest from GHCR, so the local image this
+      # step would produce is not needed for the mirror.
       - name: Build + verify + push
+        if: ${{ !inputs.mirror_only }}
         env:
           IMAGE: ghcr.io/jikig-ai/soleur-inngest-bootstrap
           TAG: ${{ steps.tag.outputs.tag }}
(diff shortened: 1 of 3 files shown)


deruelle:
## Review brief — 5 items, not 4

Rebased onto main (was 14 behind) and pushed `da7cb63b6`. The original four review points stand as written; a **fifth** was found by self-review and fixed in the second commit.

### 5. A soft-failed bridge skipped the mirror and still reported the run green — **FIXED in `da7cb63b6`**

The first commit dropped the `continue-on-error` belt on the **mirror** step under `mirror_only`. It left it on the **bridge** step above, which re-opens the identical hole one step earlier:

```yaml
- name: Bridge to zot registry (CF Tunnel)
  id: zot_bridge
  continue-on-error: true          # ← was unconditional
- name: Mirror inngest image GHCR→zot
  if: steps.zot_bridge.outcome == 'success'
```

A soft-failed bridge **skips** the mirror and leaves the job **green**. The belt-off on the mirror alone cannot close it — a skipped step never reaches its own degrade path, so `degraded()`'s `exit 1` is unreachable in exactly the case it exists for.

This is load-bearing for #7144 finding 2: step 3 asserts *"the run is green"* before step 4 adds `@sha256` to `cloud-init.yml`. This was the precise path by which a green run could license a pin to a digest zot cannot serve.

Fix mirrors the existing pattern: `continue-on-error: ${{ !inputs.mirror_only }}`. Belt stays **on** for push/rebuild, where the bridge is genuinely secondary to the GHCR release (#6274).

### The original four

| # | Item | Where |
|---|---|---|
| 1 | `crane copy` replaces `docker tag`+`push` | `crane` SHA-pinned v0.20.2, same version as the two sibling mirrors |
| 2 | Parity assertion treats an empty source digest as failure | `-z "$SRC_DIGEST"` — the capture uses `\|\| true`, so `"" == ""` would otherwise certify nothing |
| 3 | `mirror_only` skips the build | a rebuild re-pushes the same tag and **moves** its GHCR digest, invalidating `cloud-init-inngest.yml`'s live pin |
| 4 | `degraded()` exits 1 under `mirror_only` | on the build path the mirror is secondary; in a backfill the mirror **is** the run |

### Coverage added

`apps/web-platform/infra/inngest-bootstrap-mirror-only.test.sh` — 25 assertions covering all five items, registered in `infra-validation.yml` (88 derivable suites, zero orphans).

Every structural assertion **parses the file as YAML**. A grep would pass vacuously: both the workflow's header and the suite's own prose discuss `continue-on-error`, `mirror_only` and `crane copy` at length (`cq-assert-anchor-not-bare-token`). Exact-string equality throughout — the real risk on each is **operand inversion**, invisible to a substring check while flipping the meaning. The behavioral leg **executes** the workflow's own extracted `degraded()` under both `MIRROR_ONLY` values rather than modelling it.

**Mutation-proven — 8 mutations, each verified applied by `diff` before running.** Reverting the bridge belt-off, inverting the build-step guard, dropping the `-z "$SRC_DIGEST"` guard, removing `degraded()`'s `exit 1` arm, `crane copy` → `docker push`, `mirror_only` typed as string, and writing `mirror_status=ok` before the parity check all turn the suite RED.

### Verification

- Registered infra suites **88/88**, exit 0.
- actionlint unchanged — the single SC2015 note is pre-existing on `origin/main` (verified by linting main's copy of the file).
- GHCR digest for `v1.1.24` re-confirmed unchanged: `sha256:6cdaa63d1496642e681898a831234b712f75d3b09bd0844bcabec3de74b0a0f8`.

### Still needs you

Not marked ready, not merged. This holds `DOPPLER_TOKEN_PRD` and publishes the image the scheduler execs as root. **#7144 finding 2 steps 3–4 are blocked on this merging** — the `mirror_only` backfill of `vinngest-v1.1.24` must be dispatched from `main`.

Nothing here pins anything yet.

deruelle:
## Sibling-mirror sweep — answered, no change needed here

The open question from the brief was whether the bridge belt-off should be swept across the sibling mirrors. Checked both, read-only:

**`reusable-release.yml` — already fixed, and fixed better.** Its `zot_mirror` step is gated on `docker_build`, **not** on `zot_bridge`; it reads `BRIDGE_OUTCOME` internally and `degraded()` exits non-zero with no `continue-on-error`. The step comment names this exact failure mode ("the bridge could fail on every single release, zot would never receive an image, and the release stayed green and silent") and records why a sibling step can't fix it — step outputs are namespaced by step id. Nothing to sweep.

**`build-inngest-config-bundle.yml` — has the literal shape, but the belt is correct there.** `zot_bridge` is unconditionally `continue-on-error: true` and the mirror carries `if: steps.zot_bridge.outcome == 'success'` + `continue-on-error: true`. But that workflow has **no `mirror_only` mode** (single `workflow_dispatch` with a `version` input), so the mirror is never the deliverable and GHCR stays authoritative for the host pull. Belting off would be wrong.

It does still carry the weaker half of the same root cause: a bridge failing on every dispatch means zot never receives the bundle, no output is emitted, and the run is green and silent. Filed separately as #7243 rather than grown onto this PR — different workflow, different review surface, and strictly weaker (silence, not a false digest pin).

**Bottom line: this PR's scope is right as-is.** No sweep commit needed.

---

Also re-confirmed the pin target live, unchanged:

```
ghcr.io/jikig-ai/soleur-inngest-bootstrap:v1.1.24
→ sha256:6cdaa63d1496642e681898a831234b712f75d3b09bd0844bcabec3de74b0a0f8
```

Note for whoever runs the backfill: the **image** tag is `v1.1.24`; `vinngest-v1.1.24` is the **git** tag and is the correct `-f ref=` input. Querying the registry for `vinngest-v1.1.24` returns 404 — that is not a moved digest.

Issue #6897, opened 2026-07-24
encryption-posture: lower-severity items (superseded plaintext volumes, zot HTTP link, legal-doc reconcile)


deruelle:
Consolidated lower-severity encryption-posture items from the 2026-07-23 audit (`knowledge-base/engineering/architecture/encryption-posture-audit-2026-07-23.md`):

- [ ] **Superseded plaintext volumes still declared** — `hcloud_volume.workspaces` (`apps/web-platform/infra/server.tf:1569`) and `hcloud_volume.git_data` (`apps/web-platform/infra/git-data.tf:196`) are `format = "ext4"`; web-1 `/mnt/data` now runs on the LUKS `workspaces_luks` mapper (cutover certified 2026-07-23, verify run 30040444418). Confirm these are detached and remove them, or ledger their live attachment.
- [ ] **zot registry link is plain HTTP by design** (`10.0.1.30:5000`, private net; integrity via cosign digest-pinning) — a real `cert_verification: off` connection exception; keep its ledger exception current.
- [ ] **Legal-doc reconciliation** — run `/soleur:legal-audit` against the audit's measured postures so `docs/legal/{privacy-policy,gdpr-policy,data-protection-disclosure}.md` encryption claims are substantiated, not asserted.

Ref #6588

deruelle:
**Verification + audit-trail (PR #6918 — this issue stays OPEN).**

- **Checkbox 1 (superseded plaintext volumes):** VERIFIED ledgered & current. `hcloud_volume.workspaces` (`server.tf:1569`) and `hcloud_volume.git_data` (`git-data.tf:196`) are recorded as `plaintext-exception` rows in `scripts/encryption-posture-ledger.json` with accurate evidence + `reevaluate_when` triggers. `git_data` stays as the rollback backstop pending DL-2; no removal (out of scope — zero live-infra mutation). Layer A PASS.
- **Checkbox 2 (zot HTTP link):** VERIFIED — the `cert_verification: off` connection exception is present + current.
[...]

Issue #6808, opened 2026-07-20
flag(infra): WORKSPACES_LUKS_HEARTBEAT_URL unwired after the web-1 cutover — a dead luks-monitor probe cannot page


deruelle:
Observed in `workspaces-luks-verify` run 29783424497 (2026-07-20), immediately after the live `/workspaces` LUKS cutover on web-1 landed:

```
[luks-monitor] WARN: WORKSPACES_LUKS_HEARTBEAT_URL absent — heartbeat not pushed (operator wires it at cutover)
[luks-monitor] OK: /mnt/data is LUKS-backed (device_type=crypto_LUKS mount_source=/dev/mapper/workspaces escrow=ok header=readable)
```

The LUKS probe itself is healthy. The problem is the **alerting path**.

## Why this matters now

The runbook (`workspaces-luks-cutover-6604.md` §Failure signals) names the Better Stack heartbeat as one of three off-host failure signals:

> **Better Stack** `betteruptime_heartbeat.workspaces_luks` — a missed daily push = a dead probe.

That detection depends on the heartbeat actually being pushed. With `WORKSPACES_LUKS_HEARTBEAT_URL` absent, the daily `luks-monitor` probe runs and succeeds but **pushes nothing** — so a *dead* probe and a *never-configured* probe are indistinguishable, and neither pages. The volume is now encrypted and load-bearing, so this is live exposure, not a pre-cutover nicety.

## The structural issue

The warning text says "operator wires it at cutover" — i.e. this is a deferred manual step. Per `hr-all-infrastructure-provisioning-servers` / `hr-fresh-host-provisioning-reachable-from-terraform-apply`, the heartbeat URL should be provisioned by the terraform apply that creates `betteruptime_heartbeat.workspaces_luks` and delivered to the host the same way `WORKSPACES_LUKS_KEY` is — not typed in by a human at cutover time.

## Acceptance

- [ ] `WORKSPACES_LUKS_HEARTBEAT_URL` is provisioned by terraform (sourced from the `betteruptime_heartbeat.workspaces_luks` resource, as a reference not a literal) and readable by the host
- [ ] `luks-monitor` pushes the heartbeat on a healthy probe; verified by an observed push in Better Stack
- [ ] the absent-URL case is an **error**, not a WARN — a monitor that cannot alert should fail loudly rather than log and continue green
- [ ] the runbook's §Failure-signals claim about the heartbeat is true rather than aspirational

[...]

Issue #7220, opened 2026-08-03
infra-config apply dies pre-logger on prod: exit_code=1 reason=unhandled files_total=0 after #7146


deruelle:
## Symptom

After PR #7146 merged (`701e76e6`), the auto-fired `Apply deploy-pipeline-fix` workflow **failed at its verify step**. The live infra-config channel on `soleur-web-platform` reports:

```json
{"schema_version":2,"start_ts":1785758310,"end_ts":1785758311,"exit_code":1,
 "reason":"unhandled","files_written":0,"files_failed":0,"files_total":0,
 "files":[],"restarts":[]}
```

The handler **ran and died in 1 second**, delivering zero of the 19 `FILE_MAP` entries. `reason:"unhandled"` is the EXIT-trap default — it died on an unhandled path, not through any of its own error branches.

Failing run: https://github.com/jikig-ai/soleur/actions/runs/30811367645

## Evidence chain (all measured, 2026-08-03)

**1. The webhook listener is UP — this is NOT the bricked-channel P1.**
`GET /hooks/infra-config-status` answered 200 on every attempt (three CI retries at 11:58:43/49/54, plus a manual pull at 14:04:55). The workflow's "channel is DOWN" error text does not apply.

**2. Prod IS running the new handler.** The terraform apply replaced both resources:

```
Apply complete! Resources: 2 added, 0 changed, 2 destroyed.
  terraform_data.infra_config_handler_bootstrap: Creation complete after 10s
  terraform_data.deploy_pipeline_fix: Creation complete after 1s
```

So #7146's `infra-config-apply.sh` changes cannot be excluded on "they never shipped".

**3. The handler emitted NOTHING to journald before dying.** A Better Stack query for `infra-config-apply` over the window returns zero rows from `soleur-web-platform`. This is a real absence, not a dead instrument — verified with a positive control in the same window:

```
[zot-probe] SOLEUR_PROBE_CANARY web-zot-consumer-probe source4_live=1 ts=1785765639
  host_name=soleur-web-platform
```

The R3 canary is shipping. So the handler died **before its first `logger` call** — i.e. before or at the very top of the script, after `START_TS` was taken.

[...]

PR #7266, opened 2026-08-04
fix(infra): make a luks-monitor that cannot report fail loudly (#6808 AC3/AC4)

deruelle:
The heartbeat limb of #6808 is now applied and live, so the `WARN` that described it as pending
becomes the error it always should have been.

## What was done out-of-band first (not in this diff)

AC1 was code-complete and unapplied. No `workflow_dispatch` target set in
`apply-web-platform-infra.yml` contains either resource — grep confirms both are absent from every
`-target` list — so the sanctioned path is the agent-run targeted apply under the
`OPERATOR_APPLIED_EXCLUSIONS` contract, not an operator checklist.

```
betteruptime_heartbeat.workspaces_luks          Plan: 2 to add, 0 to change, 0 to destroy
doppler_secret.workspaces_luks_heartbeat_url    value = betteruptime_heartbeat...url (a reference)
```

Then **unpaused via the Better Stack v2 API, not the UI** — `PATCH /api/v2/heartbeats/{id}`
`{"paused": false}`, confirmed by an independent `GET` (`paused=false`, `paused_at=null`).
Terraform's `lifecycle { ignore_changes = [paused] }` means later applies will not revert it.

**AC2, pulled rather than eyeballed.** The heartbeat went `pending` -> `up` on verify run
[30913027207](https://github.com/jikig-ai/soleur/actions/runs/30913027207), whose probe logged
`rc=0`, `workspace_count=8 expected=8`, and `OK: /mnt/data is LUKS-backed`. A transition to `up`
can only be caused by a received push, so this is not an empty result read as a pass. Negative
control: the `WORKSPACES_LUKS_HEARTBEAT_URL absent` WARN is gone from that run. Positive control:
the same string is present in run 29783424497, the run that opened the issue.

One credential note, recorded rather than left silent: the first heartbeat's push URL was printed
to a session transcript by a redaction pattern that did not match the real host. It was rotated
immediately via `-replace` of both resources — the old heartbeat (id 478793) now returns HTTP 404,
and the live secret was re-verified equal to the new heartbeat's URL by SHA-256 comparison rather
than by echoing either value.

## AC3 — both heartbeat arms are now fatal

| condition | outcome | why |
| --- | --- | --- |
| URL absent | `rc=3` `heartbeat_url_absent` | no retry — a config fault does not heal |
| push exhausted | `rc=3` `heartbeat_push_failed` | 3 attempts, so one transient blip is not an outage |

**`exit 3`, never `exit 1`.** Every at-rest assert has already passed by the time control reaches
this block, so routing an alerting fault through the at-rest exit would file the p0 `type/security`
verdict whose body asserts the published Article 32 claim is false. Both reasons are registered in
the workflow's `is_probe_integrity()` and its `rc=3` headline branch, and in the runbook triage
table — the `(M)`/`(N)` parity gates fail loudly if that ever drifts.

The suite had **no coverage of the heartbeat in either direction**, which is why the WARN could
never have been caught here. Added a positive control (a healthy probe really pushes) *before* the
two fatal arms, so neither can pass against a build that dropped the push entirely, plus an
assertion that a heartbeat fault never exits 1. The harness already modelled `MON_HB_URL` but
defaulted it empty — correct while absent was survivable, fatal to every healthy fixture now — so
that default flips and `MON_HB_PUSH_RC` is added.

## AC4 — the runbook claim is now true rather than aspirational

Un-struck the Better Stack entry, with a no-dashboard pull command and an explicit statement of
what a green heartbeat still does **not** cover: it is `policy_id`-gated on the paid tier so on the
free tier it emails rather than pages, and it has no host-side readyz/inventory visibility. The
scheduled verify stays beside it — it is no longer the only automatic check, but it remains the
only channel that can see a `workspace_count_shortfall`. The soak note flips from blocked to
started; the >=7d span is first satisfiable **2026-08-11**, which unblocks #6897.

## Known residual, stated rather than implied

`/usr/local/bin/luks-monitor` is baked by `workspaces-cutover.sh` (`install -D -m 0755`), while the
verify workflow ships a fresh copy of the script per run. So this AC3 flip binds the **verify path
immediately** and the host's **daily systemd unit only at the next cutover-channel re-bake**. The
host's daily probe keeps pushing the heartbeat either way — that code has been in `main` since
#6604 and only ever lacked the secret, which now exists.

## Verification

- `luks-monitor.test.sh` — 43 passed, 0 failed (was 36/3 mid-change; the 3 were this change, now covered)
- `workspaces-luks-verify-workflow.test.sh` — 121 passed, 0 failed
- `workspaces-luks-freeze.test.sh` — 91 passed, 0 failed
- `workspaces-luks.test.sh` — 26 passed, 0 failed
- `shellcheck -S warning luks-monitor.sh` — clean

Ref #6808 — deliberately never `Closes`. That issue's closure condition is a conjunction (heartbeat
wired **and** schedule exists). Both limbs are now satisfied, but closure is the operator's call
once the daily scheduled run and the host-side push are both observed in steady state.

### DIFF
--- a/.github/workflows/workspaces-luks-verify.yml
+++ b/.github/workflows/workspaces-luks-verify.yml
@@ -274,7 +274,7 @@ jobs:
           # future producer/consumer drift of exactly this kind.
           is_probe_integrity() {
             case "${1:-}" in
-              readyz_gate_regression|readyz_unparseable|readyz_unreachable|readiness_helper_unavailable|workspace_count_baseline_missing|workspace_count_unreadable|mapper_path_override_refused|doppler_unreachable)
+              readyz_gate_regression|readyz_unparseable|readyz_unreachable|readiness_helper_unavailable|workspace_count_baseline_missing|workspace_count_unreadable|mapper_path_override_refused|doppler_unreachable|heartbeat_url_absent|heartbeat_push_failed)
                 return 0 ;;
             esac
             return 1
@@ -493,6 +493,8 @@ jobs:
           # undone one layer up. `sed` reads the reason off the FAIL line the host already emitted.
           if [[ "$probe_rc" -eq 3 ]]; then
             case "$reason" in
+              heartbeat_url_absent|heartbeat_push_failed)
+                echo "::error::workspaces-luks ALERTING-PATH failure (rc=3, reason=$reason). Every at-rest assert PASSED — the volume is fine and this is NOT data loss. What failed is the probe's ability to REPORT: the Better Stack dead-probe heartbeat was not pushed, so from here on a dead probe and a healthy one look identical off-host (#6808). Do NOT run a data-recovery procedure. Fix: confirm doppler_secret.workspaces_luks_heartbeat_url is applied and that the prd_workspaces_luks boot token can still read it. Sentry: feature=workspaces-luks op=workspaces-luks-drift." ;;
               readyz_gate_regression|readyz_unparseable|readyz_unreachable|readiness_helper_unavailable|workspace_count_baseline_missing|workspace_count_unreadable|mapper_path_override_refused)
                 echo "::error::workspaces-luks PROBE-INTEGRITY failure (rc=3, reason=$reason). This run proves NOTHING about the volume — it is a routing/config/tooling fault, NOT data loss. Do NOT run a data-recovery procedure. Sentry: feature=workspaces-luks op=workspaces-luks-drift." ;;
               *)
--- a/apps/web-platform/infra/luks-monitor.sh
+++ b/apps/web-platform/infra/luks-monitor.sh
@@ -281,14 +281,43 @@ if [ "${LUKS_MONITOR_ASSERT_READYZ:-0}" = "1" ]; then
 fi
 
 # All asserts passed — push the heartbeat so a dead probe FAILS it (P1-4).
+#
+# #6808 AC3 — BOTH arms below are FATAL, not WARN. Until the 2026-08-04 apply, an absent URL logged
+# a WARN and exited 0, so a never-configured probe and a dead probe were indistinguishable and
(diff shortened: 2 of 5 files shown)


deruelle:
### On the red `Infra Validation / deploy-script-tests` check

It is **pre-existing and not caused by this PR**, and I would rather say that with evidence than let a red check ride quietly.

Proof: checked out this PR's exact merge base `d7eb6e6e8` in a clean worktree with none of these changes applied, and got the identical result — `144 passed, 2 failed`, the same two assertions.

```
FAIL: pre-fix handler carried NO fatal_line (this is #7220)
FAIL: pre-fix handler reported the hardcoded files_total=0
```

Root cause (full write-up on #7220): `infra-config-apply.test.sh:test_fatal_channel_red_against_main` pins its "pre-fix handler" to the **moving ref** `origin/main` rather than an immutable SHA. PR-A merged as `c2de2581e`, so `origin/main` now carries the fixed handler and the proof-of-red inverted the moment it landed. The fix is `origin/main:` -> `701e76e6b:`.

Not repaired here on purpose: #7220 has PR-B in flight, and folding an unrelated CI repair into a `type/security`-adjacent diff would make both harder to review. This PR touches none of those files — its diff is `luks-monitor.sh`, its harness and tests, the verify workflow, and the runbook.

Gates owned by this change, all green after rebase onto `d31d8a2c7`:

| gate | result |
| --- | --- |
| `luks-monitor.test.sh` | 43 passed, 0 failed |
| `workspaces-luks-verify-workflow.test.sh` | 121 passed, 0 failed |
| `workspaces-luks-freeze.test.sh` | 91 passed, 0 failed |
| `workspaces-luks.test.sh` | 26 passed, 0 failed |
| `shellcheck -S warning` | clean |
| `plan` / `validate` (apps/web-platform/infra) | pass |

PR #7269, opened 2026-08-04
docs(learnings): a redaction that does not match is indistinguishable from one that worked

deruelle:
Session-error capture from the #6808 heartbeat apply, per `wg-every-session-error-must-produce-either`. Docs-only — one new learning file, no code paths touched.

## What happened

A masked Doppler secret was echoed through a `sed` redaction whose pattern anchored on the **provider name** (`betteruptime`, which is the Terraform resource type) while the **live URL** uses a different host (`uptime.betterstack.com`). The pattern did not match, `sed` passed the line through unchanged, and a bearer token reached the transcript at full strength.

No error. No non-zero exit. `sed` exits 0 on a non-matching substitution — that is its contract, not a bug — so the guard's failure mode is silence, and silence reads exactly like success.

## Why it is worth a file

The leaked value was a Better Stack heartbeat push URL, whose only capability is *marking the probe alive*. That is precisely the signal #6808 exists to make trustworthy: it lets a holder forge liveness for the LUKS at-rest probe, making a dead probe indistinguishable from a healthy one — the defect the issue was opened to remove, reintroduced through its own verification step.

## The generalisation

Stop redacting by **subtraction** (print, then match and remove the dangerous part). A redaction regex is a guess about a value's shape, evaluated against a value you have not seen. Assert on a **derived property** that cannot carry the secret regardless of shape:

```bash
V=$(doppler secrets get NAME --plain -c cfg -p proj) && echo "PRESENT len=${#V}"
[ "$(printf %s "$A" | sha256sum)" = "$(printf %s "$B" | sha256sum)" ] && echo MATCH
```

That second form is what actually proved the rotated secret equalled the live heartbeat URL. Where a value must be displayed, redact by **construction** (`${V:0:8}…`). Construction fails closed; subtraction fails open.

## Recovery trap, recorded because it is easy to get wrong

Rotating the heartbeat alone would have left Doppler pointing at a destroyed endpoint: `doppler_secret.workspaces_luks_heartbeat_url` carries `lifecycle { ignore_changes = [value] }`, which suppresses **updates** but not **creates**. Both resources needed `-replace` for the new value to land. Confirmed dead rather than assumed dead — the old URL returns HTTP 404.

## Why a learning and not a new hard rule

`hr-never-paste-secrets-via-bang-prefix` already covers the operator-paste path, and `wg-every-session-error-must-produce-either` notes the rule corpus is under explicit budget pressure ("4.7 rules/day consumed the 100→115 raise in 2 days"). A learning file is the right weight for a variant of an already-ruled class.

Ref #6808

### DIFF
--- a/knowledge-base/project/learnings/security-issues/2026-08-04-redaction-by-subtraction-fails-open-on-a-non-matching-pattern.md
+++ b/knowledge-base/project/learnings/security-issues/2026-08-04-redaction-by-subtraction-fails-open-on-a-non-matching-pattern.md
@@ -1,0 +1,77 @@
+---
+module: Agent workflow / secret handling
+date: 2026-08-04
+problem_type: security_issue
+component: development_workflow
+symptoms:
+  - "A masked Doppler secret printed in full to the session transcript despite a sed redaction in the same pipeline"
+  - "The redaction pattern matched the provider NAME (betteruptime) while the live URL used a different host (uptime.betterstack.com)"
+  - "No error, no non-zero exit — sed passed the line through unchanged and the command reported success"
+root_cause: incorrect_assumption
+resolution_type: workflow_fix
+severity: high
+issues: ['#6808']
+---
+
+# A redaction that does not match is indistinguishable from one that worked
+
+## What happened
+
+While applying the #6808 heartbeat limb, a verification step read the freshly-created Doppler secret
+and piped it through a redaction before echoing it:
+
+```bash
+echo "$V" | sed -E 's#(https://[a-z.]*betteruptime[^/]*/api/v[0-9]+/heartbeat/).*#\1<REDACTED>#'
+```
+
+The Terraform provider is `betterstackhq/better-uptime` and the resource is `betteruptime_heartbeat`,
+so `betteruptime` looked like the right anchor. The **live URL** is
+`https://uptime.betterstack.com/api/v1/heartbeat/<token>`. The pattern did not match, `sed` emitted
+the line unchanged, and the bearer token went into the transcript at full strength.
+
+Nothing failed. `sed` exits 0 on a non-matching substitution — that is its contract, not a bug — so
+the guard's failure mode is silence, and silence reads exactly like success.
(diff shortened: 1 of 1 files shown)


deruelle:
### Heads-up: this branch will hit a modify/delete conflict on its first sync after #8384 merges

**Why you are seeing this:** #8384 (ADR-235) untracks four generated knowledge-base caches and regenerates them on demand. Your branch still tracks all four, so the first `git merge origin/main` (or rebase) after it lands will conflict — `deleted in origin/main and modified in HEAD` — on any of these your branch has touched:

```
knowledge-base/INDEX.md
knowledge-base/kb-tags.txt
knowledge-base/kb-categories.txt
knowledge-base/project/rule-metrics.json
```

This is a one-time transition, not a defect. These files are caches of the tree with no product or human consumer; nothing is lost by dropping them, and `scripts/ensure-kb-index.sh` regenerates them on demand.

**Remedy — run this on your branch, then sync:**

```bash
git rm --cached -- knowledge-base/INDEX.md knowledge-base/kb-tags.txt \
  knowledge-base/kb-categories.txt knowledge-base/project/rule-metrics.json
git commit -m "chore: untrack the generated kb caches (ADR-235 transition)"
```

If you have already started the merge and are sitting in the conflict, `git rm` the four paths to accept the deletion, then continue.

**Note the automated resolver will NOT do this for you.** `plugins/soleur/scripts/resolve-regenerable-conflicts.sh` deliberately refuses `modify/delete` conflicts: it cannot know which side of a delete-vs-modify is intended, and guessing would be the dangerous behaviour. Refusing is correct — the remedy above is the manual step it is refusing in favour of.

**Why you are being told here rather than in the docs:** the remedy is documented on the *new* `main` (`plugins/soleur/skills/merge-pr/SKILL.md`), but your branch reads its own copy of that file, so it cannot reach you until after you have already hit the conflict.

Posted to every open PR whose branch still tracks these files. No action needed until #8384 merges.

Issue #7248, opened 2026-08-03
The zot health verdict is read-path only — a registry that serves reads while every push fails grades green


deruelle:
The registry's health verdict only ever probes the **read** path. Nothing anywhere probes the **push** path — which is the path releases actually depend on, and the one that was dead for 3.5 hours on 2026-08-03 while every health surface reported fine.

Surfaced while fixing #7242.

## The gap

`zot-liveness-heartbeat` (`apps/web-platform/infra/cloud-init-registry.yml`) probes `http://${private_ip}:5000/v2/` and accepts **200 or 401**, then arms a Better Uptime heartbeat. Two independent problems:

1. **It cannot see a restart loop.** It takes a single sample per 60 s. A loop running at ~4 restarts/min (measured 2026-08-03, over four hours) has zot up for most of any given instant, so the probe samples a healthy-looking origin and the heartbeat stays armed the whole time.
2. **It cannot distinguish "restarting" from "healthy" at all.** A freshly-restarted zot answers `401` within the sample window, and `401` is an accepted verdict. So the probe's success is compatible with a zot that has restarted a thousand times.

And because `401` is the *anonymous-read* response, the probe carries no push credential and exercises no write. A registry that serves reads while refusing or dropping every push grades **green**.

That is what happened: the read path was genuinely fine the entire time, so every advisory that spoke to it said so, while the push path had been dead since 17:11 UTC.

## Why this is not a small change

Measured surface (6 files), which is why it is filed rather than folded into #7242:

| File | Why it must change |
|---|---|
| `apps/web-platform/infra/cloud-init-registry.yml` | `zot-liveness-heartbeat.sh` — the read-only probe itself |
| `apps/web-platform/infra/cloud-init-registry.yml` | a second unit + timer for a push probe |
| `apps/web-platform/infra/zot-registry.tf` | the push beat needs its **own** `betteruptime_heartbeat` + `*_heartbeat_url`; folding it into `registry_prd` re-introduces the OR-masking the file already warns about |
| `plugins/soleur/lib/heartbeat-manifest.ts` | manifest row + arming construct for the new beat |
| `plugins/soleur/test/heartbeat-reprovision-parity.test.ts` | parity suite gates that the row names a real arming construct |
| `apps/web-platform/infra/zot-liveness-heartbeat.test.sh` | the unit-existence gate |

[...]

Issue #7247, opened 2026-08-03
zot is crash-looping on the registry host (~4 restarts/min since 17:08 UTC 2026-08-03) — the active cause of the release blockage


deruelle:
zot has been in a continuous crash-restart loop on the registry host since ~17:08 UTC on 2026-08-03. This is the **active cause** of the release blockage tracked in #7242 — production is pinned three releases behind — and #7242 does **not** fix it.

This is a recurrence of the *non-OOM* class of the closed #6288.

## Measured

From Better Stack `SOLEUR_ZOT_DISK` (`doppler run -p soleur -c prd_terraform -- bash scripts/betterstack-query.sh --since 12h --grep SOLEUR_ZOT_DISK`):

| time (UTC) | `zot_restarts` | delta | `oom_kills` |
|---|---|---|---|
| 17:00 | 0 | +0 | 0 |
| 17:05 | 0 | +0 | 0 |
| **17:10** | **2** | **+2** | 0 |
| 17:15 | 22 | +20 | 0 |
| 17:20 | 45 | +23 | 0 |
| 19:05 | 477 | +19 | 0 |
| 20:22 | 765 | +16 | 0 |
| 21:05 | 950 | +20 | 0 |
| 21:25 | **1032** | +20 | 0 |

Onset is between 17:05 and 17:10, and the rate is a steady **~4 restarts/minute, unbroken for over four hours**. The first `Web Platform Release` failure is 17:11:50 — one to six minutes after onset.

`oom_kills=0` and `oom_killed=false` throughout, so this is **not** the OOM signature that #6288's memory bump addressed.

## Why it blocks releases

A `docker login` plus a three-tag `crane copy` takes tens of seconds, so at 4 restarts/min the push is near-certain to straddle a restart. When it does, the tunnel's origin dial fails mid-push and surfaces as `cloudflared: ERR failed to connect to origin error="websocket: bad handshake"` plus `docker: ... read: connection reset by peer`.

That pair reads exactly like an edge/credential refusal, which is why the standing error message blamed a stale Access token. It was wrong — see #7242.

## What is deliberately NOT asserted here

**Why zot restarts.** `oom_kills=0` rules out the OOM class. The proximate crash reason lives in the container's own exit status and logs on the host, which has not been read. No cause is claimed.

## Next probe (cheapest first, and it needs no SSH)

[...]

Issue #7265, opened 2026-08-04
infra-config-apply.test.sh: its #7220 mutation proof is pinned to origin/main, which now HAS the fix — 2 unsatisfiable assertions


deruelle:
`apps/web-platform/infra/infra-config-apply.test.sh` proves the #7220 fix by running the
PRE-fix handler and asserting it lacks the new behaviour. It obtains that "pre-fix" handler
with:

    git -C "$SCRIPT_DIR" show "origin/main:apps/web-platform/infra/infra-config-apply.sh"

`origin/main` is not a frozen baseline. #7221 (`c2de2581e`, "give the config handler an error
channel it can die into") merged that fix TO main, so `origin/main:infra-config-apply.sh` **is
now the fixed handler**. The two assertions that follow are therefore unsatisfiable:

    assert_eq "pre-fix handler carried NO fatal_line (this is #7220)" "MISSING" "$old_line"
    assert_eq "pre-fix handler reported the hardcoded files_total=0" "0"       "$old_total"

Measured: `git show origin/main:apps/web-platform/infra/infra-config-apply.sh | grep -c fatal_line`
returns **5**.

## Reproduce

    bash apps/web-platform/infra/infra-config-apply.test.sh
    # => Results: 144 passed, 2 failed
    #      FAIL: pre-fix handler carried NO fatal_line (this is #7220)
    #      FAIL: pre-fix handler reported the hardcoded files_total=0

## Why it has not reddened main yet

`infra-validation.yml` is path-filtered and has not run on main since #7221 landed — its last
three main runs are at `c2146e7a5`, `c8151fe16`, `87bc92273`, all pre-#7221. The `deploy-script-tests`
job checks out with `fetch-depth: 0`, so `origin/main` resolves and the suite does NOT take its
"SKIP (loud): origin/main unavailable" branch. The failure is latent on main and fires on the
first PR that triggers infra-validation with a post-#7221 `origin/main`.

Found that way: PR #7241 (per-config Doppler read tokens) touches neither
`infra-config-apply.sh` nor its test, synced main, and its `deploy-script-tests` job went red.

## The class

This is the documented golden-parity baseline trap, from
`knowledge-base/project/learnings/test-failures/2026-07-05-parity-baseline-must-not-be-git-show-main-of-the-replaced-file.md`:

> A golden-parity / "compare new-impl vs old-impl" `.test.sh` must pin its OLD baseline to a
[...]

PR #7274, opened 2026-08-04
fix(registry): make the zot crash cause reachable without host access

deruelle:
## Why

zot has been restarting **~4.4×/min on the registry host for ~21 hours** (#7247), blocking five releases. `zot_restarts` was 5032 and climbing at 14:25Z.

**This PR does not claim a cause.** It fixes the two reasons nobody could measure one — the deciding datum was structurally unreachable via telemetry, and the escalation path #7247 proposed (journald via the Vector allowlist) does not exist: **the registry host has no log shipper at all.** `vector.tf` is co-located with `inngest-server.service` and installed by `inngest-bootstrap.sh` — Inngest host only. The registry host's sole Better Stack egress is one `curl` POST of the `SOLEUR_ZOT_DISK` line every 5 minutes.

That is the 2026-07-28 sshd-allowlist precedent in a stronger form: an empty journald query would have read as "no crashes" when the channel was never shipped.

## Defect 1 — `zot_last_err` could not carry a crash line

It was `docker logs --tail 3 zot`. zot logs **every** request at info (`:128`), and this host's own liveness probe GETs `/v2/` every 5 minutes, so the last 3 lines are near-always routine HTTP traffic. Across the entire 21-hour loop the field carried a cause **zero times** — every sample was an HTTP 200/401 session line or a gc/dedupe info line.

The existing comment already called this *"the exact crash case this field targets."* The window was the bug.

Now scans a wide buffer and **filters** for cause-bearing lines, falling back to the original tail-3 when nothing matches — so it is never emptier than what it replaces. `--restart unless-stopped` restarts the **same** container (5032 restarts on one container id), so `docker logs` accumulates across restarts and a prior run's fatal line is still buffered.

`grep -a` is load-bearing: a truncated binary write would otherwise make GNU grep print nothing and silently re-create the empty-field bug.

## Defect 2 — `exit_code=0` was unfalsifiable

`.State.ExitCode` is `0` for **any running container**; it only carries a real value while stopped. A 5-minute probe against a ~14-second restart cycle reads `exit_code=0 state_status=running` forever — which reads as *"clean exit"* and actually means *"sampled mid-run."*

Not hypothetical: this is the misreading #7247's triage had to back out of. The `(2b)` comment warned these fields are *"often stale when a 4/min loop is sampled mid-run,"* but nothing **emitted** let a reader tell stale from real.

Adds **`zot_uptime_s`** — the age of the current run. Always-a-few-seconds is positive proof the sample is mid-loop and `exit_code` carries no information; a large value means `exit_code` is trustworthy. One field turns an ambiguous `0` into a self-describing one, with no host access.

## Placement is load-bearing

`zot_uptime_s` sits **before** `zot_last_err`. `scripts/lib/zot-telemetry-parse.sh` strips `zot_last_err=` and everything after it as an anti-spoof measure (a crafted log tail must not spoof `boot_id=`/`exit_code=137`), so a field emitted after it would be invisible to every consumer. Guarded in `registry-boot-guard.test.sh`'s field allowlist so it cannot be silently dropped later.

## Deployment note

The heartbeat script is written by cloud-init `write_files`, so this reaches the host via the existing gated `workflow_dispatch` (`apply_target=registry-host-replace`). That replace will also clear the current loop — which is why instrumenting **first** matters: the replace deploys the probe, so a recurrence self-reports instead of leaving us blind a third time. This is already a recurrence of #6288's non-OOM class.

## Verification

| Suite | Result |
|---|---|
| `registry-boot-guard.test.sh` | 80/80 |
| `zot-restart-loop-alarm.test.sh` | 54/54 |
| `zot-mirror-diagnosis.test.sh` | 54/54 |
| `reusable-release-zot-mirror-retry.test.sh` | 24/24 |

Also verified: no bare `${` introduced (Terraform templatefile safety), and the filter picks a `panic:` line out of a buffer dominated by routine HTTP traffic.

No cause is asserted anywhere in this change (ADR-166).

Ref #7247
Ref #7248

### DIFF
--- a/apps/web-platform/infra/cloud-init-registry.yml
+++ b/apps/web-platform/infra/cloud-init-registry.yml
@@ -179,11 +179,17 @@ write_files:
       # default the WHOLE capture to sentinels. Adapted from container-restart-monitor.sh:118-119
       # (adds .State.Status). Loop DETECTION keys on zot_restarts delta (reliable); the exit-reason
       # fields are for CAUSE attribution (often stale when a 4/min loop is sampled mid-run).
-      INSPECT=$(docker inspect -f '{{.Id}} {{.RestartCount}} {{.State.Status}} {{.State.OOMKilled}} {{.State.ExitCode}}' zot 2>/dev/null)
+      # StartedAt joins THIS call rather than getting its own (#7247). A second `docker inspect`
+      # would break the one-inspect invariant above in a way that defeats the field's purpose: the
+      # two calls can straddle a restart and pair an exit_code from one run with an uptime from
+      # the next, and if the first succeeds while the second misses you get
+      # `exit_code=0 state_status=running zot_uptime_s=-1` — ambiguous again, which is the exact
+      # condition zot_uptime_s exists to remove.
+      INSPECT=$(docker inspect -f '{{.Id}} {{.RestartCount}} {{.State.Status}} {{.State.OOMKilled}} {{.State.ExitCode}} {{.State.StartedAt}}' zot 2>/dev/null)
       if [ -n "$INSPECT" ]; then
-        read -r ID ZOT_RESTARTS STATE_STATUS OOM_KILLED EXIT_CODE <<<"$INSPECT"
+        read -r ID ZOT_RESTARTS STATE_STATUS OOM_KILLED EXIT_CODE STARTED_AT <<<"$INSPECT"
       else
-        ID=; ZOT_RESTARTS=-1; STATE_STATUS=unknown; OOM_KILLED=unknown; EXIT_CODE=-1
+        ID=; ZOT_RESTARTS=-1; STATE_STATUS=unknown; OOM_KILLED=unknown; EXIT_CODE=-1; STARTED_AT=
       fi
       # (2b2) #7247 — the exit_code STALENESS discriminator. `.State.ExitCode` is 0 for ANY RUNNING
       # container; it only carries a real value while the container is stopped. A 5-minute probe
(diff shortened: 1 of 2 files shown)


deruelle:
### `deploy-script-tests` failure is pre-existing and unrelated — tracked as #7265

Verified rather than assumed:

- This PR's diff is **two files**: `apps/web-platform/infra/cloud-init-registry.yml` and `apps/web-platform/infra/registry-boot-guard.test.sh`. Neither is in `infra-config-apply.test.sh`'s scope.
- Ran `infra-config-apply.test.sh` on this branch and on a worktree that does **not** contain this change: **2 `FAIL` lines in both**. Identical.
- The two failures are `infra-config-apply.test.sh:1374-1375` — `"pre-fix handler carried NO fatal_line"` and `"pre-fix handler reported the hardcoded files_total=0"`.

That is exactly **#7265**: *"infra-config-apply.test.sh: its #7220 mutation proof is pinned to `origin/main`, which now HAS the fix — 2 unsatisfiable assertions."* The mutation proof asserts the pre-fix handler misbehaves; #7221 merged the fix to `main`, so the fixture it pins to no longer reproduces the pre-fix behaviour. Self-referential, and it went red the moment its own fix landed.

Also relevant: per #6766, `deploy-script-tests` is **not a required check**.

Not fixing it here — it belongs to #7265 and folding it in would put an unrelated test repair in a registry-telemetry PR.

Tests that **are** in this PR's scope, all green:

| Suite | Result |
|---|---|
| `registry-boot-guard.test.sh` | 80/80 |
| `zot-restart-loop-alarm.test.sh` | 54/54 |
| `zot-mirror-diagnosis.test.sh` | 54/54 |
| `reusable-release-zot-mirror-retry.test.sh` | 24/24 |

Issue #7297, opened 2026-08-05
The #7220 fatal-channel follow-through PASSes while the defect is still occurring, so it auto-closed the issue it gates


deruelle:
`scripts/followthroughs/infra-config-fatal-channel-7220.sh` has two PASS arms, and one of them is
satisfied by the very failure the issue it gates is about.

```
if [[ "$n_fatal" -gt 0 ]]; then          # arm 1
  if [[ "$n_attributed" -ge 1 ]]; then
    echo "PASS: $n_fatal fatal row(s), $n_attributed carrying a line number..."
    exit 0
...
  echo "PASS: zero fatal rows AND the state frame carries the fatal_rc key..."   # arm 2
  exit 0
```

* **arm 1** — the handler IS still dying, and the death is instrumented.
* **arm 2** — the handler ran clean.

Both exit 0, and `sweep-followthroughs.sh` CLOSES the tracked issue on exit 0. So the probe cannot
distinguish "repaired" from "still broken but now loud", and it auto-closed #7220 on arm 1 at
2026-08-05T19:12:37Z while the ungranted `daemon-reload` was still aborting every apply. #7220 has
been reopened.

Arm 1 is a legitimate verdict about PR-A's DELIVERABLE (the fatal channel reports attribution — no
pre-#7220 handler can emit that marker at all). It is not a legitimate verdict about the DEFECT.
The two questions were folded into one probe, so the answer to the first is being read as an
answer to the second.

Note the issue body on #7220 predicted a mis-close from the OTHER direction — "the PR-A probe
above PASSes once the handler stops dying, which would auto-close this issue before activation was
ever graded". The prediction was right that it would mis-close and wrong about the mechanism: it
fired while the handler was still dying, not after it stopped.

## What to decide

Whether this probe should:

1. keep arm 1 but as **TRANSIENT (exit 2)** rather than PASS — "the instrument works, the defect
   persists, do not close" — with only arm 2 closing; or
2. split into two probes on two issues, one per question; or
3. keep arm 1 as PASS but gate the sweeper's auto-close on something else.

Option 1 looks right and is a ~5-line change, but it inverts a verdict PR-A deliberately shipped,
so it belongs with whoever owns that probe rather than being changed in passing.

## Why this is filed separately

Discovered while shipping PR-B for #7220. It is a defect in a DIFFERENT artifact (PR-A's probe),
and per the net-issue-flow convention a discovered defect in another subsystem stays its own issue
[...]

Issue #7296, opened 2026-08-05
follow-through: verify infra-config ACTIVATION after the #7220 daemon-reload grant


deruelle:
The #7220 repair (PR-B) grants `systemctl daemon-reload` to the deploy user. This issue carries the post-merge soak that proves **activation actually happened** — a different question from the PR-A probe on #7220, which proves the *instrument* is live.

A host can satisfy the PR-A probe and fail this one, and that combination is precisely the #7220 state: the channel described its own death perfectly and no unit was ever reconciled.

**Why this is a separate issue and not a second directive on #7220.** `scripts/sweep-followthroughs.sh` honours only the FIRST `soleur:followthrough` directive per issue body and merely logs a warning about the rest. Three independent review agents found that putting both probes on #7220 meant this one would never execute — and worse, the PR-A probe PASSes once the handler stops dying, so the sweeper would auto-close #7220 before activation was ever graded. That is the auto-close failure this probe's own header exists to prevent, one level up.

**The positive control is structurally impossible for the pre-fix handler.** `SOLEUR_INFRA_CONFIG_RESTART` is emitted by the reconciliation loop, which sits AFTER the `daemon-reload`. The pre-#7220 handler died at that reload under `set -e`, so it could never reach the loop. A signal both versions emit would prove nothing — that mistake shipped in the first version of the sibling probe, which PASSed on 40 rows while the host was dying.

Closes automatically when the probe passes.

Issue #7242, opened 2026-08-03
Web Platform Release blocked at the zot mirror since 17:11 — prod is 3 releases behind, and the error's own diagnosis is refuted in-job


deruelle:
Every `Web Platform Release` run since **17:11 UTC today** has failed at the zot-mirror bridge. Production is serving `build_sha=0ed105b4c349` (the 16:35 release) while `main` is at `e861e341da6b` — **3 releases behind**. Each blocked release stays an unpublished draft, so nothing is half-shipped, but nothing is deploying either.

## Timeline

| time (UTC) | commit | event |
|---|---|---|
| 16:35 | `0ed105b4` | Web Platform Release **SUCCESS** — last good; this is what prod still serves |
| 17:11 | `7bbd0c15` | Web Platform Release **FAILURE** |
| 17:13 | — | registry host NIC boot race → guard **rebooted** the host (#7232) |
| 17:14 | — | #7232 filed: *"self-healed … serving is FINE"* |
| 19:55 | `a6b9720c` | Web Platform Release **FAILURE** |
| 20:22 | `e861e341` | Web Platform Release **FAILURE** ([run 30849886324](https://github.com/jikig-ai/soleur/actions/runs/30849886324)) |

The first failure precedes the reboot by ~2 minutes and every run since has failed the same way, so #7232's registry host reboot is the leading hypothesis for the cause — but **#7232's "serving is fine" is only true of the read path.** The push path has been dead for 3.5 hours. A self-heal advisory that reports healthy while the mirror cannot authenticate is the gap worth closing.

## Failure

```
cloudflared: ERR failed to connect to origin error="websocket: bad handshake" originURL=https://registry.soleur.ai
docker: Error response from daemon: Get "http://127.0.0.1:5000/v2/": read tcp ...->127.0.0.1:5000: read: connection reset by peer
```

## The standing error message misdiagnoses this

The `zot mirror FAILED` error tells the operator the cause is a stale `REGISTRY_PUSH_ACCESS_TOKEN_*` and to run the drift check. **That diagnosis is refuted by a check in the same job:**

- In-job (log line ~4361): `Registry-push Access service token verified live — the zot bridge should authenticate.`
- Local `doppler run -p soleur -c prd_terraform -- bash scripts/check-cloudflare-token-drift.sh`: `live entries: 3   dead entries: 0   unverifiable: 0`

[...]

Issue #6090, opened 2026-07-06
fix(infra): web-2 fresh boot dies silently after the seed-extract (before :9000) — extend baked-DSN observability into soleur-host-bootstrap.sh + check #6023 cosign ENFORCE


deruelle:
## Problem

The **web-2 standby host has never completed a fresh boot** — after a `web-2-recreate` (`terraform apply -replace`), web-2's cloud-init dies **before `:9000` binds**, so it never becomes a working warm standby. Every deploy fan-out to web-2 reports `ok_peer_fanout_degraded`. **Prod is unaffected** (web-1 is the sole live origin at 200; web-2 is weight-0, non-serving).

## What's already fixed (this arc — all merged + live)

- **#6047** — CI GHCR digest-resolve auth (was 401 on the private package).
- **#6051** — verify retry-through-fresh-boot timing.
- **#6066** — cross-pipeline web-1-swap serialization (firehose interference).
- **#6076** — **cloud-init seed-pull GHCR `docker login` before the anonymous `docker pull` of the now-private image** (`ci-deploy.sh`/`soleur-host-bootstrap.sh` had auth; the early cloud-init seed-pull did NOT since #6005 privatized GHCR). Also baked `${sentry_dsn}` into cloud-init's `on_err` so the **seed-block** (pull/extract/verify) fatal emit fires without depending on doppler.

## Confirmed current state (definitive)

- The `web-2-recreate` **apply fully succeeds** (`-replace` + attach-proof + volume-preserved) with the fixed cloud-init.
- **No `stage=pull` fatal in Sentry** → cloud-init gets PAST the seed pull (the #6076 fix works and is confirmed by its own emit).
- **8/8 tunnel probes hit web-1** → web-2's cloudflared never comes up → cloud-init dies **before the webhook/cloudflared stage**.
- **Sentry silent** on any host-boot event → the failure is between the seed-extract and `:9000`-bind, and it is NOT observable off-host.
- A controlled `deploy web-platform v0.198.7 peers=10.0.1.10,10.0.1.11` fan-out settled `exit_code=0` but web-2 rejected the peer deploy → `ok_peer_fanout_degraded`.

[...]

PR #7298, opened 2026-08-05
fix(infra): grant the config handler's daemon-reload so delivery actually activates (#7220)

deruelle:
## Summary

The ACTIVATION repair for #7220. Since ~2026-05 the config handler's `systemctl daemon-reload` ran
ungranted, returned "Interactive authentication required", and `set -e` aborted **after** all 19
managed files were written and **before** any unit was reconciled. Delivery looked healthy;
activation never happened, and every managed unit kept running its start-time environment.

`Apply deploy-pipeline-fix` is red **by design** until this merges — turning GREEN post-merge is the
acceptance signal for the whole plan.

**Not related to the current deploy block.** app.soleur.ai/health is 200. Deploys are blocked by the
zot mirror crash-loop (#7247 / #7242), which this PR neither causes nor fixes.

Ref #7220

**Deliberately `Ref`, not `Closes`.** #7220 is enrolled in the follow-through sweeper, which only
evaluates OPEN issues — closing it on merge would silently retire the verification it exists to
carry. And the proof for this repair lands POST-merge (`Apply deploy-pipeline-fix` turning green),
so `Closes` would close it on a promise. That is precisely how #7220 got closed once already, on a
probe arm satisfied by the defect continuing; it was reopened earlier today. It will be closed by
hand once the apply is green, with the run URL as evidence.

## User-Brand Impact

- **Artifact at risk:** the production web host's config-delivery channel — the only no-SSH
  remediation path to a `cx33` host that is orderable in 0 of 6 EU datacenters.
- **Vector:** a managed unit silently running a stale environment (including a revoked credential)
  because delivery reported success while activation never occurred; and, for the alerting changes,
  a red run that notifies nobody.
- **Brand-survival threshold:** `single-user incident` — one operator hitting a wedged remediation
  channel on an unreplaceable host is a brand-survival event, independent of aggregate frequency.

### P1 — the red-gate alert notified nobody when the gate never ran

The condition was `failure() && steps.infra_config_gate.outcome == 'failure'`, which scopes the
alert to "the gate ran and reded" and excludes "the gate never ran". A step failing before it leaves
the gate `skipped`, so nothing fired: no issue, no Sentry event. This PR adds four new ways for the
apply to fail (two sudo probes, two DropInPaths assertions), so the PR whose premise is *a red gate
must notify someone* was increasing the share of red runs that notify nobody. Now
`failure() && outcome != 'success'`.

A **third `ungraded` alert mode**, rather than the reviewed-prescribed `"unreachable"` text.
Measured by deleting the new branch and driving `GATE_OUTCOME=skipped`: it falls through and emits
`REACH=unreachable`, telling a non-technical operator the delivery channel did not answer and
pointing at recovery guidance from a verify step that never executed. Both existing bodies are false
there — `reachable` asserts the files landed. `ungraded` says the one thing that is known:
activation is UNKNOWN, here is the step that failed.

### P1 — the AC6 lint certified four argv shapes sudo would DENY

`_lint_privileged_verbs` resolved each call to the bare string `systemctl <verb>` and
substring-grepped it against `Cmnd_Alias` lines, discarding the UNIT and ignoring the User_Spec.
Measured against the pre-fix lint, all of these returned `rc=0`:

| invocation | matched (wrongly) |
|---|---|
| `$SYSTEMCTL_PRIV stop webhook.service` | `INNGEST_STOP`'s `systemctl stop` |
| `$SYSTEMCTL_PRIV restart nginx.service` | `INNGEST_RESTART`'s `systemctl restart` |
| `$SYSTEMCTL_PRIV disable webhook.service` | `INNGEST_QUIESCE`'s `systemctl disable` |
| `sudo /usr/bin/systemd-run <ANY argv>` | the bare word `systemd-run` |

A `Cmnd_Alias` with no `deploy ALL=(root) NOPASSWD:` grants nothing, so the lint blessed calls that
would be denied — and a denial is the exact `set -e` abort after delivery the lint was named for.
Now resolves each invocation to its FULL argv (seam default read from the handler, not hardcoded)
and requires exact membership in a User_Spec-gated granted set. Extraction blind spots closed:
`/bin/systemctl`, variable indirection (fails closed), backslash continuation, a second
`systemd-run` line, and an unanchored comment strip that cut at the `#` of `$#`. Vacuity floor
2 → 4 (the live handler yields exactly 4).

### P1/P2 — the security framing claimed a boundary that does not exist

Verified by reading the artifacts, not accepted from review: `deploy` is **already root-equivalent**
via three independent paths — `groups: docker` (argued and accepted in the 2026-03-20 privilege
plan), 14 ungated `/usr/local/bin/*` dests one of which is root-executed, and an `INNGEST_BOOTSTRAP`
that roots a deploy-written `/tmp` path. The `*.service` digest pin is a real **integrity** control
over what systemd ADOPTS; it is not a privilege boundary. Every comment implying otherwise is
scoped, and the canonical statement lives once in `deploy-inngest-bootstrap.sudoers`.

### P2 — the activation soak PASSed on the failure it exists to detect

It counted `action=failed reason=sudo_denied` rows toward PASS, and PASS auto-closes the tracked
issue — so it would have closed the issue on the evidence of its own recurrence. Also matched the
sibling `_STDERR` marker by substring, and the unit name anywhere in the payload. Now split on the
handler's own vocabulary, with the freshness guard hoisted so a PASS is unavailable without it.

### P2 — four false claims corrected

`model.c4` (the edge did **not** report per-unit verdicts — the handler died before the loop and the
EXIT trap published hardcoded zeros); ADR-159 + `server.tf` + the bootstrap suite (`inngest-server`
**is** in the grant set, and `ci-deploy.sh`'s `case inngest)` is a second reconciliation channel);
`webhook.service` (`StartLimitBurst` is 5 in 10s — two restarts cannot blow it; the real mechanism
is a crash loop); `server.tf ~1209` (it said `deploy_pipeline_fix` does NOT `depends_on` the
bootstrap bridge while line ~1448 lists it — that comment is the load-bearing ordering argument for
the digest pin).

### P2 — operator-facing alert

The frame already carried per-file verdicts and the step already had `jq`, yet the issue told a
non-technical operator to go read a CI log. Now names the failed FILES with the host's own reason,
and renders `$SYSTEMCTL_PRIV` as the command that actually ran.

### P3 — eleven test-quality items

New `webhook.service` repo-vs-cloud-init parity check (directive-wise, because the two copies
deliberately differ in comments); pin comparison by SET not COUNT; the B6 fail-loud tail sweep;
a new floor on the bootstrap suite; `INSTALL_MIN_ASSERTIONS` hardcoded (it was env-overridable, so
`=0` disabled it); the APPLY floor reconciled with its own loud-skip arms; six invalid `hooks.json`
shapes; B7 argument-count enforcement; a de-duplicated `files[]` verdict; and the `sudo -l`
exemption scoped to sudo's own option run.

## Three pre-existing branch failures also fixed

`origin/main` was green on all three; the recorded branch state did not include them.

1. **`web-host-provisioner-parity` was 12/1, not 13/0.** The extractor classifies a segment on its
   FIRST token, and this PR's probe is `runuser -u deploy -- sudo -n /usr/bin/systemctl
   daemon-reload`; `runuser` is not in `READONLY_VERBS`, so `/usr/bin/systemctl` was credited as a
   *delivered artifact* and the guard demanded a fresh-boot writer for a base-image binary. Bisected
   to before `7256d75c9`. Fixed by resolving the effective verb through privilege wrappers.
2. **cloud-init `user_data` was 604 B over the #6090 sub-cap** at `7256d75c9` — caught only by
   `plugins/soleur/test/cloud-init-user-data-size.test.ts`, which lives outside the infra runner, so
   91/91 there and six green infra suites were all blind to it. Fixed with bytes (comment-only;
   every non-comment line byte-identical), now 68 B under.
3. **The alert step exported `SECCOMP_ALERT_*` while the helper reads `INFRA_CONFIG_ALERT_*`** — so
   every alert body ever filed omitted the Commit and CI-run lines *its own prose promised*. The
   helper's test set the correct names itself, so it passed against a mis-wired caller. Fixed and
   pinned by a lockstep that derives the variable list from the helper.

## Issue hygiene

**#7220 was auto-closed on evidence the defect was still occurring** and has been reopened. The
sweeper closed it because PR-A's fatal-channel probe exited 0 on *"the handler died and SAID
WHERE"* — arm 1 of that probe (`n_fatal > 0` and attributed) is satisfied **by the defect
continuing**. The issue body had predicted a mis-close from the opposite direction. That probe
ambiguity is filed as #7297 rather than folded in here, since it is a defect in PR-A's artifact.

Net issue flow: closing 1 (#7220), filing 1 (#7297) = **net 0**.

## Review status

The 8-agent review ran before this session; its P1/P2/P3 findings are all resolved above. The new
logic written *during* this session — privilege-wrapper verb resolution, `bash -c` unwrapping, the
`ungraded` alert mode, and three new/extended suites — carries mutation evidence but no independent
review pass.

## Test plan

- [x] `scripts/test-all.sh` — **264/264 suites, rc=0**, terminal marker present, epilogue confirms
      `apps/web-platform/infra/ IS covered above` via the nested runner
- [x] `apps/web-platform/infra/run-registered-suites.sh` — **91/91, rc=0**
- [x] `infra-config-apply` 189/0 · `infra-config-install` 58/0 · `infra-config-handler-bootstrap`
      50/50 · `infra-config-gate` 64/0 · `web-host-provisioner-parity` 14/0 · parity mutation 62/0
- [x] `infra-config-red-alert` 29/0 (up from 13) · `infra-config-activation-7220` 17/0 (new suite,
      registered in `test-all.sh`)
- [x] `plugins/soleur/test/` 2319/0 (covers the `SKILL.md` edit)
- [x] `terraform fmt` clean · `visudo` parses · cloud-init renders 8/8 · `shellcheck` clean
- [x] Every guard mutation-proved on a **sandbox copy**, with the mutation's landing asserted
      against a pristine backup before trusting any result
Post-merge acceptance (already tracked — deliberately **not** marked ⏳, so Phase 7 Step 3.5 does
not file a duplicate): `Apply deploy-pipeline-fix` turning GREEN is the acceptance signal for the
plan, and the activation soak is tracked by **#7296**, which is already enrolled in the
follow-through sweeper with an on-disk verification script
(`scripts/followthroughs/infra-config-activation-7220.sh`, hardened in this PR).

**net-issue-flow override — stated plainly.** This PR is net +1 (closes 0, files 1).

- **#7299** — `registry-userdata-budget` is RED on pristine `origin/main` (−3,636 B), introduced by
  the v2.1.20 zot pin in #7283 and never caught because `Infra Validation` has not run on `main`
  since 2026-06-11. It is a **discovered defect in a different subsystem**, which the convention
  requires stay its own issue and PR rather than be absorbed into an unrelated feature branch. It
  is not caused by this PR — verified by running the budget script on a pristine `origin/main`
  worktree.

The closes-0 is also deliberate rather than backlog neglect: #7220 is `Ref`'d, not `Closes`'d,
because its proof lands post-merge and it is sweeper-enrolled (see the note under the summary). It
will be closed by hand once `Apply deploy-pipeline-fix` is green, which returns the net to 0.


deruelle:
[deploy_pipeline_fix-drift-gate] This PR touches 4 trigger files (`deploy-inngest-bootstrap.sudoers`, `infra-config-apply.sh`, `infra-config-install.sh`, `webhook.service`), so `apply-deploy-pipeline-fix.yml` auto-applies on merge — no operator action required.

That auto-apply IS the acceptance signal for this PR: the workflow currently reds by design (the ungranted `daemon-reload`), and this change is what makes it green. Kill switch: add `[skip-deploy-fix-apply]` to a commit message.

Note the workflow itself is modified here (the alert step now fires when the gate is SKIPPED, and renders failed files + the resolved seam name), so the first post-merge run exercises the new alerting path as well as the grant.

Issue #7332, opened 2026-08-06
feat(sync): KB blueprint manifest — one source of truth for knowledge-base composition


deruelle:
## Summary

Four separate declarations of "what a Soleur knowledge base should contain" exist and none agree:

1. `/soleur:sync` Output Locations — 6 types, all under `knowledge-base/project/`
2. `apps/web-platform/lib/kb-constants.ts` `DASHBOARD_FOUNDATION_KB_PATHS` — 10 business paths
3. `apps/web-platform/app/(dashboard)/dashboard/page.tsx` `FOUNDATION_PATHS` + `OPERATIONAL_TASKS` — same 10, duplicated
4. Soleur's own `knowledge-base/` — 10 domains, ~8.4k files

(1) and (2) share zero paths. Nothing produces the `engineering/` artifacts (C4 diagrams, domain-model register) that make up the bulk of Soleur's own KB.

Measured on the first alpha tester: their synced knowledge base has exactly **23 entries, 100% under `project/`** — no C4, no domain-model register, no business domains.

Replace all four with one plugin-owned blueprint manifest that sync, the dashboard, and a KB coverage report all read.

## Artifacts

- Brainstorm: `knowledge-base/project/brainstorms/2026-08-06-kb-blueprint-manifest-brainstorm.md`
- Spec: `knowledge-base/project/specs/feat-alpha-onboarding-motion/spec.md`
- Branch: `feat-alpha-onboarding-motion`
- Draft PR: #7328

## Acceptance criteria

1. A single manifest edit adds a KB artifact type end-to-end (sync generation + dashboard card + coverage report) with no other declaration touched.
2. Re-syncing the alpha tester's repo produces a C4 model whose elements are its 7 components and whose relationships match their `## Dependencies` links.
3. The tester can open the generated diagram in the KB viewer.
4. The coverage report names absent entries without asserting the business lacks anything.
5. `/api/dashboard/foundation-status` performs no whole-tree walk (ADR-067 preserved).
6. The dashboard coverage test fails, rather than silently matching nothing, if card derivation breaks.

## Explicitly out of scope

- **ADRs mined from git history** — cut entirely, not deferred. The C4 model and domain-model register cover the context need without the fabrication risk.
- Synthesizing runbooks/checklists/policies — they stay declared, agent-owned entries.
- A `when` predicate DSL (v1 hardcodes 3-4 predicates).
- Detect-and-adopt of existing project docs.

deruelle:
[...]

PR #7336, opened 2026-08-06
feat(sync): KB producers for alpha onboarding — C4 generator, domain-model promotion, coverage summary (PR 1)

deruelle:
## What this is

PR 1 of #7332: the `/soleur:sync` producers the alpha tester actually receives. A
connected repo's knowledge base stops being `project/`-only.

Measured baseline on the dogfood repo: **23 KB entries, 100% under `project/`** —
no C4 model, no domain-model register. That is exactly sync's declared output
surface, so this is a missing-producer gap, not a de-duplication one.

PR 2 (blueprint manifest + dashboard) stays deferred and is scoped in the plan.
The split is at the CLI/hosted line, which is also the tester/no-tester line —
the hosted platform is deferred, so everything dashboard-shaped reaches nobody today.

## Changelog

- **`/soleur:sync c4`** — new area generating a LikeC4 model from component docs,
  with the full artifact set the viewer needs (`generated-components.c4`,
  `model.likec4.json`, `c4-model.md`). Non-destructive: distinct composing file,
  `GENERATED` header, refuses to overwrite anything it did not write.
- **Component dependency contract** — `dependencies:` frontmatter added to the
  component template and to sync's Project Analysis, with markdown-link fallback
  for docs written before it.
- **`domain-model-drift.sh init`** — bootstraps an absent register so the area
  works on a fresh repo.
- **`domain-model` in `all` dispatch** — new non-interactive path (init → drift →
  append) for headless runs; the standalone invocation stays terminal.
- **`/soleur:sync` coverage summary** — deterministic `kb-coverage.md` plus a
  `SOLEUR_KB_SYNC_PRODUCERS` marker on stdout and in the artifact.
- **Observability layer 7 (`cli-stdout-artifact`)** — added to
  `observability-coverage-reviewer`, for code running on a customer's self-hosted
  CLI where no Soleur-side sink exists or should.
- **ADR-171** — KB composition manifest + the observability boundary, with claim
  **3a** reconciling its consent argument against the #7342 controller/processor
  determination (see below).
- **`scripts/lint-shell-capture-exit.py`** — new repo-level gate for the
  shell-capture-exit class (`x=$(grep …)` aborting under `set -e`; `grep -c`
  printing `0` *and* exiting 1 so `|| echo 0` yields `"0\n0"`). Sibling to
  `lint-workflow-errexit-capture.py`, not a widening of it. Registered in
  `scripts/test-all.sh` as `scripts/lint-shell-capture-exit` (unit, 25 assertions)
  and `scripts/lint-shell-capture-exit-live` (gate). Ships with a 216-entry
  baseline so it blocks NEW occurrences only.

## ADR-171 reconciled against #7342 — confirmed, and amended

Phase 0.2 originally gave two independent reasons the Better Stack path was
unreachable. The review **refuted the second** (the `.claude/settings.json`
PostToolUse claim is true of the local CLI and false of this code, which is
vendored into the prod image and runs under the `Bash` marker extractor). That
left the consent argument carrying the entire layer-7 decision — asserted
freestanding, by an ADR, with no legal authority behind it. Then `e9a44b055`
(#7342) landed on `main` determining exactly this data. Reconciled, not assumed:

- **Confirmed.** The determination's §2 machine/key/purpose test puts a
  Soleur-owned sink serving Soleur's purpose in **Posture C — CONTROLLER**
  (Art. 4(7); Art. 28(10)). Its §1 limb (ii) records a live instance that is this
  case almost verbatim: reading a tester's `knowledge-base/` tree *"to measure
  knowledge-base growth as a **Jikigai** product metric."*
- **Amended — the credential limb.** The old unqualified "makes Soleur a data
  controller" missed that §2 enters **Posture B — PROCESSOR** (Art. 28(3)
  instrument required *before* processing) the moment content reaches a Jikigai
  machine, credential **or account**, and that it *"fires without anyone noticing,
  because no file moves."* A Better Stack ingest token inside `plugins/soleur/**`
  is exactly that.
- **Amended — a second independent reason.** The determination's §4 egress
  evidence rests on a claim it recorded as falsifiable in one grep over
  `plugins/soleur/`: *"no automatic or background telemetry; all egress is
  explicitly operator-invoked."* That sentence is what holds up §4(a)'s finding
  that the published position "remains TRUE within its scope." A Soleur-owned
  default sink would falsify it directly. **Layer 7 is no longer carried by the
  consent argument alone.**
- **Amended — status.** The determination is `draft-requires-counsel-review`,
  disposition **BLOCKED** pending its §11, and its **C9** requires a re-run before
  tester #2 — cited as the governing internal determination, not as settled law.
- **Also named:** this ADR's own baseline measurement of `2my8r9ry2t-wq/Skouer`
  ("23 entries, 100% under `project/`") is itself a Posture C act — the processing
  recorded as **PA-35**, *"Jikigai's collaborator access to an alpha tester's
  private repository, read for Jikigai's own metrics."*

## The gate this exists for

The dependency convention was **emergent, not specified**. Until this PR the only
place it was written down was a prose placeholder in the component template;
nothing ever told the producing side to emit a parseable form. Skouer's docs happen
to use markdown links — Soleur's own use prose, 4 times out of 4.

So the failure mode is not a crash. A link-free corpus renders elements, exits 0,
validates clean, and produces **a diagram of disconnected boxes reported as
success**. The relationship-count gate closes that, reporting `degraded` rather
than `failed` — the docs are the defect, not the run.

AC2 required proving the gate goes red. The same producer, same run:

```
frontmatter corpus  → status=ok       elements=3 relationships=3
prose corpus        → status=degraded elements=3 relationships=0
```

## Phase 0.2 — the plan's observability design was refuted

The plan specified `alert_target: Better Stack` for the liveness marker and flagged
it unverified. Measured: the ingest path exists and is citable (in-sandbox stdout →
`git-lock-marker-telemetry.ts` `MARKER_RE` → pino WARN+ → vector Source 3 →
`[sinks.betterstack]`) but is **unreachable from PR 1** for two independent reasons.

1. `MARKER_RE` is an exact-sentinel allowlist, not a prefix match — a new sentinel
   is dropped at the extractor.
2. The hook is registered only on the hosted path
   (`agent-runner-query-options.ts:38`); `.claude/settings.json` registers no
   PostToolUse `Bash` matcher. The self-hosted CLI — PR 1's only surface — is
   unmirrored by construction, and a customer machine has no route to Soleur's
   Better Stack regardless.

Routed to the `cto` agent, which rejected widening `MARKER_RE`: it would add a
permanently-dead allowlist entry on a surface no tester is on (the
`inngest-boot-phone-home` anti-pattern `vector.toml` already names), and the CLI
half would be a consent violation — shipping repository-derived metadata from a
customer machine to a Soleur vendor with no consent mandate in this PR.

Ruling: emit the marker **twice** — stdout and the committed artifact — so the
`discoverability_test` becomes a real grep with no network and no credentials:

```
grep -n 'SOLEUR_KB_SYNC_PRODUCERS' knowledge-base/project/kb-coverage.md
```

A drift guard asserts both surfaces carry a byte-identical field set, and a
negative AC asserts the marker carries **counts only** — keeping ADR-171's consent
claim true in code, not just prose.

## Two bugs the unit tests structurally could not see

- **likec4 exports `relations`, not `relationships`.** The first `countModel` read
  `undefined` → 0, so *every* corpus would have reported degraded. A gate that
  always fires is exactly as useless as one that never does. `assessRender` takes
  plain numbers, so the key mapping was invisible to it; `countModelJson` now lives
  in the lib and is pinned against the repo's own committed `model.likec4.json`.
- **The E2E suite died silently mid-file** on the `set -e` command-substitution
  trap — the producer exits 1 on a source fault, which aborted the assignment
  before any assertion printed.

## Two claims refuted rather than shipped

- **AC13's `npx tsc` cannot typecheck this diff.** 0 files under
  `apps/web-platform/`, and no tsconfig covers `plugins/`. Separately, bare
  `npx tsc` in a worktree with no `node_modules` installs an unrelated registry
  package `tsc@2.0.4`. Ran the pinned typescript@5.7.2 under `--strict` against the
  three new modules instead — clean. Recorded in `tasks.md` rather than checked off
  as written.
- **A hypothesized `pipefail`/SIGPIPE fail-open in `domain-model-drift.sh`.** With a
  working positive control (`yes | grep -q y` → 141) the existing form still
  returned 0: `printf` is a bash *builtin*, and bash does not propagate SIGPIPE as
  141 for builtins. Reverted — no churn on an unproven premise.

## Verification

- Mutation-proved: **13 mutants across three batteries, all killed** (relationship
  gate, overwrite refusal, deprecated filter, dangling-edge drop, version pin,
  DIAG_RE anchor, register confinement, idempotence, table separator, deficiency
  wording, embedded timestamp, artifact field drop, repo-data leak).
- One mutation exposed a real fixture gap: the byte-identical idempotence check
  could not distinguish "init no-ops" from "init rewrites identical bytes". Added
  the case that matters — `all` dispatch calls `init` every run, so a
  non-idempotent init would clobber appended rows.
- The full-suite run caught `tests/commands/test-sync-domain-model.sh` asserting
  the exact behavior AC6 required inverting. Not in the diff, so no touched-file
  loop reaches it; re-pinned to assert both halves of the new two-contract shape.

### Rebase + first full local suite (this PR was a draft, so CI had never run)

The branch was cut on top of `feat-alpha-onboarding-motion`, whose 8 commits had
since merged as the *squashed* #7328. A plain `git rebase origin/main` replayed
them and conflicted repeatedly; the correct form was
`git rebase --onto origin/main <prefix-tip>`. Dropping the already-merged prefix
surfaced two defects that had been cancelling out:

- The branch **net-deleted** `knowledge-base/project/specs/feat-alpha-onboarding-motion/spec.md`
  (a worktree-reuse artifact), invisible while the commits that created it were
  still in the branch's own history. Restored from `origin/main`.
- `model.likec4.json` conflicted with **1149** `"id"` occurrences on main vs **1163**
  on the branch and **neither a superset** — both sides had legitimately regenerated
  it. Picking a side yields an artifact matching neither source tree and fails
  `c4-model-freshness.test.sh`. Re-derived via `scripts/regenerate-c4-model.sh`.

Because the PR had been a draft throughout, no `pull_request` workflow had ever run
on any of its commits — the first full local `scripts/test-all.sh` was the first
real signal. It reported **269/271 suites**:

- **1 real failure**, now fixed: `grok-inspect-contract.test.ts` — the layer-7 edit
  to `observability-coverage-reviewer.md` had left `.grok/agents/*.md` and
  `agents.manifest.json` stale. Regenerated; `--check` exits 0.
- **1 environmental**: `apps/web-platform` → `vitest: not found`. This worktree has
  no `node_modules`, and the PR touches **zero** files under `apps/web-platform`.

Post-rebase re-verification: `lint-shell-capture-exit` unit **25/25**, live gate
**0 new findings / 216 baselined**, `c4-model-freshness` ALL TESTS PASSED.

Preflight: PASS (security headers, env isolation `dev≠prd`, node-only-encoding scan;
remaining checks fast-path SKIP). ADR-ordinal gate clean (171 is next after main's 170).
Net-issue-flow **+0**.

## Issue linkage — deliberately `Ref`, not `Closes`

Issue #7332's subject is *"KB blueprint manifest — one source of truth for
knowledge-base composition"*, and **this PR does not deliver the manifest** — PR 2
does. Auto-closing on merge would mark the issue done while its stated deliverable
is unbuilt, so the reference is intentional and #7332 stays open for PR 2.

Ref #7332


deruelle:
### `secret-scan` rename-guard: evidence for `secret-scan-allow-rename`

The `rename-guard (allowlist destinations)` job fails on a rename introduced by an
earlier planning-phase commit on this branch:

```
knowledge-base/project/specs/feat-alpha-onboarding-motion/spec.md
  -> knowledge-base/project/specs/feat-kb-blueprint-manifest/spec.md
```

The guard exists to catch a file moving **from a scanned path into a
gitleaks-allowlisted one**, where content that was previously scanned would stop
being scanned. That is not what happened here — verified rather than asserted:

| Check | Result |
|---|---|
| Source matches the allowlist regex `knowledge-base/(plans\|project/(plans\|specs))/.*\.md$` | **yes** |
| Destination matches the same regex | **yes** |
| Net change in scanning posture | **none** — allowlisted → allowlisted |
| Canonical secret-shape scan of the renamed file (PEM headers, `ghp_`/`ghs_`/`github_pat_`, `sk-ant-api03-`, `sk_live_`/`sk_test_`, `sbp_`, `xoxb-`, `AKIA`, Doppler `dp.*`) | **0 matches** |

Both paths sit in the same allowlist class, so no content transitioned out of
scanning coverage. The rename is a spec directory being renamed to match the
branch it belongs to.

The guard offers two escapes: this label, or a `Rename-Allowed-By:` commit
trailer. The trailer route would require rewriting an already-pushed commit, so
the label is the appropriate mechanism.

Issue #7340, opened 2026-08-06
follow-through (Dependency): re-evaluate the BLOCKED registry-host actions when #6929 lands


deruelle:
## Close criteria

**#6929 reaches state `CLOSED`.** Verified by
`scripts/followthroughs/registry-luks-blocker-6929.sh` — a `Dependency`-shape probe
(`gh issue view 6929 --json state --jq .state`) run by the follow-through sweeper.

## Why this exists

#7278 / ADR-171 record three registry-host actions as **BLOCKED ON A PROVISIONING EVENT**:

| Action | Why it is blocked |
|---|---|
| `restart` | Needs host execution; there is no in-place path (the host is cloud-init-only, ADR-096). *Independently refuted anyway* — `--restart unless-stopped` has already restarted zot **15,640** times into the same 100 %-full volume. |
| `push-config` (tighten the keep-set, grant `delete`, add per-path disk telemetry) | `/etc/zot/config.json` and the monitor scripts are cloud-init-written. This is the action that would unlock the other two. |
| `reclaim` | **No zot user holds `delete`** — measured: the deny-by-default `accessControl` grants the pull user `["read"]` and the push user `["read","create","update"]`. Granting `delete` *is* a `push-config` change. And no on-demand GC endpoint exists (measured: `/v2/_zot/gc`, `/v2/_catalog/gc`, `/_zot/gc`, `/v2/_zot/ext/gc` all 404, and the pinned build's `BinaryType` excludes mgmt/scrub/search). |

All three converge on the same chain: a config change → cloud-init must re-run → the host must be
**replaced** → and a replace while #6929 is open opens `/dev/mapper/registry` against a
still-plaintext ext4 volume, takes the `refusing-non-luks-device` arm, and darks the sole pull path
**permanently**.

**The problem this issue solves is not the blocker — it is the prose.** That blocked-actions list
lives in a plan, an ADR and a runbook. Nothing anywhere notices when its premise stops holding. The
day #6929 closes, three documents silently become wrong and only a human re-reading all three would
find out. This tracker converts "remember to re-check" into an automatic sweep.

## What to do when this closes

Re-evaluate the three actions above, and correct the documents that assert the blocker as standing:

- `knowledge-base/engineering/architecture/decisions/ADR-171-ci-side-observability-emission-and-read-only-registry-inventory.md` (§8 and the blocked-set framing)
[...]

Issue #7339, opened 2026-08-06
follow-through: observe a real SOLEUR_ZOT_INVENTORY marker with enumeration_complete=true (flips ADR-171 adopting -> accepted)


deruelle:
## Close criteria

A **real dispatch** of `registry-zot-inventory.yml` produces a `SOLEUR_ZOT_INVENTORY` line that is
**observed in Better Stack Logs** with `enumeration_complete=true`.

Verified by `scripts/followthroughs/zot-inventory-marker-7278.sh`, which the follow-through sweeper
runs on its own cadence. Nobody needs to come back to this issue.

## Why this is a dedicated tracker and not a comment on an existing issue

Two obvious hosts were considered and both are wrong:

- **Not #7278.** That issue is closed by the PR that ships this lever. The sweeper lists
  `--state open`, and its reopen path fires only on exit **1** — so a correctly-written exit-2
  probe hosted there would be a **permanent silent no-op**, never running and never reporting.
- **Not #7247.** That is a live P1 incident. It will close when the incident does, and it would
  take this tracker with it while the ADR flip condition was still unmet.

## What is actually pending

`ADR-171` ships at status **`adopting`**. Its flip to `accepted` is gated on this measurement, and
the measurement cannot be taken before merge — the workflow does not exist on a runner until it
lands.

The open sub-question is **H6: does a GitHub runner's egress reach the Better Stack ingest pin?**
What was measured pre-merge, and its bound:

| Measured | Value | Bound |
|---|---|---|
| Ingest POST | HTTP **202** | from a **workstation**, not a runner |
| POST → queryable | **17 s** | same bound; this is what sizes the workflow's poll budget |
| Prior GitHub Actions POSTs to Better Stack | **none, ever** | this PR is the first, which is why an ADR was needed |

A workstation 202 is not a runner 202. If the region pin refuses runner egress, the dispatch emits
`reason=ingest_rejected_http_<code>` from the POST's own status code, and *that* is the evidence
on which the Sentry-mirror fallback would be chosen — not a guess.

## Reading the result when it lands

- **Require `enumeration_complete=true`.** An incomplete sweep produces a large `delta_gb` that is
  indistinguishable from the finding, so it licenses **no** conclusion. A prototype sweep run from
  a workstation on 2026-08-06 returned 14.78 GB manifest-referenced against ~56 GB used — with
  **12 manifest errors**, i.e. `enumeration_complete=false`. That number must not be quoted as a
  result.
[...]

Issue #7287, opened 2026-08-05
Apply the staged zot v2.1.20 pin — ordered path, recut BEFORE host-replace


deruelle:
The zot pin was moved to **v2.1.20** in #7283 (closes #7282). The change is **staged, not shipped**: `zot-registry.tf` resources are `OPERATOR_APPLIED_EXCLUSIONS`, so merging is inert in production by construction. This issue tracks getting it applied.

## Do NOT start with `registry-host-replace`

**Firing `registry-host-replace` in today's state darks the registry permanently.** The new host's boot guard requires `/dev/mapper/registry` while the live volume is still plaintext ext4, so cloud-init FATALs and the sole pull path is gone. Both the recut runbook (§*"Do NOT use `registry-host-replace` for this"*) and #6929 say so.

This issue therefore does **not** carry that invocation. The order below is the whole point.

## Ordered path

1. **Fix #7277** — the `registry-luks-recut` D10 gate has no valid PASS condition, so the recut cannot run at all.
2. **Run the runbook's five REQUIRED pre-first-fire cold-vehicle re-verifications** (`knowledge-base/engineering/operations/runbooks/registry-luks-recut-6929.md` §"Before the FIRST-EVER fire"). These are declared REQUIRED, not advisory.
3. **Fire `registry-luks-recut`** (menu-ack, destructive — empties the store and opens a Sentry-alarmed window).
4. **Fire a release** to close the empty-store window.
5. **Record the new volume id.**
6. **Only now** is `registry-host-replace` the correct tool.

## Live precondition probe — run this, do not fire blind

```bash
# (a) is the server type orderable RIGHT NOW? (.available, never .supported)
#     A revert also needs a second successful create, so re-probe immediately before firing.
gh api /v1/datacenters --jq '.datacenters[] | {name, available: .server_types.available}' 2>/dev/null || \
  echo "use the hcloud probe from PR #7280"

# (b) has #7280 MERGED? (state != OPEN is not sufficient — a close-without-merge is not "cleared")
gh pr view 7280 --json state,mergedAt

# (c) HARD CAP CHECK — rendered user_data must be < 32768 B
bash apps/web-platform/infra/registry-userdata-budget.sh --json
```

**(c) is NOT breached. It never was, after #7280 — the gate was reading its own render wrong.** (corrected 2026-08-06 from PR #7303; the gate fix is PR #7300)

[...]

Issue #6929, opened 2026-07-24
encryption-posture: add guarded registry-luks-recut workflow_dispatch (removes registry-host-replace FATAL footgun) — follow-up to #6895


deruelle:
Follow-up to #6895 (registry guest-side LUKS at-rest). The landing PR shipped the cloud-init + Terraform + ledger-flip apparatus but **deliberately did NOT** add the guarded recut dispatch (operator scoped that PR to cloud-init + Terraform + ledger). This issue tracks adding it.

## What to build
A lean, destroy-guarded, typed-confirm `registry-luks-recut` `workflow_dispatch` target in `apply-web-platform-infra.yml`, mirroring the existing `registry-region-migrate` / `workspaces-luks-recut` guards. It performs the sanctioned recut as ONE atomic apply:

```
terraform apply \
  -replace='hcloud_volume.registry' \
  -replace='hcloud_volume_attachment.registry' \
  -replace='hcloud_server.registry'
```

A fresh raw volume ⇒ cloud-init luksFormats it (guest-side LUKS) ⇒ zot re-fills from GHCR. All three resources `-replace`d **together**. Adding the job is code-only (unfired ⇒ no live mutation); firing it is the gated operator step.

## Why (the footgun this removes)
The operator must **NOT** use the existing `registry-host-replace` dispatch for the recut — it **preserves** the (still-plaintext) volume, so cloud-init boots it into the D1/Option B `blkid TYPE` else→FATAL refuse arm and **darks the registry**. Forgetting the `-replace` on the volume has the same effect. Leaving the operator only a bare `-replace` recipe is exactly the manual, error-prone step Soleur's non-technical-operator principle (`feedback_never_defer_operator_actions`) says to encode, not defer. The guarded dispatch removes that footgun.

## Re-evaluation criteria
- Add it when the first live recut of `hcloud_volume.registry` is scheduled (it becomes the sanctioned vehicle), OR sooner if any other volume-preserving dispatch increases the footgun surface.
- Mirror the typed-confirm + destroy-guard + id-pin patterns already used by `workspaces-luks-recut`.
- Register any new guard test in `infra-validation.yml`.

Interim vehicle until this ships: the sanctioned `OPERATOR_APPLIED_EXCLUSION` three-`-replace` apply (all three resources targeted together).

Ref #6895. See the ADR-096 amendment (2026-07-24) for the recut vehicle + footgun record.

---
[...]

PR #7343, opened 2026-08-06
feat(registry): a read-only zot disk-inventory lever — measure what is actually consuming the 59 GB

### LINKED ISSUE #7278: The registry host has no in-place restart lever — the Inngest host has one, this one has nothing
## Summary

There is **no way to restart zot on the registry host** without destroying or replacing something. The Inngest host has `restart-inngest-server.yml`; the registry host has no equivalent, and no shell.

On 2026-08-04 that gap escalated a container crash-loop into a 22-hour, 6-release outage in which the only proposed remedies were *destroy the host* or *destroy the store* — both of which turned out to be blocked anyway (#7277).

## The current operational surface for a crash-looping zot

| Option | Status |
|---|---|
| Restart the container | **Does not exist** |
| Shell in | **Does not exist** — the CF tunnel's `ssh.` ingress reaches web-1, not the registry host (deny-all-public, by design) |
| `registry-host-replace` | Blocked while the store volume is unencrypted (#6929): `out_of_scope=2`, and forcing it leaves the registry permanently dark |
| `registry-luks-recut` | Blocked — gate has no valid PASS condition (#7277) |

That is an empty set. The only remaining strategy is to wait and hope it self-heals.

## Why the escalation is disproportionate

A container restart is a **seconds-long, reversible, data-preserving** operation. A host replace is minutes-long and reprovisions everything. A store recut is irreversible and takes the only pull path offline. Having only the latter two available means every zot problem — however small — gets triaged as a candidate for destroying production infrastructure.

Note the `--restart unless-stopped` policy already restarted the container 5000+ times without recovering, so "Docker will handle it" is not the answer; what is missing is an *operator-reachable* restart with the ability to change something (pull a fresh image, clear a cache, restart with different flags).

## Sketch

Mirror `restart-inngest-server.yml`. The registry host has no SSH ingress, so the transport is the open question, not the action. Options:

- add an `ssh.`-style tunnel ingress for the registry host (widens the attack surface — needs its own review);
- a small authenticated control endpoint on the host, reachable over the existing tunnel, allow-listing a fixed set of actions (`restart-zot`, nothing else);
- reuse the deploy-webhook pattern already used for `deploy.soleur.ai` (HMAC + CF Access).

The third looks closest to existing precedent and adds no new inbound.

## Acceptance

- [ ] An operator (or agent) can restart zot without a host replace, a store recut, or a shell.
- [ ] The lever is fail-loud and its outcome is observable from telemetry, not just an exit code.
- [ ] It cannot be used to reach anything other than the zot container.
- [ ] `registry-luks-recut-6929.md` and #7247's triage notes point at it as the first thing to try.

## Related

- #7247 — the 22h crash-loop where this gap was the binding constraint
- #7277 — the recut gate that is also blocked
- #6929 — the unencrypted volume blocking the replace path
- #6126 — zot HA + read-replicas (the structural fix; this is the cheap tactical one)

deruelle:
## What this is

A **read-only disk-inventory lever** for the zot registry host, and the measurement it exists to produce.

Closes #7278.

## Why the scope changed

The 2026-08-04 plan scoped a **restart** lever. That cannot fix the live incident: `SOLEUR_ZOT_DISK` shows `pcent=100` continuously and **15,640 automatic restarts** have not recovered it, because restarting into a full volume hits the same wall. That is exactly the STOP condition the old plan's own task 0.5 defined, and it fired.

Re-scoped to the one action with a **measured capability** behind it. zot's deny-by-default `accessControl` grants pull `["read"]` and push `["read","create","update"]` — **no user holds `delete`** — so reclaim over the existing ingress is impossible without a host config change. The pull user's `read` is sufficient to enumerate the store.

`restart` / `push-config` / `reclaim` are documented as **BLOCKED ON A PROVISIONING EVENT**, not dropped. They survive this PR and are tracked in **#7377**, because closing #7278 on its original title ("the registry host has no in-place restart lever") would otherwise read as closing the gap it names.

## Review outcome — the lever had never worked

A 9-agent review found five merge-blockers. The most important: **this lever exited 2 on 100% of production dispatches and had never measured anything.**

`doppler run --` injects the WHOLE config, and `soleur/prd` carries `ZOT_PUSH_USER`/`ZOT_PUSH_TOKEN` alongside the three secrets the enumerator needs — so the script's own D7 refusal fired at entry, every time. Its two requirements (pull+logs present, push absent) are unsatisfiable under whole-config injection. Verified against live Doppler: bare injects five names, `--only-secrets` injects exactly three and every precondition passes. This also makes the ADR's NON-MATERIALIZATION claim structural rather than asserted — the push credential is never placed in the environment at all.

The numbers below therefore come from a **hand-run prototype** over the same ingress, which is why nothing caught this until review.

| Blocker | Effect |
|---|---|
| `doppler` whole-config injection | Lever exited 2 on every dispatch |
| `LINK_UNFOLLOWED` incremented inside a subshell | A refused `Link` truncated the sweep while reporting `enumeration_complete=true`. Fires on a **spec-legal absolute URL** — no attacker needed |
| `jq` parse failure swallowed by `2>/dev/null` | A truncated-but-200 manifest contributed 0 bytes with no counter — the *expected* failure at 4.8 restarts/min |
| `outcome=ok` reachable with `enumeration_complete=false` | Contradictory durable row; now impossible by invariant in `emit_and_exit` |
| No wall-clock deadline | A timed-out run produced no marker at all — byte-identical to never being dispatched |

**Every one of these biased `delta_gb` upward** — toward "there is more garbage" — which is the direction that makes destroying the store look justified. A search for a downward-biasing path found none.

Also fixed: a follow-through probe that would have announced "the plaintext-volume blocker is gone" on its first sweep (re-anchored on the unfired recut, #7287); an ADR-171 ordinal collision with open PR #7336 (renumbered to ADR-172); eight stale "#6929 is OPEN" citations, one of which regenerated into a live GitHub issue on every crash-loop alarm; six artifacts promising the marker lands on #7247 when the workflow posts to #7339; two guard suites that exited 0 with every assertion removed; and four guard probes satisfied by the comments describing them.

## The measurement

Prototyped against the live registry over the existing CF-Access ingress:

| Metric | Value |
|---|---|
| `unique_blobs` | 266 |
| `manifest_referenced_bytes` | 15,867,729,779 (**14.78 GB**) |
| `manifests_fetched` / `manifest_errors` | 45 / **12** |

**This licenses no conclusion.** 12 manifest errors ⇒ `enumeration_complete=false`, so it is a **lower bound** from an incomplete sweep. `delta_gb` is an **upper bound on unreferenced bytes**, never a cause; the two candidates (zot's dedupe cache DB, orphaned `.uploads/` staging) have different remedies.

Note the marker emits `delta_gb` **without** the ~2.95 GB ext4-reserve correction, deliberately: `df --output=used` is not in `SOLEUR_ZOT_DISK` and adding it is a cloud-init change, i.e. a host replace. So **a delta under ~3 GB is not distinguishable from zero**, and the marker's figure runs ~3 GB above a reserve-corrected one.

## Design

- **Read-only, proven at the wire.** GET only, verified by a recording origin that 405s anything else — not by a source grep, which passes on an empty file. No `-L`, so no redirect can convert a method.
- **The pass condition lives in a script**, not in workflow YAML where no test can reach it. The run is green only after the marker for **this** `run_id` is read back out of Better Stack carrying `enumeration_complete=true`.
- The same argument was applied to the disk sampler during review: ~110 lines of query/anchor/decode logic lived twice in workflow YAML, with a weaker decoder than its two sibling scripts. Extracted to `scripts/zot-disk-sample.sh` with 15 assertions and a four-way exit taxonomy so callers can tell "could not measure" from "measured absent".
- **Job-level guard** (`if: github.event_name == 'workflow_dispatch'`) so the registration `push` trigger can never fire the op in production.

## Verification

| Gate | Result |
|---|---|
| `test-zot-inventory` | **146 / 0** (was 125) |
| `test-zot-inventory-assert-marker` | **43 / 0** |
| `test-zot-disk-sample` | **15 / 0** (new) |
| `registry-zot-inventory-workflow-guard` | **52 / 52** |
| `restart-inngest-workflow-guard` | **6 / 6** |
| CI on this head | **71 pass / 0 fail** |

Every fix is mutation-verified on a sandbox copy with a GREEN unmutated control first, each mutation diffed against a pristine copy to confirm it landed. Both workflow-guard suites gained anti-vacuity floors after a measured demonstration that stripping their assertions left them exiting 0.

## Constraints held

- **Zero Terraform.** No `.tf`, no lockfiles, no registry recreate / host-replace / luks-recut fired or staged.
- **All telemetry self-pulled.** No operator was asked to fetch anything.
- **Recurring cost delta: EUR 0.00/mo.** No new vendor, no new resource, no plan change.

## Plan

`knowledge-base/project/plans/2026-08-06-feat-registry-zot-inventory-lever-plan.md` — carries the
`## User-Brand Impact` section (threshold: `aggregate pattern`) and the `## Observability` block with
the `discoverability_test.command`. The link is load-bearing: preflight Checks 6 and 10 resolve both
sections through it, and an earlier rewrite of this body dropped it.

## net-issue-flow override

Gate reports Net +2 (Closing 1: #7278 — Filing 3: #7352, #7377, #7393). Per-issue:

- **#7352 — not filed by this PR.** Created 2026-08-06T20:20:09Z by the operator, three days
  before this work began, and merely *cites* #7343 in its body. This is the FILED query's
  documented over-match ("a PR that merely CITES #N surfaces too"), not a filing. Counting it
  is a false positive; no action is appropriate.
- **#7377 — a discovered capability gap that must stay its own issue.** #7278's title is "the
  registry host has no in-place restart lever", and this PR delivers a read-only inventory lever
  instead. `restart` / `push-config` / `reclaim` remain unbuilt and are blocked on a provisioning
  event, so closing #7278 without #7377 would make a still-open gap read as closed. Folding it
  inline is impossible — the blocker is a host-config change requiring a host replace, not size.
- **#7393 — forced by a SKILL.md phase mandate carrying no rule id.** `/soleur:preflight` Check 10
  refuses to execute this plan's credentialed discoverability probe (`doppler` is on its
  denylist) and offers exactly three dispositions; option 2 is "defer with a tracker issue". The
  probe cannot be rewritten to an unauthenticated form because Better Stack has no unauthenticated
  query API, and the durable fix (probe-verb allowlist) is a change to the preflight skill, not to
  this PR.

Neither filing is review-finding residue: every one of the ~20 review findings on this PR was
fixed inline across 11 commits, and zero `deferred-scope-out` issues were opened.

## soak-followthrough-enrollment override

The gate flagged **#7278** — the issue this PR **closes**, i.e. the work item, not a soak
tracker. Enrolling it is unsatisfiable by construction: it stops existing at merge, and the
sweeper skips non-OPEN issues.

The soak-gated trackers this PR actually declares are enrolled, verified at ship time:

| Tracker | Label | Directive | Probe on disk |
|---|---|---|---|
| #7339 (observe a real SOLEUR_ZOT_INVENTORY marker) | `follow-through` | present | `scripts/followthroughs/zot-inventory-marker-7278.sh` |
| #7340 (re-evaluate the BLOCKED registry actions) | `follow-through` | present | `scripts/followthroughs/registry-luks-blocker-6929.sh` |

Both OPEN, both carrying `<!-- soleur:followthrough script=… earliest=… secrets=… -->`, both
probes committed here and executable. #7340's probe was re-anchored during review from the recut
*vehicle* (#6929, closed) to the recut *execution* (#7287, unfired) — without that it would have
posted a false all-clear on its first sweep.

**This override is a stopgap, not the fix.** Commit `cf5f41f83` in this PR corrects the gate
itself: it collected trackers via `(Ref|Tracks|Closes|Fixes)`, so every PR that closes an issue
while declaring a soak was unshippable. Now `(Ref|Tracks)` only, fixed in both the hook and
`ship/SKILL.md` (which a test pins byte-identity against), with a mutation-verified regression
pin. The override is needed here only because the running session loads the pre-fix hook.


deruelle:
## Rebased onto main, and two of this PR's blocker claims re-measured

### Merge

Merged `origin/main` (was `CONFLICTING`, now `MERGEABLE`). The only conflict was `model.likec4.json`, a **generated** artifact; `model.c4` — its source of truth — auto-merged cleanly, so the JSON was regenerated via `scripts/regenerate-c4-model.sh` rather than hand-merged. `c4-model-freshness` (3/0) passing is the proof that resolution was correct, not merely plausible.

Suites re-run post-merge, all green: `test-zot-inventory` **125/0**, `test-zot-inventory-assert-marker` **43/0**, `registry-zot-inventory-workflow-guard` **52/52**, `restart-inngest-workflow-guard` **6/6**, `c4-model-freshness` **3/0**, `c4-count-parity` **10/0**.

### Blocker claim 1 — REFUTED

This PR states that `restart` / `push-config` / `reclaim` are blocked because a host replace "opens `/dev/mapper/registry` against a still-plaintext ext4 volume (#6929, **OPEN**)", and the triage table it inherits attributes that to the LUKS resources being *declared but absent from Terraform state* (`out_of_scope=2`).

Measured against live prod state (`terraform state list`, read-only, `-lock` untouched, 217 resources):

```
random_password.registry_luks        IN STATE
doppler_secret.registry_luks_key     IN STATE
hcloud_server.registry               IN STATE
hcloud_volume.registry               IN STATE
hcloud_volume_attachment.registry    IN STATE
```

Both LUKS resources — exactly the two that would produce `out_of_scope=2` — **are** in state. The stated mechanism does not reproduce. Also, **#6929 is now CLOSED**: its deliverable shipped in PR #6937 (`dcae7bf1`) under a prose `Ref`, which is why it read as open for two weeks.

### Blocker claim 2 — NOT refuted, and not assertable in either direction

Whether the live volume is actually LUKS-formatted or still plaintext is a **different** question from state membership, and I am explicitly not claiming it is resolved. `scripts/encryption-posture-ledger.json` records the registry store as `guest-luks-volume` but with:

```
"live_verification": "unavailable:no zot-host at-rest posture probe yet; tracked #6895"
```

An unverified declaration is not evidence. Since cloud-init only `luksFormat`s a **fresh** volume and no recut has been fired, the existing volume may well still be plaintext — so the provisioning-event blocker on `restart`/`push-config`/`reclaim` **stands**, on that ground rather than the state one.

### Net effect on this PR

None on scope. The re-scope from a restart lever to a read-only inventory lever rests on its **primary** measured grounds — 15,640 restarts into a 100%-full volume — which are untouched by the above. What changes is the *citation*: the secondary blocker should read "the store volume's at-rest posture is unverified (#6895)", not "#6929 is open / the resources are absent from state."

PR #7396, opened 2026-08-10
fix(kb): archive #7373's artifacts and correct three stale cleanup-merged claims

deruelle:
## Summary

PR #7373 merged with its spec dir and plan still at their live paths. This archives them and corrects the claims that made the omission easy to make.

`compound`'s automatic consolidation is what durably archives a feature's artifacts — it invokes `archive-kb.sh` (`git mv` + commit) on the branch, so the move rides the PR into `main`. `cleanup-merged` does not, and ship's Phase 7 Step 4 already says so. Three other places still said the opposite.

## What changed

**Archival.** #7373's spec dir and plan moved to `archive/` with a timestamp prefix. Verified complete: that merge added four KB files, and the three archivable ones are all covered (the learning correctly stays live).

**Three stale twins corrected.** A prior PR fixed ship's copy of the "cleanup-merged archives specs" claim and left the rest. `plugins/soleur/skills/merge-pr/SKILL.md` — a first-class merge path that runs the same command — still asserted it, as did `worktree-manager.sh`'s own `--help` output. Marking one site and leaving its twin is worse than marking neither: unmarked reads as still-live.

**compound's note corrected.** It claimed consolidation was "the ONLY durable archival mechanism". False — `archive-kb.sh` is the mechanism; consolidation is its *automatic, unprompted* invoker and `soleur:archive-kb` is a first-class manual one. The step is Auto-Consolidation **Step E**, not "step 4", and the "no conversational output" claim holds only under `HEADLESS_MODE`. It now also names `archive-kb.sh`'s two real discovery gaps — plans found by `*<slug>*` glob (misses topic-named plans, including #7373's), specs probed only at `specs/feat-<slug>` (misses `fix-*`; 27 live, zero archived) — so a reader verifies instead of assuming.

**tasks.md.** Its 18 boxes were checked by verifying each, not bulk-toggling; three that did not hold as written carry `[SUPERSEDED]`/`[PARTIAL]` annotations. Review caught one annotation asserting "two to four sites" — the shipped count is **five**, verified against `main`. Corrected, and the `266/267` figure anchored as-of #7373.

## What was removed before merge

This branch originally added a `check-kb-artifacts-archived` gate. A 4-agent report-only panel found it must not ship, and it is preserved on `wip-kb-archival-gate-rework` for rework rather than deleted:

- **It would have caused a new failure.** The carve-out is per-*file*, so a normal spec dir fails the gate, the agent archives the whole directory, and Phase 6 step 2.5 can no longer find `decision-challenges.md` — Model Dissents silently dropped from the PR body and the `action-required` issue never filed, defeating ADR-084. The gate's own T7/T8 encoded that as intended behaviour.
- **Its central claim was false.** It called itself "the mechanical version" of prose that failed twice. Every sibling Phase 5.5 blocking gate has a PreToolUse hook in `settings.json`; this had none, and its documented `<!-- gate-override: kb-archival -->` marker has no reader anywhere in the repo.
- **It failed open where CI runs.** Missing/unfetched `origin/main` skipped the primary detector and printed `OK` (reproduced on `--depth 1`, `actions/checkout`'s default); detached HEAD exited 0 unconditionally.
- **Its prescribed remedy could not fix what it flagged**, for both `fix-*` branches and topic-named plans.
- **~8 mutations survived its suite green**, and a live nested plan (`plans/feat-one-shot-reconcile-no-workspace-match/plan.md`) proves its `-maxdepth 1` blind spot is instantiated today.

The diagnosis worth keeping: its fixtures were derived from the SUT rather than from the repository, which is why the coverage holes clustered exactly where the implementation was wrong.

## Test plan

- [x] `lease-protects-active.test.sh` 17/17 (the only suite covering the touched `worktree-manager.sh`)
- [x] `components.test.ts` 1297/0 (skill-description budget, both SKILL.md edits)
- [x] `bash -n` clean on `worktree-manager.sh`
- [x] No test asserts the `--help` text that changed
- [x] Full `scripts/test-all.sh`
- [x] Zero references to the removed gate remain

## Known limits — not claimed fixed

`origin/main` carries **1,527 live-path spec dirs and 1,527 live-path plans**. This PR archives one feature's artifacts and corrects the claims; it does nothing about that backlog, and no mechanical gate prevents the next occurrence. Whether per-feature archival is a convention worth enforcing at all — given 27 live `fix-*` spec dirs and zero archived ones, i.e. a path that appears never to have worked — is an open question that should be decided before another gate is built.

### Plugin

- Corrected three sites claiming `cleanup-merged` archives spec directories (`merge-pr/SKILL.md`, `worktree-manager.sh --help`, and compound's consolidation note)
- compound now names `archive-kb.sh`'s slug-glob and `feat-`-prefix discovery gaps

## Cost

EUR 0.00/mo recurring.

## Net-issue-flow

0 filed, 0 closed, **NET 0**. No override marker and no `Closes` link — no open issue describes this cleanup.


deruelle:
The open question this PR raised in **Known limits — not claimed fixed** has been decided. Record: `knowledge-base/project/brainstorms/2026-08-10-kb-archival-convention-brainstorm.md` (branch `feat-kb-archival-convention`, PR #7398).

**Answer: per-feature archival is not worth keeping as a `git mv` convention, and should not be deleted either.** It buys exactly one thing git history does not — removal from `knowledge-base/INDEX.md`, which `generate-kb-index.sh` excludes `archive/` from and which `kb-search` Tier 1, `learnings-researcher` Step 0 and `learning-retrieval-bench.sh` all read. That benefit is attached to *index membership*, not to the move. Take it directly and retire the move.

Two corrections to the framing this PR left behind, both of which strengthen its conclusion:

1. **The convention is not fictional.** The `fix-*` half is exactly right and structural — `derive_slug()` strips `fix-`, then the probe is `specs/feat-${slug}`, so 27 live dirs are unreachable and zero archived. But `specs/archive/` holds 266 archived `feat-*` dirs, `plans/archive/` 283, `brainstorms/archive/` 123, still firing 7–75/month. It works on the happy path and silently no-ops elsewhere — which is a different diagnosis than "never worked", and implies fixing the mechanism rather than building one.

2. **This PR's own backlog figure understated the cost by roughly half.** INDEX.md as committed on `main` is stale by 3,711 rows — regenerated it is **7,507**, not 3,801. Filed as #7401.

Your revert reasoning holds and generalises further than stated: ADR-084 §5 needs `specs/<branch>/decision-challenges.md` readable until `ship` Phase 6, so the collision is between `git mv` and a live read — **any** gate on the move inherits it, not just the one you reverted. `wip-kb-archival-gate-rework` should not be reworked.

Scope now filed as #7399 (Tier 1: stop indexing `tasks.md`/`session-state.md`/`decision-challenges.md`, 2,408 rows, nothing moves on disk) and #7400 (Tier 2, deferred — needs a reliable merged signal). `archive-kb.sh` stays untouched; its retirement is Tier 2's exit criterion. The 3,054-artifact backlog is not migrated, because under index-exclusion there is nothing to migrate.

Nothing here blocks this PR — the archival and the three stale-claim corrections it ships all stand.

Issue #5934, opened 2026-07-03
infra: stop Concierge sandbox from materializing a char-device at .git/config.lock (durable root-cause fix for #5912 wedge)


deruelle:
## Upstream platform fix — stop the Concierge sandbox from materializing a char-device at `.git/config.lock`

**Companion to #5912** (repo-side targeted fix, branch `feat-config-lock-wedge-fix`, PR #5932).

### What
The confirmed root cause of the worktree-creation wedge (#5912) is that the Concierge
agent-sandbox filesystem/masking layer materializes `.git/config.lock` as a **character
device** (non-regular, write/remove-protected). This is NOT a stale git lock — a genuine
git config lock is always a regular file. Every `git config` write then fails `EEXIST`.

The repo-side fix (#5912) makes `worktree-manager.sh` self-heal past this in-session via a
lockless `atomic_git_config` writer. That is the pragmatic primary fix (in-session,
no operator SSH). **This issue tracks the only truly durable fix: stop the sandbox from
placing a device node at that path in the first place.**

### Why deferred / separate
- Lives in Concierge sandbox infra (mount/masking policy), NOT this repo.
- Has no in-session self-heal → cannot be the primary fix; the repo-side #5912 fix ships first.

### Re-evaluation criteria
- Determine whether the masking is single-path (`.git/config.lock` only) or glob (`*.lock`).
  This ALSO de-risks the #5912 repo-side fix, which assumes single-path (the temp file's
  `.lock` sibling must not be masked). Probe the sandbox mount config or the next fuller
  wedged-session forensic.
- If the masking layer can allow-list git's own lock acquisition (or not mask `*.lock`
  under `.git/`), that is the durable fix and #5912's fallback branch becomes dead-code
  insurance rather than a load-bearing path.

Lineage: #5912 (tracker), PR #5907 (instrument), the 2026-07-03 config.lock-wedge-fix
brainstorm.

deruelle:
Reopened: auto-closed on merge of PR #5935 by a commit-body `auto-close #5934` keyword adjacency (GitHub's close-parser reads squash-commit bodies). This issue is intentionally soak-gated — it must stay OPEN until the AC10 non-recurrence soak (`chardevice-wedge-nonrecurrence-5934.sh`, enrolled via the follow-through directive) PASSes 7 days post-deploy. The durable fix (sweep + forensic) shipped in #5935; ADR-081 status is `adopting` until the soak confirms non-recurrence.

[...]

PR #7407, opened 2026-08-10
fix(git-worktree): ensure_bare_config was dead code on the bare-repo-in-.git layout, so every worktree reported bare

### LINKED ISSUE #7394: worktree-manager: ensure_bare_config skips its own surgery on a bare-repo-in-.git layout, so every worktree reports bare
`worktree-manager.sh draft-pr` (and every other subcommand gated on `require_working_tree`) fails with **"Error: Cannot run from bare repo root (no working tree available)"** when run from inside a perfectly valid worktree.

Observed 2026-08-09 in `.worktrees/feat-one-shot-7307-main-health-monitor-silent-timeout`:

```
$ git rev-parse --is-bare-repository     # from INSIDE the worktree
true
$ git rev-parse --show-toplevel
fatal: this operation must be run in a work tree
```

Every worktree in the clone reported the same, so `IS_IN_WORKTREE` was false and `require_working_tree` refused.

## Cause

`core.bare = true` was present in the **shared** `.git/config` while `extensions.worktreeConfig = true` was set. Under that extension git reads `core.bare` from the per-worktree `config.worktree`; linked worktrees had an EMPTY one, so they inherited `bare = true` from the shared config and git considered them bare.

`ensure_bare_config` exists precisely to prevent this — its header says *"core.bare must ONLY exist in `.git/config.worktree`, never in `.git/config`"* — but its non-bare guard returns early before the surgery can run:

```bash
if [[ "$git_dir" == */.git && -d "$git_dir" ]]; then
  ... return 0
fi
```

The guard treats "`git_dir` is a `.git` **directory**" as a pure-filesystem proof of a normal clone. That is true for the Concierge workspace layout it was written for (#5934 round 6), but this repo is a **bare repo stored in a `.git` subdirectory** — `GIT_ROOT/.git` is a real directory *and* `core.bare=true`. So the guard matches, the surgery is skipped, and the shared-config value keeps bleeding into every worktree.

## Workaround applied

```bash
git config --file .git/config --unset core.bare
```

The bare root still reports bare (its own `.git/config.worktree` carries `bare = true`), and linked worktrees now resolve correctly. This is the script's own documented invariant, applied by hand.

## Suggested fix

Decide bare-ness from `core.bare` rather than from the gitdir's shape — or, keep the filesystem fast-path but fall through to the authoritative check when `core.bare` is true in the shared config. The round-6 comment explains the guard was added because `--is-bare-repository` degrades under a char-device config mask; a narrower condition (e.g. also requiring `core.bare` to be absent or false in the shared config) would preserve that protection without misclassifying this layout.

Worth a regression fixture: a bare repo whose gitdir IS `<root>/.git`, with a linked worktree, asserting `require_working_tree` succeeds from the worktree.

Discovered while implementing #7307 (it blocked `draft-pr` at pipeline start); filed separately as a defect in a different subsystem.

deruelle:
Closes #7394.

`worktree-manager.sh draft-pr` — and every subcommand gated on `require_working_tree` — refused with *"Cannot run from bare repo root"* when run from inside a perfectly valid worktree.

## Root cause, and where the issue's diagnosis was incomplete

`ensure_bare_config`'s round-6 guard decided "this is a normal clone, skip" from a **filesystem shape**: `git_dir` is a `.git` DIRECTORY. There is a third surface that test does not separate — a **bare repo stored in a `.git` subdirectory** (ADR-099 row 3, the operator's own clone), where `<root>/.git` is a real directory *and* `core.bare = true`. The guard matched it, so `ensure_bare_config` was **dead code** on that layout and nothing ever broke the config pair that wedges worktrees.

The issue names two conditions. Measured on **git 2.53.0**, the wedge needs a **conjunction of three**:

1. `extensions.worktreeConfig = true` in the shared `.git/config`
2. `core.bare = true` in the shared `.git/config`
3. the linked worktree's own `config.worktree` not setting `bare = false`

Removing **(1) alone** fixes it, and (1) is the condition the issue omits. That matters practically: on this clone the extension is currently *absent*, so the bug is **latent, not absent** — `worktree-manager.sh` was the only in-repo production setter of that key, so the tool being fixed was also the only thing causing it — and after this PR there is none.

This is not hypothetical. It already fired once: 2026-08-09 (#7332) took **all 14 worktrees down mid-rebase**, and `git-worktree/SKILL.md` has carried a manual repair for it since. That runbook prescribed by hand exactly the two repairs this PR automates.

## What changed

**The polarity is reversed.** The old code enabled `extensions.worktreeConfig` and unset `core.bare`; both halves are backwards.

- The extension is what makes git resolve `core.bare` per-worktree. `git worktree add` writes **no** `config.worktree` at all, so every worktree created while it was on started out inheriting `true` from the shared config — i.e. born broken.
- Unsetting shared `core.bare` (the workaround applied by hand while diagnosing) makes the bare **root** report as a normal working tree. Retired rather than codified.

The durable state is the one git itself produces: `core.bare` shared, extension off. Both remaining operations are *removals*, so **an already-healthy repo performs zero writes**.

**Bare-ness is now decided from `core.bare` read out of the shared config FILE**, never the gitdir's shape, still defaulting to SKIP. That preserves the #5934 round-6 mask protection by construction — under the char-device mask the read degrades to empty, which is not the literal `true` needed to fall through, so a masked repo takes the same skip branch as before.

**Two recovery paths**, because the create path cannot help worktrees that already exist: a `core.bare = false` seed into each new worktree's own `config.worktree` (bare repos only), and a detection-time self-heal that repairs the *current* worktree the moment git reports it as bare. The self-heal fires on a two-way conjunction, refuses a symlinked target, re-probes before claiming success, and writes solely to that worktree's own admin dir — **blast radius one**. The repo-wide normalization is the other function, and it runs only from `create` / `create-for-feature` / `cleanup-merged`.

**A retired-polarity repo is migrated, not skipped.** Repos the previous release normalized carry the extension with `core.bare` *absent*; the guard falls through on either signal, and restores `core.bare` **before** removing the extension, because until the extension is gone the root's own `config.worktree` is what stands in for it.

## Findings that changed the plan

- **A P0 in the plan's own prescribed fix.** `atomic_git_config`'s read-first fast path matched only the literal `--unset`, so the new `--unset-all` would have fallen through to the native writer. Measured, `git config --unset-all` on an absent key exits **5** — that would have failed `create`, `create-for-feature` and `cleanup-merged` on **every already-healthy repo**. Fixed first, before anything depending on it.
- **A falsified plan fact.** The plan claimed `git worktree add` leaves an *empty* `config.worktree`. It creates **none**. Asserted as non-existence; a 0-byte check would have been wrong.
- **A fourth sibling breakage the plan did not enumerate.** `stale-lock-diag` Test 8b asserts "fails loud on a held lock", but its fixture carried only `core.bare`, so under the reversed polarity the run performs zero writes and never contends the lock. Had the polarity change alone made it pass, it would have been passing *without any write ever touching the lock*. Its fixture now seeds the extension so exactly one write is attempted.
- **An inverted comment.** The file header claimed linked worktrees default to `core.bare=false`. The direction is the opposite, and believing it is why round 6's author saw no risk in skipping on a `.git` directory.

## What the 13-agent review changed

The panel found more than the original implementation contained, and the most severe item was introduced by the fix. Highlights, each reproduced before acting:

- **A data-destruction path.** The self-heal downgraded `IS_BARE` unconditionally and restored it only if a re-probe returned `true` — the probe this file documents as degrading under the config mask. A degraded answer left `IS_BARE` false with `GIT_ROOT` on a real working tree, where `cleanup-merged`'s non-bare tail runs `git reset --hard HEAD`.
- **The installed base was unreachable.** Three agents independently found that deciding on `core.bare == true` skipped every repo the previous release normalized. Fixed with the migration branch above.
- **A symlinked `config.worktree` wrote through to the shared config**, falsifying the blast-radius invariant and permanently disabling the guard afterward.
- **`branch=ok` asserted an unverified repair** — `atomic_git_config` returns 0 on its no-write fast path.
- **Silent arms.** Several skip paths emitted nothing, which is the same unobservable-no-op class that let the round-6 guard stay dead code for months.
- **Deleted the ancestor walk.** Four agents concluded it was redundant; one measured identical verdicts across seven repo shapes. Removing it plus caching the init-time probes took source-time subprocesses back to the pre-PR baseline.
- **Declined one finding, with reason.** Escalating the masked-config case to a wedge was implemented and reverted: under that degradation `GIT_ROOT` is the relative `.git`, and `git -C .git rev-parse --is-bare-repository` reports true for *any* normal clone, so it re-opened #5934 D3 on the production surface. Now reported via `reason=masked-cannot-determine` instead.
- **Corrected my own false claims:** `--get` returns rc 0 on a multi-valued key (not 2); `require_working_tree` has one call site so it was never the blast-radius mechanism; "both keys required to reproduce" is false (the extension wedges at `repositoryformatversion` 0); two stale `line-532` citations; and a retained comment that reasserted the exact premise this PR retires.

## Verification

git version **2.53.0** — every behavioural claim is scoped to it.

| Suite | Result |
|---|---|
| `worktree-manager-bare-in-dotgit-layout.test.sh` (new, 22 cases) | 66 / 0 |
| `worktree-manager-atomic-config.test.sh` | 82 / 0 |
| `worktree-manager-stale-lock-diag.test.sh` | 42 / 0 |
| `worktree-manager-bare-sync.test.sh` | 10 / 0 |
| `git-lock-marker-telemetry.test.ts` | 26 / 0 |
| `tsc --noEmit` (web-platform) | clean |

**Non-vacuity proven per-behaviour, not by whole-file revert.** Against the pre-fix script the suite fails **20** of 66. More usefully, six individual behaviours were mutated back on a sandbox copy against a green control and each turned the suite RED: the `IS_BARE` downgrade, the symlink guard, the retired-polarity migration, the post-write re-probe, the guard-skip marker, and `--unset-all`. Two fixes are pinned structurally rather than behaviourally because they only manifest under probe degradation; the test says so explicitly.

Case 13 is worth calling out: it measures the placement hazard directly. With the recovery arm placed *after* the `WORKTREE_DIR` assignment, `WORKTREE_DIR` resolved to `<root>/.git/worktrees/<name>/.worktrees` and every subsequent `create` landed inside the admin dir.

## Observability

Three markers registered, classified by `branch=` rather than by name, since one name covers both a recoverable and an unrecoverable outcome:

- `SOLEUR_GIT_BARE_POISON` — `branch=healed|clean`, always **mirrored, not paged** (the run proceeded).
- `SOLEUR_GIT_BARE_SELFHEAL` — `branch=failed` **pages**: the script then refuses every mutating subcommand from that worktree. `branch=ok` / `branch=skipped` do not.
- `SOLEUR_GIT_BARE_SEED` — the create-time pin, a **separate name** rather than a reserved branch value, so a future author tidying its branch string cannot re-arm paging on an inert pin.

The discriminator is anchored to the trailing field: `=` is legal in a git ref name, so an unanchored scan made a worktree named `branch=failed` page on a *successful* heal.

Observability layer 7 (customer-side CLI) → stdout markers → the existing `git-lock-marker-telemetry` PostToolUse hook → pino/Better Stack + Sentry. The existing sentinel drift guard derives its expectations from the shell script itself, so it began enforcing both new names with no edit.

## User-Brand Impact

Carried forward from the plan (now archived) so the ship-time gate can read it — the diff touches `apps/web-platform/server/`.

**If this lands broken, the user experiences:** every `git` write command in every one of their worktrees failing with `fatal: this operation must be run in a work tree`, and every `worktree-manager.sh` subcommand refusing with *"Cannot run from bare repo root"* — from inside a worktree that is manifestly valid. In-flight work is not lost but is completely inaccessible to tooling until they hand-edit git config, which a non-technical Soleur user cannot do. Not hypothetical: this happened on 2026-08-09 to 14 worktrees at once, mid-rebase (#7332).

**The review found a sharper version of this.** An earlier revision of the self-heal downgraded `IS_BARE` unconditionally and restored it only on a probe that degrades under the config mask. A degraded answer left `IS_BARE` false with `GIT_ROOT` pointing at a real working tree, where `cleanup-merged`'s non-bare branch runs `git -C "$GIT_ROOT" reset --hard HEAD`. That is **destruction of uncommitted work**, reachable from a session-start command. It is fixed (upgrade-only), and pinned by Case 21 plus a structural assertion, because the failing state is not portably synthesizable.

**If this leaks, the user's data is exposed via:** nothing — the change reads and writes only local git configuration (`core.bare`, `extensions.worktreeConfig`, `core.worktree`) plus two stdout telemetry markers carrying filesystem paths the platform already owns. No user content, no credentials, no new network surface, no new persisted field.

- **Brand-survival threshold:** `single-user incident` — one user losing their local working surface to a config key Soleur's own tooling wrote is brand-damaging on its own. **This is not only the founder's machine:** the self-hosted CLI plugin carrying `worktree-manager.sh` has been with an external alpha tester since 2026-08-06, and unlike the founder they have no way to diagnose a git-config wedge at all.

## Post-merge

**None.** The operator's repo normalizes on the next `create` / `create-for-feature` / `cleanup-merged`, and any already-wedged worktree self-heals on first use rather than waiting for that gate — which matters, because the session-start gate is an AGENTS.md rule rather than a hook.

## Deferrals

One, filed as required before ready: #7411 — shrink `ensure_bare_config` to what the measurements leave load-bearing. Not folded in here because it would restructure the function in the same PR that reverses its polarity, with a brand-new suite pinning the old shape. Net issue flow: closing 1, filing 1 → **net 0**.

## Changelog

- **Fixed:** a bare repo stored in a `.git` subdirectory had `ensure_bare_config` skip itself entirely, so `extensions.worktreeConfig` was never removed and every linked worktree reported as bare — breaking `draft-pr` and every other working-tree subcommand.
- **Fixed:** `atomic_git_config` now treats `--unset-all` idempotently; without it the new call would have failed worktree creation on every already-healthy repo.
- **Added:** create-time `core.bare = false` seed per worktree, plus a detection-time self-heal with a one-worktree blast radius.
- **Added:** `SOLEUR_GIT_BARE_POISON` / `SOLEUR_GIT_BARE_SELFHEAL` telemetry, paging only on an unrecoverable heal.
- **Docs:** ADR-173 records the polarity and its git-version scope; ADR-099 row 3 now names the root's `.git` directory and states that `[[ -d <root>/.git ]]` cannot discriminate it from a normal clone.


deruelle:
**rename-guard attestation** (`secret-scan-allow-rename`)

Four files move into gitleaks-allowlisted `archive/` paths. That is `/soleur:compound`'s standard artifact consolidation at the end of a feature — `git mv` of this session's own planning artifacts — not an import of outside content.

The control exists because a rename *into* an allowlisted path can smuggle a secret past scanning, so the attestation is backed by evidence rather than asserted:

| Evidence | Result |
|---|---|
| Secret-shape scan of all four files (PEM / `ghp_` / `ghs_` / `github_pat_` / `sk-ant-api03` / `sk_live` / `sbp_` / `xoxb-` / `AKIA` / `dp.ct.`) | **0 hits** |
| CI `gitleaks scan` | **pass** |
| CI `allowlist-diff (.gitleaks.toml paths surface)` | **pass** |
| Provenance — all four first committed in `08e495d5a` (this session, compound) | authored here, not imported |

Files: the `#7394` plan, plus `tasks.md`, `session-state.md`, and `decision-challenges.md` from this branch's spec directory.

Issue #7376, opened 2026-08-09
run-registered-suites.sh is flaky under its default -P on a 4-core runner (3 different suites across 2 of 6 executions)


deruelle:
`apps/web-platform/infra/run-registered-suites.sh` executes its 92 registered suites in parallel at `-P min(nproc, 6)` — i.e. `-P 4` on a 4-core GitHub-hosted runner. Measured across 6 executions of the **same suite set on an unchanged tree**, two failed, and they named **three different suites**:

| execution | parallelism | cores | result |
|---|---|---|---|
| [31328156153](https://github.com/jikig-ai/soleur/actions/runs/31328156153) tests-nested | -P 4 | 4 | 92 passed, 0 failed |
| [31328156153](https://github.com/jikig-ai/soleur/actions/runs/31328156153) infra-step | -P 4 | 4 | **90 passed, 2 failed** — `inngest`, `soleur-host-bootstrap-observability` |
| [31329991410](https://github.com/jikig-ai/soleur/actions/runs/31329991410) | -P 4 | 4 | 92 passed, 0 failed |
| [31331269522](https://github.com/jikig-ai/soleur/actions/runs/31331269522) tests-nested | -P 4 | 4 | **91 passed, 1 failed** — `web-host-provisioner-parity-mutation` |
| [31331269522](https://github.com/jikig-ai/soleur/actions/runs/31331269522) infra-step | -P 4 | 4 | 92 passed, 0 failed |
| local | -P 6 | 16 | 92 passed, 0 failed |

The two runs of run `31331269522` are the sharpest evidence: the **same 92 suites, in the same job, minutes apart** — one execution reported `91 passed, 1 failed` and the other `92 passed, 0 failed`.

## Why this is a harness defect, not three suite defects

- Different failure sets on an unchanged tree is the repo's own documented litmus for a harness defect (`work/SKILL.md`: "three runs of an UNCHANGED tree giving three DIFFERENT failure sets is a harness defect, never a re-run to repeat").
- Each named suite passes when run individually (`inngest.test.sh` rc=0, `soleur-host-bootstrap-observability.test.sh` rc=0).
- The same 92 suites pass **sequentially** in `infra-validation.yml`'s `deploy-script-tests` job, which is green on `main`.
[...]

Issue #7076, opened 2026-07-30
run-registered-suites.sh derivation misses 8 registered infra suites (7 subdir + 1 sudo-bash)


deruelle:
Surfaced while implementing #7068 (registering the seven orphan infra suites). **Pre-existing on `main`, not caused by that PR** — filed under `wg-when-an-audit-identifies-pre-existing` rather than folded in, which is the same rule #7068 itself was filed under.

## What

`apps/web-platform/infra/run-registered-suites.sh` claims in its own preamble that "this runner and CI cannot drift". **They do drift, by 8 suites.** The runner derives its execute set from `run: bash …` steps in `.github/workflows/infra-validation.yml`, and two shapes escape that derivation:

1. **The 7 subdirectory suites.** The derivation's character class `[A-Za-z0-9._-]` excludes `/`, so every suite under a subdirectory is invisible despite carrying a proper single-line `run: bash …` step:
   ```
   apps/web-platform/infra/inngest-rls/apply-inngest-rls-dev-workflow.test.sh
   apps/web-platform/infra/inngest-rls/inngest-rls-mutation.test.sh
   apps/web-platform/infra/inngest-rls/inngest-rls.test.sh
   apps/web-platform/infra/scripts/gen-github-egress-cidr.test.sh
   apps/web-platform/infra/scripts/sigpipe-triage-feasibility.test.sh
   apps/web-platform/infra/supabase-advisor/scan-workflow-mutation.test.sh
   apps/web-platform/infra/supabase-advisor/scan-workflow.test.sh
   ```
2. **`workspaces-luks-loopback.test.sh`**, invoked as `sudo bash …` inside a `run: |` block (`infra-validation.yml:688`), which the single-line `run: bash` anchor cannot match.

Narrow derivation = **79**. Permissive = **87**. (Post-#7068 those become 86 and 94.) CI runs all 8; the local runner runs none of them.

Measured identity, useful as the fix's landing target: **`79 derived + 7 orphans + 8 invisible = 94 on disk`**, exactly.

## Why the naive one-line regex widening is WRONG

This is the load-bearing part, and it is why this is its own issue rather than a one-line fix.

[...]

PR #7423, opened 2026-08-10
fix(infra): give the parallel suite runner diagnostics on RED, and close its false green

deruelle:
## Summary

`apps/web-platform/infra/run-registered-suites.sh` runs 93 registered infra suites in parallel and produced **different failure sets on an unchanged tree** — the repo's own litmus for a harness defect. It also discarded every suite's output (`bash "{}" >/dev/null 2>&1`), so a CI failure carried **no diagnostic at all**: issue #7374, filed by the health monitor, is 21 `PASS` lines and a count, naming no suite. Characterising the flake took eight executions.

This PR ships three things and deliberately declines a fourth.

**1. The instrument.** Each suite's stdout+stderr now goes to its own file; on RED the parent prints a bounded excerpt with `rc`, `elapsed` and a start offset. The excerpt is **anchored on the suite's own failure marker**, not a blind tail — these suites do not stop at the first failed assertion, so a failure at arm 5 of 43 is hundreds of lines from EOF and a tail shows only trailing passes. The marker ERE is corpus-derived: measured across all 93 suites, the obvious `^\[FAIL\]` covers **11**, the shipped ERE covers **93**.

**2. A pre-existing false green.** A child whose wrapping shell is killed emits neither `PASS` nor `RED`, so `RED` stayed 0 and the runner **exited 0** while printing e.g. `91 passed, 0 failed (of 93)` — two numbers visibly disagreeing with nothing asserting they must match. A suite could vanish and the gate went green. `PASS + RED` must now equal the derived count, vanished suites are named as `UNACCOUNTED`, and their captured output is dumped.

**3. A collision provable by reading.** `run-registered-suites.test.sh` is itself a registered suite, so it ran concurrently with the other 92 while creating and deleting a fixture **inside the live infra directory** that `credential-persist-home-guard.test.sh` was copying and `diff -rq`-ing against the still-live source. The orphan scan now takes an injected candidate list, and the comparison side is a frozen snapshot.

**4. Not claimed: the capacity (H2) and polling-deadline (H4) hypotheses remain UNKNOWN.** See "What was and was not established".

**Ref #7376 — deliberately NOT `Closes`.** This is the *instrument* half. The plan's acceptance
criteria 10-11 (measured H2/H4 verdicts + a 20-run proof-of-fix) are **not satisfied**, so the
flake is not proven fixed and auto-closing on merge would close the issue on a promise. The plan
anticipated this split: *"if the Split Trigger fires, ACs 1-9 gate the instrument PR; ACs 10-11
move to the follow-up fix PR."* Strictly the trigger's own precondition (≥20 non-reproducing
iterations) was **not reached** — I ran 2 and stopped for the cost reason below, so I cannot claim
it fired by measurement. The disposition is the same either way. #7376 closes when a real CI
occurrence is attributed; Tracks #7432.

Net-issue-flow override. The gate reports `Filing: 1 (#7429) / Closing: 0 / Net: +1`. Both halves
of that need stating accurately:

- **#7429 is not this PR's filing.** A sibling session filed it, and it cites `#7423` only as its
  *reason for deferring* — it declined to edit `run-registered-suites.sh` inline because this PR
  had a live session on that file. The gate's FILED query matches any issue body containing the
  PR number, which is deliberately broad, so it attributes a sibling's cross-reference to me.
  Nothing to inline here: it is a different subsystem (`test-all.sh` signal-shape propagation)
  and correctly its own issue.
- **This PR's own filing is #7432**, which the gate does not see because its body references
  #7376 rather than #7423. It consolidates two follow-ups that cannot be inlined: removing the
  `JOBS: 1` stopgap is blocked on a soak of *this* fix (a temporal dependency, not a size one),
  and a `| grep -q` linter would guard a mechanism measured to be unreachable today (~12x under
  the 64 KiB threshold), needing a mass rewrite or a baseline file.

Closing is 0 because #7376 stays open deliberately (see above). Closing it here to balance the
arithmetic would be the dishonest move the gate is not asking for.

## What was and was not established

**Established by measurement:**

- H1 (`pipefail` + `producer | grep -q` → SIGPIPE) is **REFUTED**. The mechanism reproduces, but the threshold is the 64 KiB pipe capacity and every real call site is ≤5 KB — ~12× under the bar, with zero above-threshold streaming sites in the repo. It had in-repo documentation, a prior fix inside one of the failing suites, ~450 candidate sites and a flawless narrative fit. Committing to it would have bought a mass rewrite, a new linter, and a still-flaky runner. Written up in the plan rather than deleted, so it is not rediscovered.
- The live-tree collision (H3, one instance) is **CONFIRMED by reading** and fixed.
- The false green is **CONFIRMED by reading** and fixed.

**Not established — H2 (capacity) and H4 (polling deadline) remain UNKNOWN.**

The reproduction I failed to obtain, and why: the plan asked for a 20-run pre-fix baseline plus 20 post-fix, locally under `taskset -c 0-3`. One run costs **23 minutes**, so that is ~7.75 hours — and the plan itself records that `taskset` bounds **CPU only**, not RAM, disk or `/var/tmp` headroom, so a local non-reproduction establishes nothing about a hosted runner. I ran 2 clean pinned runs (93/93 green, 1395s and 872s). A third run reproduced 12 REDs, but **that run is void**: I edited the runner while it held it open, which deleted its log dir mid-run. Those REDs were mine, not the flake's.

The honest pre-fix baseline is CI's own record on the hardware where the flake occurs: **8 executions, 4 failed, 6 distinct suites implicated**. Post-fix, `test-all.sh` is green at 288/288 with the infra runner nested and covered.

So this PR does not claim the flake is fixed. What it buys is that the next occurrence **self-reports** instead of costing eight executions to characterise.

## User-Brand Impact

- **Artifact exposed:** none directly. The new diagnostic output is the only new distribution surface — per-suite excerpts now reach a world-readable GitHub Actions run log (90-day retention) where previously they went to `/dev/null`.
- **Exposure vector:** the health monitor posts an excerpt into a **public** GitHub issue. Mitigated architecturally: every dumped line carries a `SOLEUR| ` sentinel that the monitor filters out of the issue body, so the published excerpt stays byte-equivalent to today's. The redaction set was widened for the run-log surface (JWTs, Doppler service tokens, AWS keys, presigned URLs, Stripe restricted/webhook keys, Sentry DSN + tokens, bare `Authorization:` headers) and the PEM rule was rewritten.
- **Brand-survival threshold:** `none`.
- `threshold: none, reason: the change is confined to a CI/local test harness and its suites; it reads no user data, has no runtime path into the product, and creates no store or connection.`

## Review

8 agents, ~40 findings, **all fixed inline — zero scope-outs filed** (net issue flow: filing #7432, closing nothing → **net +1**, overridden with justification — see above).

Coverage note, stated plainly because the trailer's label does not: `Reviewed-Coverage` reads `full 8/8` but **three canonical code-class agents were substituted, not merely absent** — `performance-oracle`, `data-integrity-guardian` and `agent-native-reviewer` were replaced with `observability-coverage-reviewer`, `test-design-reviewer` and `code-simplicity-reviewer`. A bash test-harness change has no DB, no runtime perf path and no agent-facing surface; the substitutes matched the actual risk. `semgrep` was skipped per the bash-only rule and `shellcheck` substituted (all 5 changed scripts clean).

Two P1s, each found independently by two agents and reproduced before fixing:

- **The accounting verdict never reached the operator.** It was emitted inside the sentinel-prefixed block, so the monitor's own filter stripped it — and because a vanished suite emits no `RED ` line either, the monitor titled the issue *"did not complete … usually a timeout"*, a cause it never measured, on the exact failure the assertion exists to catch. That is the AP-021/ADR-166 defect #7371 removed, re-armed by its own fix.
- **The log dir leaked on every non-green run** (414 dirs / 23 MB measured locally, self-amplifying because this suite drives the runner ~10× per invocation).

Three security defects, all reproduced:

- The PEM rule I rewrote introduced a `sed` range that **consumes to EOF** on an unterminated `BEGIN`, deleting the summary line, run URL and every suite name — with a live trigger already in the corpus (`git-data-emit.test.sh:266`).
- The widened Doppler rule missed the real `dp.st.<config>.` shape; my own test used `dp.scim.`, which matches — the fixture dodged the gap it existed to find.
- Caps were lines, not bytes: one 200 KB line breaks `gh issue create`, so the monitor would file nothing.

Three test vacuities found by mutation, not reading:

- **Two of seven mutation rows scored kills against an UNMUTATED runner** — `ok()` prints the assertion id, so the scorer substring-matched the *passing* line. A noop-control row now proves the scorer cannot match noise.
- The fixture put its marker on log line 1, so a `head -n 5` selection mutant survived the entire suite.
- The excerpt invariant was `1 == 1` by construction (the fixture used a `FAIL:` shape no monitor anchor matches).

## Known gap, now ratcheted

An independent, more permissive oracle shows **8 suites registered in CI that this runner never derives** — 7 in subdirectories the derivation ERE cannot match, plus one registered as `sudo bash`. This is pre-existing and already tracked by #7076; this PR does not fix it, but T2e pins the set as a **shrink-only ratchet** so a 9th cannot land silently, and the runner's false "runner and CI cannot drift" header claim is corrected.

### Infra test harness

- `run-registered-suites.sh` captures per-suite output and prints a bounded, marker-anchored excerpt on RED with `rc`, elapsed and start offset.
- Closes a false green: `PASS + RED` must equal the derived suite count; vanished suites are named `UNACCOUNTED` and their output dumped.
- Log dirs are reaped from the EXIT trap plus an age-reap of older siblings (ADR-133's reaper scopes to `/tmp`; this runner uses `/var/tmp`).
- New `SOLEUR_INFRA_DIR` / `INFRA_ORPHAN_LIST` test seams so fixtures never touch the live tree.
- `credential-persist-home-guard.test.sh` diffs against a frozen snapshot instead of the live source.

### CI

- `main-health-monitor.yml` filters the diagnostic sentinel out of the public issue body, widens its anchor to name vanished suites, points the operator at the run log, and widens the redaction set.
- `deploy-script-tests` timeout re-derived 12 → 14 min (suite 3s → 67s).

### Docs

- ADR-133 records two departures: this instrument is **not** observe-only (186 files/run), making it a confound for the very capacity hypothesis it measures; and the ships-ahead ordering was departed from while honouring the property that rule protects.

## Test plan

- [x] `bash scripts/test-all.sh` — **288/288 suites passed**, epilogue confirms `apps/web-platform/infra/ IS covered`
- [x] `run-registered-suites.test.sh` — 50/50, 9 mutation rows all killed incl. the noop-control
- [x] `credential-persist-home-guard.test.sh` — 40/40, with a positive control proving the race reproduces
- [x] `main-health-monitor-workflow.test.sh` — 25/25, redaction verified against a synthesized corpus
- [x] `workspaces-luks-header.test.sh` — 62/62
- [x] `shellcheck` clean on all 5 changed scripts; `actionlint` clean
- [x] Mutation-proven: removing the monitor filter, widening the sentinel, dropping the JWT rule, and reverting PEM to header-only each red a **named** assertion
- [x] 2 clean pinned `taskset -c 0-3` runs at default parallelism (93/93)
- [ ] ⏳ H2/H4 attribution from a real CI occurrence — the instrument now makes the next one self-reporting (Tracks #7432)


deruelle:
**rename-guard attestation.**

The three flagged renames are compound's standard artifact archival — a pure `git mv` of this feature's plan and spec into `archive/`, with no content change:

- `plans/2026-08-10-fix-infra-suite-runner-parallel-flake-plan.md` → `plans/archive/…`
- `specs/feat-one-shot-7376-…/session-state.md` → `specs/archive/…/session-state.md`
- `specs/feat-one-shot-7376-…/tasks.md` → `specs/archive/…/tasks.md`

The property this gate protects is that a rename into a gitleaks-allowlisted path cannot smuggle a secret past the scanner. Verified before applying the label rather than after: all three files scan clean against the canonical secret-shape set (PEM headers, `ghp_`/`ghs_`/`github_pat_`, `sk-ant-api03-`, `sk_test`/`sk_live`, `sbp_`, `xoxb-`, `AKIA`, `dp.{pt,st,sa,ct}.`, and JWT `eyJ…` triplets). They contain plan prose and a task checklist only.

Labeled rather than amended with a `Rename-Allowed-By:` trailer because the commits are already pushed and auto-merge is queued — a history rewrite here would be the riskier of two sanctioned options.

Issue #7462, opened 2026-08-11
infra: restore the dedicated inngest host — replace, diagnostic boot, cutover window


deruelle:
Sequenced restoration of the dedicated inngest host (10.0.1.40), deferred out of the #7228 detection/safety PR because every step below needs either a host replace or a maintenance window.

## Why this is one issue and not four

These are not independent tasks — each is the precondition of the next, and doing them out of order is the failure mode. In particular the replace is the operation that DISARMS the old cutover latch, so the monotonic latch has to be on the host before a replace is dispatched, not after.

## Preconditions (satisfied by the detection/safety PR)

- [x] Monotonic, append-only cutover latch on surviving storage — the replace must not authorize a FLUSHALL against the preserved AOF volume
- [x] Consumer-side detection live on web-1, independent of this host booting at all

## The sequence

- [ ] **1. Dispatch the replace.** `apply_target=inngest-host-replace`. This is the ONLY delivery path for cloud-init/bootstrap changes to 10.0.1.40 (`hr-prod-host-config-change-immutable-redeploy`). Do not dispatch until the checkbox above is on `main`.
- [ ] **2. Diagnostic boot.** The replaced host comes up against a NON-PROD `INNGEST_POSTGRES_URI`, so the flip guard's prod detection yields `is_prod=false` and takes its ALLOW arm. This is mandatory, not optional: the brake is set (`INNGEST_CUTOVER_FLIP=rollback`, measured in run 31486949232 on 2026-08-11), and `rollback` is outside the guard's `{armed, flipping, flushed, done}` allowlist, so a prod-URI start is REFUSED. Without the non-prod backend the replaced host can never attempt a bind and every hypothesis stays UNKNOWN forever.
- [ ] **3. Read the `net-health` marker.** It carries `bind=`, `priv8288=` and `nft=` — the whole diagnosis in one row. This is the measurement that decides between the open hypotheses (nftables / routing / application / bootstrap-never-enabled / doppler symlink). The original 2026-07-30 trace is irrecoverable: Better Stack retention is ~3 days, measured.
- [ ] **4. Fix whatever the trace names.** Deliberately not pre-written.
- [ ] **5. Cutover window:** quiesce-web → confirm → arm → verify. Atomicity is structurally unavailable across two disjoint control planes, so accept a GAP, never an overlap. Gated behind the `inngest-cutover` required-reviewer environment.
[...]

Issue #7228, opened 2026-08-03
incident: inbound-email dispatch dead ~3 days — app dispatched at a host that never bound :8288, health probe certified a different server


deruelle:
Tracking issue for the inbound-email dispatch outage. Filed from `tasks.md` task 0.1 of `feat-one-shot-resend-inbound-webhook-500`, which the #7144 nine-agent review flagged as still unchecked while its own finding 7 recorded the alarm as "absent AND untracked".

**Label note:** task 0.1 specifies `type/incident`. That label does not exist in this repo — the taxonomy is `type/{bug,chore,feature,question,security}`. Filed as `type/bug` + `priority/p1-high` rather than minting a sixth type label unilaterally. If an incident taxonomy is wanted, that is a deliberate governance call, not a side effect of this filing.

## Evidence

- Every `inngest.send()` returned `connect ECONNREFUSED 10.0.1.40:8288` from `CONTAINER_NAME=soleur-web-platform`. A Better Stack pull during the review returned 30 rows in a 2-minute window (2026-08-02 12:25:56 → 12:27:57), including a non-email dispatch (`"actionClass": "engineering.pr_review_pending"`) — so the failure was **fleet-wide**, not scoped to the inbound-email route.
- The ADR-100 cutover repointed the app at the dedicated host on 2026-07-24, but that host never bound `:8288`.
- `scheduled-inngest-health.yml` reported SUCCESS every ~1–2h throughout, because it was certifying the **surviving co-located server** rather than the app's actual dispatch target.
- Sentry measured **zero** issues for `op:inngest-send` across the entire outage. The route-level captures fired immediately after a `logger.error({ err })` whose pino hook had already mirrored the same `Error` instance, marking it caught — so the tagged call early-returned and the tag never landed.

## Detection

Our own `cron-email-ingress-probe` first fired **2026-07-31T06:15:24Z**; Resend reported at **18:23 UTC**. Detection preceded the vendor by ~12h. This was a **response** failure, not a detection gap — which matters, because it points remediation at an operator-reaching channel rather than at more signal.

## Remediation status

Landed on `feat-one-shot-resend-inbound-webhook-500`:

- **finding 8** — the probe derivation read the wrong `INNGEST_BASE_URL` duplicate.
- **finding 1** — a third dispatch write site in `cloud-init.yml` was never swept.
[...]

Issue #6500, opened 2026-07-15
P1: enroll soleur-inngest as a zot client — adding ZOT_* trips the 5/5 boot-isolation floor (was: hard-pins GHCR with no zot path)


deruelle:
## The 7th GHCR-served path — and the soak cannot see it

`scripts/followthroughs/zot-soak-6122.sh` documents **six** ways the fleet can still be GHCR-served. There is a **seventh**: the dedicated inngest host. It is **live**, it is **GHCR-only**, its pull is **fail-closed**, and it reports to a system the soak does not query.

**Consequence: ADR-096 task 5.3 revokes the GHCR PAT ⇒ the next fresh inngest-host boot 401s ⇒ `exit $pull_rc` ⇒ the host never comes up — while the zot soak reports PASS.** 5.3–5.5 are irreversible (they rotate *and* revoke the PAT), so there is no rollback.

Surfaced during planning for #6462. Out of scope there (different host, different file, #6122 Phase 5 scope) — but #6462 **hard-wires this issue's number into the soak's C1 blocker arm**, so the soak cannot `exit 0` while this is OPEN.

## Evidence (re-verified against `main` 2026-07-15)

**1. GHCR-only pin, no alternative** — `apps/web-platform/infra/cloud-init-inngest.yml:337`
```
    IREF=ghcr.io/jikig-ai/soleur-inngest-bootstrap:v1.1.19
```
A case-insensitive sweep for `zot|ZURL|ZIREF|/v2/|soleur-boot-emit` across the whole file returns **0 hits**. No zot primary, no `/v2/` probe, no fallback.

**2. The pull is fail-closed** — `:349`
```
    [ "$pull_rc" -eq 0 ] || { echo "FATAL: OCI pull failed rc=$pull_rc" >&2; exit "$pull_rc"; }
```

**3. Baked token, fail-open login feeding that fail-closed pull** — `:260`
```
      if printf '%s' "$GHCR_READ_TOKEN" | docker login ghcr.io -u "$GHCR_READ_USER" --password-stdin >/dev/null 2>&1; then
```
The token is **baked into user_data** at render time (`:245`), not fetched at runtime. The login is fail-**open** (`set +e`, `:256`; emits `ghcr-login-FAILED` and continues). So a revoked PAT does not fail at login — it degrades silently into a 401 at the fail-**closed** pull. Abrupt, not diagnosable.

[...]

PR #7516, opened 2026-08-12
fix(inngest): give the dedicated host a zot-primary bootstrap pull path

deruelle:
The dedicated inngest host (`10.0.1.40`) cannot bootstrap. Its cloud-init hard-pins a `ghcr.io` ref for the `soleur-inngest-bootstrap` OCI image, the GHCR read credential is revoked, and the pull 401s six seconds into every boot. `inngest-bootstrap.sh` ships *inside* that image, so nothing installs, nothing binds `:8288`, and every app-originated dispatch fails.

Every other pull site in the fleet was flipped to prefer the self-hosted zot registry in the ADR-096 Phase-3 dark launch. This file was missed — ADR-096 names it as the gap and predicts this exact failure, including that the PAT would stop working ahead of the Phase-5 task meant to retire it deliberately.

Ref #7462 (step 4) · Ref #7228 · Ref #6500 — **all stay OPEN**.

Plan: `knowledge-base/project/plans/archive/20260813-110650-2026-08-12-fix-inngest-zot-primary-bootstrap-pull-plan.md` (archived by `/compound`; carries the `## Observability` block and the two review addenda).

> #6500 is titled *"enroll soleur-inngest as a zot client — adding `ZOT_*` trips the 5/5 boot-isolation floor"*. This PR is its **code half**, and its title names the exact obstacle that forced the bake mechanism. It stays open deliberately: closing it is an *authorization act* gating ADR-096 Phase 5.3–5.5 (the soak's blocker arm reads its state via `gh`), and code landing is not delivery.

## Phase 0 preconditions — both hold

**The pinned digest is in zot.** Re-probed **live**, not read off the prior record: `build-inngest-bootstrap-image.yml` run [31681702541](https://github.com/jikig-ai/soleur/actions/runs/31681702541) (`mirror_only=true`) reported `ghcr=sha256:6cdaa63d…a0f8 zot=sha256:6cdaa63d…a0f8` — identical to each other and to the pin — plus `PASS: 127.0.0.1:5000/…` from `crane validate --remote`.

That distinction is the point: zot runs gc + dedupe, and the workflow's own comment records that a reclaimed layer yields **green digest parity and `blob unknown` on the host's `docker pull`**. `crane validate --remote` fetches every blob. A manifest HEAD would not have answered it.

**`10.0.1.40 → 10.0.1.30:5000` is permitted.** Same subnet (`hcloud_server_network.inngest` / `.registry`); `hcloud_firewall.registry` has zero inbound rules and Hetzner firewalls filter only the public interface; `inngest-nftables.sh` declares an `input` chain only, so egress is unrestricted.

## Why the mechanism diverges from the web host

The web host reads `ZOT_*` from **Doppler at boot**. Unavailable here: this host's Doppler token is scoped to project `soleur-inngest`, and adding the keys to `soleur-inngest/prd` would break its fail-**closed** boot isolation self-check (`n_total != n_inngest` → `FATAL`, no boot at all) — precisely what #6500's title records. So the config is **baked as terraform template vars**, the precedent this host already sets for `ghcr_read_*` and `betterstack_logs_token`.

All three values come from **this same root** — `local.registry_endpoint`, `local.zot_pull_user`, `random_password.zot_pull`. No new root variables.

## What changed

- **`inngest-host.tf`** — bake the endpoint and credential from the in-root owners.
- **`cloud-init-inngest.yml`** — allowlist zot as an insecure registry and **restart** docker (the `docker.io` package starts the daemon during `packages:`, before `runcmd`, so `enable --now` cannot re-read the config); log in to zot **before** the pull; resolve the effective ref once, zot-primary, carrying the `sha256` pin unchanged onto both legs; bound both zot calls with `timeout`; emit `inngest_zot` / `inngest_ghcr_fallback` / a per-leg total-failure marker; the zot token joins `inngest-redact.sh`'s known-value list.
- **Guards** — `cloud-init-inngest-bootstrap.test.sh` (Guard 1) and a new mutation battery, registered in `infra-validation.yml`.
- **`zot-soak-6122.sh`** — widened anchors + corrected header (see below).

## The review round found more than the implementation

An 8-agent panel found defects in the PR's own guards and in three claims the PR published. Recorded here rather than quietly fixed, because two were assertions made to the operator.

**The guard certified placement, not behaviour.** Inverting **one character** in the arm's gate — `[ -n "$ZOT_EP" ]` → `[ -z … ]`, which runs the zot leg only when zot is *unconfigured* — left the guard at **117/117** and the battery at **9/9**. The assertion grepped the whole file for a token occurring three times. Now anchored positionally and mutation-proven.

| Defect | Consequence |
|---|---|
| Battery attribution grepped the log for the assertion *description*, which `assert()` echoes on PASS as well as FAIL | 7 of 9 rows reported KILLED with their named assertion neutered; one misrouted on the unmutated tree. "9/9 killed" did not mean what it said |
| Comment strip `sed 's/#.*$//'` destroyed executable lines — `#` is sed's delimiter in this file's redaction expressions | The phone-home's and `inngest-redact.sh`'s entire scrub patterns were deleted from what every assertion reads |
| `grep -c` counts LINES; no floor over the section's own assertions | Two literals on one line evaded the "one resolution" guard; deleting **all** Guard-1 assertions exited 0 |
| Four `producer \| grep -q` sites under `pipefail` | SIGPIPE (141) on an early match = false NEGATIVE, presenting as an unreproducible flake |

**Three published claims were false:**

1. **"GHCR retained as break-glass."** AP-016 **LAPSED 2026-07-30 (#7071)** — the PAT is REVOKED, so that leg returns a guaranteed 401. **After this change the host's boot depends entirely on zot reachability** plus a baked credential with no refresh channel. "Never a worse boot than today" holds only because today's path is 100% broken.
2. **"Each runcmd `- |` item is its own shell."** cloud-init's `shellify()` concatenates them into ONE `/bin/sh` — stated in `cloud-init.yml` twice, `nic-wait-gate.test.sh`, and the 2026-07-06 errexit-leak post-mortem. A `docker login ""` defect reported earlier was therefore **misdescribed**; the fix is right, the reason was not.
3. **"One query covers both hosts."** `soleur-boot-emit` reaches Sentry only, this emitter Better Stack only. No single query sees both.

**A design defect the plan did not anticipate.** Phase 2 added two operator-supplied `TF_VAR_*` for values `zot-registry.tf` already owns. That goes stale on rotation (reintroducing the very "baked credential with no refresh channel" failure #7462 exists to fix), created the whole-apply hazard the plan documented *and accepted*, and broke every `terraform test` run block. Both variables deleted.

**This PR falsified a downstream gate.** `zot-soak-6122.sh`'s corroboration arm anchors on the web host's syntax; measured against this implementation both legs return **zero** hits, so closing #6500 would have made the ADR-096 Phase-5 blocker fail **forever** on a false premise. Anchors widened; its harness case 7 (which relied on the real repo still being GHCR-only — corpus accident) now uses a synthetic fixture.

## Verification

| Gate | Result |
|---|---|
| Guard suite | **121/121** |
| Mutation battery | **9/9 killed**, attribution scoped to `^  FAIL` |
| Soak harness | OK |
| `terraform fmt -check -recursive` / `validate` / `test` | clean / valid / **3 passed, 0 failed** |
| Template render + `cloud-init schema` | 9/9 files ok |
| `bash -n` + `dash -n` | all runcmd blocks clean |
| Review coverage | `full 8/8 agents` |

CI caught one thing the local gate structurally could not: adding required no-default root variables breaks every `terraform test` run block while `validate` stays green, because it never resolves variable *values*. Local `fmt -check` also lacked `-recursive`.

## Delivery

**The zot arm itself is inert at merge.** `hcloud_server.inngest` is in no `-target` on the per-merge apply path, so nothing in `cloud-init-inngest.yml` reaches the host on this merge. Delivery is the gated **`apply_target=inngest-host-replace`** dispatch, out of scope here and belonging to #7462 steps 4–6.

Note the merge is **not** inert overall — see the `[ack-destroy]` section below. That is a separate, deliberately-authorized production apply covering pre-existing main drift (#7539), not delivery of this PR's arm.

The cutover-sequence arm step in #7462 additionally requires `INNGEST_DIAGNOSTIC_BOOT` to be **cleared** from `soleur-inngest/prd`. It is currently **set**, and this PR deliberately leaves it set.

## Production apply — `[ack-destroy]` carried deliberately

Main's `Apply web-platform infra` is red on the destroy-guard (#7539), and this PR touches
`apps/web-platform/infra/**`, so its merge fires that apply. The merge commit carries
`[ack-destroy]` on **explicit operator authorization**, given after reviewing the exact plan.

It authorizes, on production:

- create `betteruptime_heartbeat.inngest_consumer`, `doppler_secret.inngest_consumer_url`,
  `terraform_data.inngest_consumer_probe_install`,
  `github_repository_environment_deployment_policy.web_platform_infra_apply_main`
- update `cloudflare_bot_management.soleur_ai`
- **replace** `terraform_data.journald_persistent` → re-runs create-time provisioners against
  web-1: `systemctl restart systemd-journald`, `journalctl --flush`, re-render `vector.toml`,
  restart the Vector agent

Per #7539 the resource has **no destroy-time provisioners** and self-guards with positive
assertions (render-sanity gate before the live agent is touched, `test -d /var/log/journal`,
`journalctl --header` check, Vector liveness) — a botched render fails the apply while the
running agent stays up.

This unblocks #7228's boot-trace, which has been file-only-never-delivered since #7457.
Post-merge verification is via the cat-deploy-state webhook (`vector_journal_tail` non-empty,
`journald_storage.persistent=true`) — no SSH.

**Net-issue-flow override.** The gate counts #7539 and #7542 as filed by this PR. They were not:
both were filed independently by the operator and merely *reference* `#7516` — #7539 designates
this PR as the vehicle for the ack above, and #7542 is its follow-on. This PR files zero issues
and closes zero; the +2 is an artifact of the gate's bare-`#N` FILED query matching inbound
references. No issue is being deferred here.

## User-Brand Impact

- **If this lands broken, the user experiences:** nothing new at merge — this PR's arm is inert until separately delivered. If delivered broken, the dedicated inngest host stays dark exactly as it is now, so app-originated dispatch (inbound email, PR-review events) keeps failing silently. There is no post-change state worse for a user than today's.
- **If this leaks, the user's data is exposed via:** the zot pull credential baked into `user_data`, retrievable through the Hetzner metadata API by anyone with host access — the same trust boundary the existing `ghcr_read_token` and `betterstack_logs_token` bakes already occupy. Read-only, private-net registry, no widening.
- **Brand-survival threshold:** `none` — the failure mode of the change is "stays dark", not "new user-facing breakage".

*Scope-out override:* `threshold: none, reason: every sensitive path this diff touches is boot-time registry resolution on a single host that is 100% non-booting today, never an app-serving surface.*

Corrected from the plan's original reasoning: the `none` threshold does **not** rest on "GHCR retained as break-glass" — AP-016 lapsed 2026-07-30 (#7071) and that PAT is revoked. It rests only on the narrower claim above. After delivery the host's boot depends entirely on zot reachability plus a baked credential with no refresh channel; that risk shift is recorded in the ADR-096 amendment.

## Soak follow-through — enrolled, not promised

The ship gate blocked `gh pr ready`: this PR declares a post-deploy soak close-criterion and none of #7462, #7228, #6500 carried a `follow-through` label or a sweeper directive. The gate was right — the soak was prose. Two probes, because the two criteria differ in kind:

| Probe | Trackers | PASS requires |
|---|---|---|
| `inngest-zot-boot-7462.sh` | #7462, #7228 | `stage=inngest_zot` **and** `stage=bootstrap-done`, read back out of Better Stack |
| `inngest-zot-client-authz-6500.sh` | #6500 | a human `RESULT: PASS` verdict — no telemetry |

`bootstrap-done` is load-bearing: `inngest-bootstrap.sh` ships *inside* the pulled image, so "zot served the manifest" is not the deliverable — the pull half succeeding while nothing installs **is** the #7228 incident. And #6500 stays human because closing it authorizes ADR-096 Phase 5.3–5.5, which `zot-soak-6122.sh`'s blocker arm reads; a green boot marker must not authorize a supply-chain retirement.

Both harnesses assert the stub's argv, which is what stops query-construction mutations (a dropped `--grep`, a `--no-archive` that collapses a 7-day window to a ~40-minute keyhole) from staying green. Three defects they caught before commit: one malformed warehouse line aborted the whole `jq` parse and reported a clean PASS window as `channel_dark`; the banned-`${VAR:?}` check false-positived on the header paragraph documenting the ban; the anti-vacuity floor was off by two.

## Scope

Untouched: the cutover FSM, the flip guard, the monotonic latch, the Redis AOF volume, the `soleur-inngest/prd` Doppler config and its boot isolation allowlist.

### Infrastructure
- The dedicated inngest host now resolves its bootstrap image from the self-hosted zot registry first, carrying the `sha256` digest pin unchanged onto both registry legs, with GHCR retained as a (currently unauthenticated) structural fallback.
- Every registry outcome is reported off-box via the boot phone-home channel, including a per-leg marker naming why each leg failed.
- The zot pull credential is redacted from every shipped log tail.

### Tests
- New mutation battery proving the zot-primary guard can fail, registered in `infra-validation.yml`.
- Fixed four `producer | grep -q` SIGPIPE flakes and a whole-file-scoped assertion in the bootstrap guard.

### Docs
- ADR-096 amended: the inngest cold-boot pull site is migrated, by bake rather than Doppler, with the risk shift recorded for whoever authorises the GHCR retirement.


deruelle:
Heads-up before this comes out of draft: `Apply web-platform infra` is currently **red on `main`** at `19bbdb768` on the `[ack-destroy]` destroy-guard — details in #7539.

This PR is the natural vehicle for the ack. When you merge, the **merge commit message** needs a line containing exactly `[ack-destroy]`, otherwise this apply fails the same way.

What that ack authorizes (confirm it is the whole intended change first):

- create 4 inngest-consumer resources already declared on main but unapplied
- update `cloudflare_bot_management.soleur_ai` in place
- replace `terraform_data.journald_persistent`, which re-runs its create-time provisioners against **production web-1**: journald restart + flush, then re-render `vector.toml` and restart the Vector agent

No destroy-time provisioners, and the resource render-sanity-gates itself before touching the live agent. Note the replace is *wanted*: it is the only path that makes #7457's `vector.toml` change live on web-1, which today is still file-only.

PR #7597, opened 2026-08-16
fix(ci): cla-evidence calls bun after its setup was removed — required check broken on every PR

deruelle:
## The required `cla-evidence` check is broken on every PR

#7566 replaced `Setup Bun` with `Setup Node.js` in `.github/workflows/cla-evidence.yml` and left two `bun run` invocations behind. The `Record allowlist-bypass (per-quarter canonical)` step now dies with:

```
/home/runner/work/_temp/….sh: line 8: bun: command not found
##[error]Process completed with exit code 127
```

`cla-evidence` is a **required status check**, so no PR opened or synced since that merge can reach auto-merge. Eight PRs are currently open. No issue or fix existed when this was filed.

It is invisible on `main` because the workflow only runs on pull requests — every `main` run since shows `skipped`.

## The fix

In the direction #7566 intended (Node + npm, one lockfile of record) rather than by re-adding bun:

- the workflow already runs `npm ci --ignore-scripts` in `apps/web-platform`
- `tsx ^4.22.4` is already a dependency there

So both call sites now use the **pinned local binary**:

```diff
-          payload=$(bun run apps/web-platform/scripts/cla-evidence/build-record.ts) \
+          payload=$(apps/web-platform/node_modules/.bin/tsx apps/web-platform/scripts/cla-evidence/build-record.ts) \
```

Not `npx` — it resolves against its own cache and can silently major-jump, which is the failure class the repo already documents for `vitest` and `tsc`.

## Verification — measured, not assumed

The remedy is a claim, so I ran it. Before, the step never executed the script at all (exit 127). After, from repo root:

| script | result |
|---|---|
| `build-record.ts` | loads, reaches its own env validation → `::error::env COMMENT_ID missing` |
| `build-bypass.ts` | loads, reaches its own env validation → `::error::env ACTOR_LOGIN missing` |

That is the correct local behaviour with no CI environment: the script now *runs* and rejects missing inputs, instead of never starting.

Also checked: zero remaining `bun` references in the workflow, and the YAML parses.

## Scope

Deliberately its own PR. This surfaced while shipping #7573, but a discovered defect in a different subsystem stays its own change rather than riding an unrelated feature branch.

### DIFF
--- a/.github/workflows/cla-evidence.yml
+++ b/.github/workflows/cla-evidence.yml
@@ -159,7 +159,7 @@ jobs:
           PR_NUMBER: ${{ github.event.issue.number }}
         run: |
           set -euo pipefail
-          payload=$(bun run apps/web-platform/scripts/cla-evidence/build-record.ts) \
+          payload=$(apps/web-platform/node_modules/.bin/tsx apps/web-platform/scripts/cla-evidence/build-record.ts) \
             || { echo "::error::evidence-record build failed"; exit 2; }
           bash apps/cla-evidence/scripts/upload-evidence.sh "$payload"
 
@@ -184,7 +184,7 @@ jobs:
           # Do NOT swallow non-zero with `|| exit 0` — that would silently
           # erase audit records on parse failures (cq-silent-fallback-must-
           # mirror-to-sentry).
-          payload=$(bun run apps/web-platform/scripts/cla-evidence/build-bypass.ts)
+          payload=$(apps/web-platform/node_modules/.bin/tsx apps/web-platform/scripts/cla-evidence/build-bypass.ts)
           if [[ -z "$payload" ]]; then
             echo "Not an allowlist-bypass; skipping."
             exit 0
--- a/knowledge-base/legal/audits/2026-08-17-clo-ruling-cla-evidence-admin-bypass-7597.md
+++ b/knowledge-base/legal/audits/2026-08-17-clo-ruling-cla-evidence-admin-bypass-7597.md
@@ -1,0 +1,114 @@
+---
+title: "CLO ruling — admin bypass of the cla-evidence required check to land PR #7597"
+type: clo-ruling
+date: 2026-08-17
+issue: 7597
+pr: 7597
+attestation-authority: clo
+status: APPROVED (CLO-agent-ruled, Soleur-as-tenant-zero v1)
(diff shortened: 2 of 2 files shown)


deruelle:
## Condition A — DISCHARGED

The CLO ruling (`knowledge-base/legal/audits/2026-08-17-clo-ruling-cla-evidence-admin-bypass-7597.md`) attached four conditions to the admin bypass. Condition A required verifying restoration on the first real post-merge `pull_request_target` run, because the fix was unverifiable pre-merge by construction.

**Verified. `cla-evidence` = `success` on run [31982035174](https://github.com/jikig-ai/soleur/actions/runs/31982035174)** (PR #7573, the first PR to sync after this merged).

The previously-failing step now executes:

```
payload=$(apps/web-platform/node_modules/.bin/tsx apps/web-platform/scripts/cla-evidence/build-bypass.ts)
worm-duplicate-quarter status=409 key=allowlist/deruelle/2026-q3.json attempt=1 (worm-idempotent)
```

Two things this establishes:

1. **The fix works in CI.** `tsx` resolves and executes under `npm ci --ignore-scripts` — the esbuild optionalDependency concern the ruling raised does not bite. This was also measured pre-merge in a clean tree (`npm ci --ignore-scripts` rc=0, `tsx v4.22.4` executing TypeScript), so the live run is confirmation rather than first contact.
2. **The ruling's zero-loss finding is confirmed live.** The step returns the same `409 worm-duplicate-quarter` on `allowlist/deruelle/2026-q3.json` that it returned in the last good run before the break (31978969901, 23:23:31Z). The quarter's canonical record exists and is Object-Locked; every run during the dark window would have been this same no-op. Nothing was lost, and the bypass forwent a guaranteed no-op — exactly as ruled.

**Dark window: 2026-08-16T23:38Z → 2026-08-17T00:30Z (~52 min). Closed.**

### Remaining conditions

- **B** — audit record committed with the fix ✅
- **C** — `Bypass:` justification and `CLO-Ruling:` trailer on the squash commit ✅
- **D1** — break-glass runbook for a self-sealing `pull_request_target` required check → #7600
- **D2** — Art. 30 PA-7 §(g) omits the R2 layer's TOMs (pre-existing, unrelated) → #7601

All four discharged or tracked.

PR #7633, opened 2026-08-19
docs(adr): ADR-194 — propose migrating the docs site off GitHub Pages to Cloudflare Pages

deruelle:
**Status: `proposed`, not accepted.** Written to support a decision, not to record one already taken. The `ssl = "full"` mitigation (#7584) has the site healthy, so there is no time pressure — this is deliberately *not* a change made under outage conditions.

## The finding

The proxy/certificate conflict is **structural, not a bug to fix**:

- `domains.md` **mandates** `proxied = true` for this zone (HSTS preload commitment)
- GitHub Pages **refuses to issue or renew** for a proxied host — `pages/health` → `is_https_eligible: false`

Renewal needs the same eligibility as issuance, so the origin certificate expires **every 90 days by construction** and cannot self-heal.

## The decisive evidence

A full renewal window was run on 2026-08-19 under conditions verified correct **in advance** — which no previous attempt did:

| Precondition | Verified |
|---|---|
| apex+www `proxied=false` | 5/5 records |
| Public DNS on GitHub anycast | 1.1.1.1 **and** 8.8.8.8, ~20s |
| `is_https_eligible` | **`True` / `True`** in ~1 min |
| `is_pointed_to_github_pages_ip` | `True` |
| `caa_error` | `null` |
| Zone AAAA | 0 |
| ACME path | 404 GitHub-shaped, 0 redirects, HTTP + HTTPS |
| Domain toggle | performed **while eligible** |
| Pages deployment | dispatched inside the window |

**After 62 minutes: no change at all.** State flat at `bad_authz` with **no intermediate state** (`new`, `authorization_pending`) at any point, origin still serving the expired Aug 16 certificate. Certificate-transparency logs show no Let's Encrypt apex cert since mid-June, so nothing succeeded silently either.

The absence of *any* transition is the point: GitHub appears not to place an order, and no API forces one. **We hold no lever.**

## This also retires the planned day-60 renewal

Option (c) — proactive zero-downtime renewal at ~day 60 — was the intended next piece of work. Its premise was *"restore eligibility and GitHub will renew."* We held perfect eligibility for an hour, toggled the domain, and deployed. It did not.

Automating a window that has **never once been observed to work** would repeat the exact mistake this ADR exists to stop. Three successive diagnoses were recorded in this repo during remediation — short DNS window, server-side wedge needing GitHub Support, slow background re-evaluation — and **each was falsified by measurement**. That difficulty is itself an operational cost, and part of the argument.

## The real cost of migrating

The `www → apex` 301 is currently **GitHub-Pages-owned and free**. `dns.tf` is explicit that it is *"NOT a Cloudflare Redirect Rule or Page Rule."* Rebuilding it runs into a constraint this repo has already hit: one user-defined ruleset per `(zone, phase)`, and `seo_page_redirects` is at the **10-rule Free-tier cap** (a blog redirect was already dropped once).

Migrating pays for that slot: retiring the ACME carve-out frees exactly the rule the www redirect needs.

## What it does NOT solve

Stated plainly, since the motivating framing was platform reliability: Actions still builds and deploys the site, the GitHub App still drives automation, issues remain the workflow backbone. **Platform exposure is essentially unchanged.**

The honest case is deleting a subsystem — ~2,000 lines, a detector cron, a marker schema, a runbook, two open issues — whose central assumption is now measured false. Not risk reduction.

## Options considered

- **(a) Cloudflare Pages** — proposed. No origin cert exists at all; build + all six gates port unchanged, only the deploy step swaps.
- **(b) Keep `ssl = "full"` permanently** — rejected as a destination, retained as the fallback and correct interim state.
- **(c) Day-60 proactive renewal** — rejected on the evidence above.
- **(d) Self-host on `web-1`** — rejected; couples marketing availability to the app host.
- **(e) Netlify / Vercel / S3+CloudFront** — rejected; new vendor, bill, and sub-processor for no gain over Cloudflare, which already holds the zone.

## Notes

- ADR guard passes (`check-adr-ordinals.sh`); ordinal 194 is uncollided; all four required headings present.
- References ADR-125, ADR-130, ADR-136 — all verified to exist.
- The runbook link is a **forward reference to #7620** and resolves on that merge.
- `INDEX.md` regenerated via `scripts/generate-kb-index.sh` (auto-generated file).

Refs #6691, #6698, #6657, #7584, #7620


deruelle:
Implementation tracked in #7640 — scoped to be picked up cold in a fresh session.

**Gated on this ADR being accepted first.** #7640 says so explicitly and also carries the two standing constraints: `ssl = "full"` stays in place throughout (it is what keeps the site up, and is removed only *after* the migration is verified live), and the `www → apex` 301 rebuild is the real work — it is GitHub-Pages-owned today and free, and must be recreated against a full 10-rule Free-tier ruleset cap.

PR #7641, opened 2026-08-20
feat(scripts): report processes orphaned by a removed worktree, on an inode-verified deleted-cwd conjunction

### LINKED ISSUE #7537: test-contention: no reaper exists for orphaned suite processes, and the fd-based discriminator is falsified as stated
## Summary

A mutation battery ran **62 minutes** with its cwd, stdout, stderr and executing script all unlinked — its output unrecoverable by construction — while still holding a container and competing for a 16-core box. Nothing in the repo reaps this class.

`scripts/tmpfs-guard.sh` does not and should not: it is **file-scoped**, own-uid, and explicitly skips entries with open fds. That skip is correct for live runs and is precisely why an orphaned *process* survives it. The guard reaps files; nothing reaps processes.

## The hard part is the discriminator, not the kill

`setsid` — which this repo's own guidance prescribes for detaching long runs (`work/SKILL.md`, the `setsid nohup env TMPDIR=/var/tmp bash -c …` idiom) — reparents to `systemd --user`. So a **deliberately detached** run and an **orphaned** one are identical under `ps`: same PPID, same session leadership, same lack of a terminal.

The signal that distinguished them by hand was fd-based: stdout, stderr and `BASH_SOURCE` all pointing at deleted inodes.

## That signal does not hold generally — measured 2026-08-13

Probed every own-uid process on the operator box (16 cores, load 3.99/7.04/9.35, no suite running). **17 processes carry at least one deleted `cwd`/`exe`/`fd1`/`fd2` link:**

| signal | live-process hits | verdict |
| --- | --- | --- |
| `exe` deleted | ~11 `claude` processes | **useless** — a self-update unlinks the running binary |
| stdout/stderr deleted | `scripts/tmpfs-guard.sh` cron instance, 75 s old, healthy | **unsafe alone** — this is cron's unlinked output buffer |
| `cwd` deleted | **0** | the only signal no live process produced |

The decisive case, verbatim:

```
pid 704313  bash /home/jean/.../scripts/tmpfs-guard.sh   (elapsed 75s, healthy)
  cwd:    /home/jean                       NOT deleted
  exe:    /usr/bin/bash                    NOT deleted
  fd/1:   /tmp/#8874817 (deleted)          <-- cron output buffer
  fd/2:   /dev/null
  fd/255: /home/jean/.../scripts/tmpfs-guard.sh   NOT deleted
```

**A reaper keying on deleted stdout/stderr kills this repo's own tmpfs guard on its next cron tick.** The fd signal must not be adopted as stated.

The strict four-way conjunction returned **zero** hits. That is evidence of *specificity* in a sample containing no orphan — it is **not** evidence of sensitivity, and must not be read as such.

## Scope

- Detector keyed on **deleted `cwd`** plus deleted `BASH_SOURCE`/`fd 255`. Explicitly **exclude** `exe`, and never match on stdout/stderr alone.
- Reap behind that conjunction, **failing toward leaving the process alive** on any unreadable `/proc` entry, unparseable link, or ambiguity.
- Own-uid only, matching `tmpfs-guard.sh`'s existing no-sudo posture.

## Test obligations

Sensitivity cannot be validated against observed data — no orphan was present to measure. Both arms are therefore fixtures:

- **Positive (synthesized):** a process whose cwd directory and executing script are both unlinked after launch; the detector must flag it. Per `cq-test-fixtures-synthesized-only`.
- **Negative (the real false positive):** the `tmpfs-guard.sh` cron shape above — deleted stdout, live cwd and script. The detector must **not** flag it.
- **Mutation battery:** every new check mutated out individually, with the suite confirmed to redden. Per this repo's guard-work history, the danger in guard work is in the guard.

## Context

Filed from the 2026-08-13 contention incident: six concurrent full-gate suites on one 16-core box, load peaked at 43.8. This is one of four items in that class; the others are #7429 (a KILLED suite reports as FAIL), #7498 (the largest ungated suite), and a pre-launch capacity decision for `test-all.sh`.

deruelle:
## Summary

`git worktree remove` on a worktree with a suite running inside it leaves processes whose working directory and executing script have both been unlinked. They keep burning CPU and tmpfs, their output is unrecoverable by construction, and nothing in this repo reclaims them: `tmpfs-guard.sh` reclaims *files* and deliberately skips entries with open fds — which is exactly why an orphan survives it — and `proc.sh` refuses a deleted cwd by design.

This adds `scripts/orphan-process-reaper.sh`, which **reports** that class and can **reap** it only when a human asks.

Closes #7537

## What it actually keys on

A pid is an **anchor** only under a conjunction — own uid, `cwd` genuinely unlinked, `fd/255` an unlinked regular file at an absolute path, same mount namespace, same pid namespace, not the scanner or an ancestor or its process group, and older than an age floor. A confirmed anchor then authorizes a **set**: own-uid processes sharing that anchor's cwd `dev:inode` **and its session id**, each re-checked against every member gate.

**"Deleted" means `st_nlink == 0`, never the `' (deleted)'` suffix.** Measured here:

| subject | `stat -Lc '%h'` |
| --- | --- |
| genuinely deleted cwd | `0` |
| healthy process whose cwd is a real directory named `work (deleted)` | `2` |
| ordinary healthy process | `21` |

The suffix is a string the kernel appends *and* a name a directory may legitimately carry, so a suffix test produces a false positive in the direction that kills.

`exe`, `fd/1` and `fd/2` are **never consulted at all** — not "never as a positive term", never. A `claude` self-update unlinks the running binary, and a deleted stdout with a live cwd **is** this repo's own `tmpfs-guard.sh` cron instance. Keeping those reads out of the code makes both false positives structurally impossible instead of dependent on a polarity a later edit can flip.

Every unreadable `/proc` entry, unparseable link and failed `stat` leaves the process **alive** and increments a counter, so a silent drop is distinguishable from a real zero.

## The evidence this rests on, stated precisely

The conjunction returned **zero hits across the own-uid processes on the probed box**. That is evidence of **specificity in a sample containing no orphan**. It is **not** evidence of sensitivity. The only sensitivity evidence that exists is the synthesized end-to-end arm. Until the detector has been observed firing on a real orphan, the first strike is a reader's judgment — which is why `report` is the only verb anything invokes automatically.

## What review found, and why the green gates could not see it

A nine-agent review produced **30 findings, 12 of them P1**, against a change that was already green on a 148-assertion suite, a 604-assertion 36-row mutation battery, both CI shards and shellcheck. All 30 are fixed in this branch. The ones worth reading, each **measured** rather than argued:

- **The test suite reaped the live box.** The end-to-end arm ran `reap` against the real `/proc` with no signal sink, no root seam, no pid restriction and an age floor of zero — so every full-gate run performed an unattended box-wide reap, twice, and *wider* than an operator could invoke by hand. It also falsified this change's own ADR sentence, "nothing invokes `reap` automatically anywhere". Now scoped through the shipped authorization path, with an assertion that no `reap` invocation lacks a seam, a sink, or a target list.
- **The reap set was neither deduplicated nor bounded.** Every nested `bash <script>` level holds its own `fd/255`, so the motivating incident is a *tree* of anchors: 7 bash + 3 python3 in one removed worktree produced `anchors=7 set_members=70 signalled=70` for **ten distinct pids**, with the cardinality cap evaluated per-anchor so it bounded nothing.
- **Members bypassed the `fd/255` conjunction**, so a repo-root cwd inode shared by 49 processes auto-authorized **41** of them — including twelve live agent sessions. Session-id equality collapses that to ≤5.
- **The own-uid gate was unreachable in production.** It sat behind the `fd/255` pre-filter, which rejects a foreign process first (its `/proc/<pid>/fd` is mode 500), so it measured **0**. As a fork-free `[[ -O ]]` in the walk it measures **390 of 627**.
- **Two of three "fail-safe by construction" seams could suppress a refusal** — one skipped the root refusal, one disabled the anti-suicide guard. Both are now refusal-only structurally, not by description.
- **A non-numeric age floor silently disabled the age gate.** `MIN_AGE_S=10m` does not abort; the arithmetic errors and the `if` evaluates FALSE, so it signalled a **two-second-old** process while every counter reported a clean walk. No seam required. Numeric seams are now validated at the boundary.
- **The anchor pass was O(anchors × pids)** and blew its own `timeout 10` at ten anchors — the tool going silent in exactly the scenario it was written for.

The safety here rests on measurement, not on assertion. Where a comment justified a term by the class it excludes, that class was constructed and checked: a real `memfd` reads back as `/memfd:victim.sh (deleted)` — absolute, suffixed, `nlink 0`, regular — and satisfied all four terms the comment claimed excluded it (pre-fix 1 anchor, shipped 0).

### Tooling
- **New** `scripts/orphan-process-reaper.sh` — detects processes orphaned by a removed worktree via an inode-verified deleted-cwd conjunction. `report` (used automatically) and `reap` (explicit, authorization-gated).
- **New** behavioural suite (148 assertions full / 136 skip-live) and a 36-row mutation battery (604 assertions, 102.8s at `-P6`).
- `scripts/test-all.sh` — two suite registrations plus a report-only preamble probe, bounded by `timeout 10`, that prints one `ORPHAN_SCAN` line per launch.
- `plugins/soleur/scripts/lib/proc.sh` — **comment only** (0 non-comment lines added): a reciprocal note recording that its suffix test asks a *containment* question where an unresolvable path is a refusal, while the reaper asks a *liveness* question where the same reading is positive evidence.
- **New** ADR-195 and a committed procfs fixture that preflight Check 10 executes.

## Test plan

- [x] `scripts` shard `rc=0` (331/334) and `bun` shard `rc=0` (7/7)
- [x] Mutation battery 604/604 in 102.8s against a derived 240s ceiling
- [x] Behavioural suite 148 full / 136 skip-live; `guard-vacuity-floor` 19/19; `proc.test.sh` 58/58; shellcheck clean
- [x] `report` on the real `/proc`: exit 0, exactly one `ORPHAN_SCAN` line on stdout, `valid=1`, nothing signalled
- [x] Preamble on a real runner launch: one `ORPHAN_SCAN` line, zero `reap` invocations, zero signals
- [x] Committed fixture probe: `anchors=0 unreadable_gone=1`; a suffix regression moves `unreadable_gone` 1 → 0
- [x] Refuses a non-procfs root with no injected sink; refuses to run as root
- [x] `tmpfs-guard.sh` diff empty (pinned by an acceptance criterion) and its suite green

No browser surface — verified mechanically (0 `## Test Scenarios`, 0 executable `Browser:` steps, 0 files under the structural-UI gate's paths), not assumed.

## Notes

No post-merge operator steps. Nothing is scheduled, nothing is provisioned, no credential or recurring cost is introduced, and the tool reads only this machine's `/proc`.


deruelle:
**ADR-ordinal re-derivation at merge time** (tasks.md 6.3b — the item deliberately left unchecked, because it is an act that can only be performed against the tree that is about to merge).

Re-derived against freshly-fetched `origin/main` immediately before `gh pr ready`:

```
highest on origin/main = ADR-194 ; next free = ADR-195
this branch introduces  = ADR-195
scripts/check-adr-ordinals.sh -> ADR ordinal + content checks passed. (rc=0)
```

**195 is still free — no renumber, no reference sweep needed.** Re-checked a second time after the BEHIND auto-sync merged `origin/main` into the branch; that merge pulled in no files under `knowledge-base/engineering/architecture/decisions/`, and the ordinal check passes again on the synced tree.

Recorded here rather than as a commit: the spec is already archived, and pushing a checkbox-only commit while auto-merge is queued would restart the required checks for no verification value — on a box whose `--capacity` probe currently reads `CAPACITY_CONTENDED` with three sibling gates running.

PR #7646, opened 2026-08-20
fix(worktree): pipefail + `grep -q` made the registration check fail when the worktree WAS registered

deruelle:
## What

`worktree-manager.sh` runs under `set -euo pipefail` (line 29) and asked "is this line in the porcelain listing?" with a pipeline:

```bash
git worktree list --porcelain | grep -qxF "worktree $worktree_path"
```

`grep -q` exits at the **first match** and closes the pipe. If the needle is not near the end of the listing, `git` is still writing, takes **SIGPIPE**, exits 141, and `pipefail` promotes that to the pipeline's status. The check reports ABSENT precisely when the needle **is** present and early.

Measured on this repo (27 registered worktrees):

```
rc=141                    (141 = SIGPIPE)
no-pipefail EARLY: MATCH
pipefail    EARLY: NOMATCH
```

Porcelain order follows `.git/worktrees` readdir order, so which branch names break is effectively arbitrary. `create feat-one-shot-7498-vitest-project-split` landed at line 49 of 120 and failed **100% of the time**, while a probe worktree that happened to land last passed — which is why this read as flaky infrastructure rather than a deterministic bug. An instrumented run confirmed the entry *was* present at line 49 while the same `grep -qxF` still reported it missing.

## Why it matters

**1. `create` was unusable (blocking).** `verify_worktree_created` Check 2 declared a freshly created worktree "not registered", and its handler **deletes** it — after checking out 13,815 files. This blocked the `/soleur:go` → `one-shot` pipeline outright for affected names.

**2. A destructive guard was fail-open.** `heal_stale_branch`'s "a branch checked out in ANY worktree is ACTIVE, never heal it" guard falls through on a false negative to `git push origin --delete "$branch"`. So a live session's remote branch is deleted exactly when it appears early in the listing. The local prune below it is annotated *"the not-checked-out condition is already guaranteed by the early return at the top"* — that assumption was what broke. Test arm A4 demonstrates the mutant deleting `origin/wt1` while it was checked out.

## How

Capture the listing into a variable, then match in-shell — no pipeline, so no reader that can quit early. `_porcelain_has_line` returns **0 = present / 1 = absent / 2 = listing unreadable**, so callers gating a destructive path can tell "absent" from "could not tell"; `heal_stale_branch` now requires a definite `1` before deleting anything (previously an unreadable listing also fell through to the delete).

Swept the same idiom at the other three sites: the bare-repo probe (line 170) and the gone/merged membership test in `cleanup_merged_worktrees`. The heal guard also moves from `grep -qx` (a **BRE**) to exact string matching, so a branch name containing regex metacharacters can no longer false-match a sibling.

## Tests

New `plugins/soleur/test/worktree-manager-porcelain-sigpipe.test.sh` — 18 assertions, all green.

**A small fixture cannot reproduce this.** ~14 lines of porcelain fit entirely in the 64 KiB pipe buffer, so git finishes and exits 0 before the consumer closes anything, and the unfixed script **passes**. An earlier revision of this suite went green against the buggy script for exactly that reason. A stub `git` pads the listing past the buffer, making the race deterministic instead of a timing lottery.

| Arm | Asserts |
|---|---|
| A1 | `verify_worktree_created` accepts a registered, non-last worktree; worktree survives |
| A2 | **Mutant** — piped idiom restored in Check 2 must fail, with the production marker |
| A3 | `heal_stale_branch` spares a checked-out, early-listed branch; `origin/<branch>` survives |
| A4 | **Mutant** — piped idiom restored in the guard **deletes `origin/wt1` while checked out** |
| A5 | End-to-end `create` succeeds in a repo that already has several worktrees |
| A6 | Static guard: no `git worktree list … \| grep -q` pipeline remains |
| A7 | An unreadable listing does not authorise a delete (fail-closed) |

Verified end-to-end against the real 27-worktree repo: `create feat-one-shot-7498-vitest-project-split` now exits 0 through env copy, dependency install, and lease.

All 9 `worktree-manager-*` suites pass, plus `lint-orphan-test-suites.sh` (363 covered, 0 orphaned) and `git-lock-marker-telemetry.test.ts` (27/27).

## Drive-by: four suites were not isolated from the operator's git config

`worktree-manager-{heal-stale-branch,bare-sync,feature-spec-dir,bare-in-dotgit-layout}.test.sh` failed with `rc=128` on any machine with `commit.gpgsign=true`. Their fixture `git commit` fails with `failed to write commit object`, and `git clone` then only **warns** "you appear to have cloned an empty repository" — so the arms go red for a missing fixture while reading as genuine SUT failures. They already unset every `GIT_*` var, which wipes a caller-supplied override, so the exports have to come *after* that loop. Fixed inline; that misdiagnosis cost real time in this session.

`worktree-manager-atomic-config.test.sh` is deliberately **not** touched — it manipulates `GIT_CONFIG_GLOBAL` per-command for its own arms.

## Changelog

- Fixed `worktree-manager.sh` reporting a freshly created worktree as unregistered (and deleting it) when its porcelain entry was not last — a `pipefail` + `grep -q` SIGPIPE bug that blocked `create` outright for affected branch names.
- Fixed `heal_stale_branch` failing open and deleting a live session's remote branch, from the same root cause; it now also fails closed when the worktree listing cannot be read.
- Test suites are now isolated from the operator's git config so commit signing no longer produces misleading failures.


deruelle:
## Review round 1 — three findings, all fixed

A `code-review high` pass raised three findings. All are real; all fixed in `c30636c08`. Two were self-inconsistencies I introduced, both on destructive paths.

**1. MEDIUM — `_porcelain_has_line` returned a definite "absent" for an rc-0-but-EMPTY listing, and that answer authorises a delete.**

The helper only distinguished `rc != 0`. But `git worktree list` can exit 0 with no records under the masked-config wedge, and this same file already treats that as a broken registry — `cleanup_orphan_worktree_dirs` refuses to reap on it and emits `SOLEUR_ORPHAN_REGISTRY_UNAVAILABLE reason=empty-parse`, on the grounds that *any valid repo emits at least the main worktree*. My rc check watched one door while the empty parse walked through the other.

Now returns `2` on a zero-length parse. New arm **A8**; verified non-vacuous — with the guard removed, `heal_stale_branch` **deletes `origin/wt1` while wt1 is checked out in a live worktree**.

**2. MEDIUM — `verify_worktree_created` collapsed rc=2 into rc=1 and took the destructive branch.**

`if ! _porcelain_has_line …` treats "could not read the registry" identically to "not registered" — directly contradicting the docstring I had just written on the helper. A transient registry failure therefore answered *"could not tell"* with `rm -rf` on the worktree `create` had just checked out, and emitted `reason=unregistered`, asserting the wrong cause in a marker operators page on.

Now bails fail-closed with `reason=registry-unavailable phase=initial|post-repair`, leaving the tree in place. New arm **A9**; verified non-vacuous — with both `rc=2` branches removed the worktree is deleted and the marker misreports `reason=unregistered`.

*(First attempt at this mutation only removed the `initial` branch and A9 still passed, because the `post-repair` branch caught it. Re-run with both removed to confirm the arm actually bites.)*

**3. LOW — the A6 static guard was evadable by any intermediate pipe stage.**

`[^|]*` required `grep -q` to be the immediately next stage, so `git worktree list --porcelain | sed -n 's/^branch //p' | grep -qx …` would reintroduce the identical bug while the guard stayed green. Broadened to any short-circuiting reader downstream (`grep -q`, `head`, `read`).

### Other changes this round

- A2's mutant now neuters the **helper body** rather than its call sites — one anchor, and it reintroduces the bug at every use.
- Suite is **23 assertions** (was 18).
- `worktree-manager-atomic-config.test.sh` was flapping `rc=128` on this host for the same signing reason as the four suites isolated in the first commit, but it drives `GIT_CONFIG_GLOBAL` itself as the subject of its identity arms — so it needs the orthogonal `GIT_CONFIG_COUNT` mechanism instead. Stable across three consecutive runs.

The reviewer confirmed the core fix is a faithful literal whole-line replacement for `grep -qxF`, which I also verified independently across 10 cases including glob (`*`, `[e]`) and regex (`.`) metacharacters and substring near-misses.

deruelle:
## Rounds 2–3: CI found two defects in the test, both in the fixture

Recorded because the pattern repeated and the arms that caught it were the anti-vacuity ones.

**Round 2 — the stub did not die the way real git dies.** The padding was emitted by a bash `printf` loop. Real `git` takes SIGPIPE when a `grep -q` reader closes the pipe; bash's `printf` builtin does not necessarily — on the runner it printed `write error: Broken pipe` and kept looping, so the stub exited 0, the mutant could not reproduce the bug, and **A2 went red while A1 was passing vacuously**. The suite was certifying the fix on a host where it never exercised it. Pad now comes from an external `cat` via `exec`, so the stub process *is* `cat` and fails as git would. Added **A0**, which asserts the stub's own failure before any arm depends on it.

**Round 3 — A0 itself over-specified the host.** It asserted the producer exits exactly `141`; CI reported `1`. Both are right:

| SIGPIPE disposition | Where | Producer outcome |
|---|---|---|
| default | dev machines | killed by signal 13 → `141` |
| ignored, inherited | GitHub Actions runners | `write()` → `EPIPE` → `exit 1` |

Either way `pipefail` promotes it and the bug reproduces identically — which is why A2 passed on that same run. Relaxed to **non-zero**: the property under test is "the producer fails when the reader quits early", not a signal number. Verified under both dispositions before pushing this time (`( trap '' PIPE; <suite> )` reproduces the runner's locally), 24/24 each way.

### Final state

- Suite is **24 assertions**: A0 (stub faithfulness), A1–A2 (verify + mutant), A3–A4 (heal + mutant, incl. deleting a live `origin/wt1`), A5 (e2e `create`), A6 (static guard), A7 (unreadable listing), A8 (empty parse), A9 (unreadable registry under verify).
- Three times this session the same trap recurred — *the fixture did not reproduce the mechanism, so the arm passed with confidence*. All recorded in the learning file, along with the recovery pattern (`git update-ref`, not `reset --hard`) for a shared root carrying other sessions' work.

Issue #7498, opened 2026-08-12
test-all.sh: make apps/web-platform gateable by splitting the repo-wide parity guards into their own vitest project


deruelle:
deferred-automation backlog item; re-evaluate when: the vitest project split below is scheduled, or the `unit` project's repo-reading guard population changes materially.

## Why this is a sized deferral, not a wait

PR for #7494 extended ADR-181's local relevance gate to two suites and **refused** a third — the `apps/web-platform` vitest suite — on measurement rather than on preference. This issue carries the work that would make gating it sound.

The refusal rests on three findings, all re-verified independently by a `code-simplicity-reviewer` CONCUR pass:

1. **The payoff for the obvious gate is ~7%.** The honest predicate for the whole suite (`apps/web-platform`, `plugins/soleur`, `.github`, `scripts`, `AGENTS(.rules).md`, `knowledge-base/{product,overview,engineering}`) is reached by 75 of the last 80 commits on `origin/main`. The suite contains genuine repo-wide parity guards, and those guards are what pin the union rate near-constant — so the rate cannot move on its own.
2. **The lever the original issue proposed does not exist.** `vitest --shard=K/N` partitions the *collected* file set arithmetically. It cannot select by path.
3. **The cheap-looking derivation is unsound.** Deriving a "repo-wide guard" subset by grepping the test files for repo-root path literals over-matches badly: 156 test files reference `knowledge-base/`, 54 of those also call `readFileSync`, and the literals are plainly synthetic in-memory fixtures (`book.pdf`, `big.md`, `x.md`, `notes.md`) belonging to the app's own document model. A grep cannot separate those from a real read of `knowledge-base/product/roadmap.md`; measured over-match is ~50 files. Vitest's `--related` walks the *import* graph, and the parity guards reach their subjects via `readFileSync`, so it misses exactly the tests that matter.

## The counterfactual that keeps this actionable

**39 of the last 80 commits on `origin/main` (48%) touch no `apps/web-platform/` file at all.** Re-derived twice, independently, at 39/80 and 39/80.

```bash
n=0; for sha in $(git log --format=%H -80 origin/main); do
  git show --pretty=format: --name-only "$sha" | grep -q '^apps/web-platform/' || n=$((n+1)); done; echo "$n/80"
```

[...]

PR #7666, opened 2026-08-20
perf(test-all): split the repo-wide parity guards out of `unit`, then gate the app-local projects

deruelle:
## What

`test-all.sh` ran the whole `apps/web-platform` vitest as **one** suite, so it could not be relevance-gated. The `unit` project mixes app-local tests with suites whose *subject is the repository* — plugin scripts, workflow YAML, the knowledge base. Gating the lot would decline those on exactly the diffs they exist to catch.

Split into three projects:

| Project | Files | Gated? |
|---|---|---|
| `repo-wide` | 54 | **Never** — subject is the repo |
| `unit` | 817 | on `apps/web-platform/` |
| `component` | 240 | on `apps/web-platform/` |

## Measured, not assumed

**42 of the last 80** commits on `origin/main` touch no `apps/web-platform` file — **52%** (96/200 over 200 commits; re-derived twice at 42/80). That is the decision variable #7498 prescribed, and it's a deterministic `git log` replay.

**No runtime figure is claimed.** Wall-clock is not resolvable on this machine right now (load average 10, a sibling `test-all.sh` running, `/tmp` at 300 MB), and #7498's measurement note says to prefer the skip rate and state the spread rather than quote unmeasured costs.

### The open question #7498 said must not be assumed

> Whether the ~60 `server/inngest/cron-*.test.ts` files belong in the new project. This decides whether the 48% ceiling holds at all.

**Measured: 6 of 72 cron files escape the app; 66 do not.** They are overwhelmingly app-local, so the ceiling holds — and the result is better than projected: **817 of 871 unit files (94%) are gateable**, versus the ~90-file carve-out the issue anticipated.

## Why a manifest, not a glob

Because the set is not a directory. 6-of-72 *within a single family* means no path shape separates repo-wide from app-local. A manifest is #7498's own sanctioned alternative to "97 file moves plus import-depth fixups", and it avoids a large, risky rename diff.

## Why the predicate is depth-aware

#7498 warned that a literal grep over-matches (~50 files). It does — **and it under-matches, which is the dangerous direction**:

- **over-match** — `join(__dirname, "..", "..")` from `test/server/` lands *on* `apps/web-platform`. App-local despite two levels of `..`. Likewise `test/c4-render.test.ts` contains `knowledge-base/…` only as a synthetic `/workspaces/ws-1/…` constant and reads nothing.
- **under-match** — `test/git-lock-marker-telemetry.test.ts` reaches the repo via `"../../../plugins/soleur/…"`. A `readFileSync`-plus-literal heuristic misses it entirely, putting a repo-reading suite *inside* the gated project.

So a file escapes only when its longest `..` run **exceeds its depth** below `apps/web-platform`, or it names a repo-root helper. Flat `../../..` grep → 97 files. Broad grep → 105. Honest predicate → **54**.

## Containment — shipped alongside, not after

`test/repo-wide-containment.test.ts` recomputes the escape set from disk on every run and fails on drift **in either direction**, naming the files to add or remove. Without it, a new repo-reading test lands in the gated project and is silently declined on the diffs it guards — trading a slow suite for a quiet one.

It runs in the `repo-wide` project, so it is **never gated by the thing it guards**. It also asserts 0 of 240 `.test.tsx` escape, which is what makes gating `component` safe.

## Scope note — `component` is gated too

#7498 rejected gating `component` when it was evaluated as a smaller slice **instead of** this work ("imports a new fail-open surface to buy back exactly what was refused"). That objection is void once the guard exists for `unit` anyway: the same guard covers `component` at zero marginal cost, and gating `unit` while still running 240 app-local UI tests on a docs-only commit would be incoherent. Measured 0 of 240 escaping. **One-line revert if a reviewer disagrees.**

## Verification

- Partition exact: **54 + 817 = 871** files (870 tracked + the new guard), **zero overlap**; `component` unchanged at 240.
- `repo-wide`: 54 files / **985 tests green**.
- Containment guard: **6/6**, including the component arm and an anti-vacuity arm proving the predicate fires.
- `scripts/test-all-webplat-gate.test.sh`: **15/15**, with two mutants (drop the gate; gate `repo-wide`) proving both shape arms can be driven red, plus integrity arms proving each mutant is still a parseable script containing the block.
- `lint-orphan-test-suites.sh`: **365 covered, 0 orphaned** — it caught two real omissions during development (an unregistered `_diff_touches` gate, and the self-include invariants) which are now satisfied.

Note the gate is a **local-dev optimisation only**: `_diff_touches` returns "run" unconditionally when `CI` is set, so CI coverage is unchanged.

## Changelog

- `test-all.sh` now declines the app-local `apps/web-platform` vitest projects when a diff touches no file in that app — 52% of recent commits — while suites whose subject is the whole repository continue to run always.


deruelle:
## Review round — 5 findings fixed, then 2 more from CI

**`code-review high`** raised five. The medium one was the same class as the change itself: my containment guard **under-matched**, in the direction that fails open.

- **chained** — `const APP = join(__dirname,"..","..")` then `join(APP,"..","..")`. Each hop within depth; the composition reaches the repo. A longest-single-run scan reports 2 at depth 2 and says "app-local".
- **imported** — a suite with no `..` of its own still reads the repo if a helper it imports does.

Both closed, and both proven able to fire: the chained fixture reports **2 naively, 4 after substitution**, with a control asserting the single-hop shape stays app-local.

Tightening it surfaced **two false positives of my own making**, both now with regression arms:
- a multi-line binding capture credited a later statement's `../../` to `const row = emitAndParse(…)`;
- a binding named `kb` matched the string **literal** `"kb"` inside `join(__dirname, "..", "components", "kb", f)`.

Over-inclusion is the safe direction, but a guard that flags app-local files teaches people to pad the manifest.

The finding with real blast radius was **#5**: `tenant-integration.yml` runs `test/server/ --project unit`, and the split moves 9 files under `test/server/` into `repo-wide` — so that job would have silently stopped running them while its path filter still named their directory. Now passes both projects.

Also fixed: `walk()` skipped `__synthesized__` though the `unit` include does not (a hole in a guard whose job is to have none); `mutate()`'s `exit 2` ran inside `$( )` and exited only the subshell; and M2 only *deleted* the repo-wide line, leaving A3's actual failure mode — repo-wide moved *inside* the gate — uncertified.

### Then CI caught two more

Adding a fifth `_diff_touches` gate is a **multi-site registration**, and I missed two watchers:

- **`GATED`** in `test-all-infra-coverage-notice.test.sh` — a separate declaration site whose own comment records that "a fifth gate added by following the HOW-TO verbatim" once left it unregistered. Verbatim what I did.
- The **orphan-linter sandbox**, which materialises every relevance-array path from a hand-maintained list; mine was absent, so the linter failed *inside the fixture* and surfaced as M16 "a correctly-formed exclusion was REJECTED".

One judgement call there: that suite's re-run assertion required `bash …*.sh`, which no vitest suite can satisfy. Printing `bash scripts/test-all.sh webplat` would fit the shape but **re-enters the same gate on the same diff and declines again** — destroying the "actionable" property the arm exists to protect. So I widened the accepted forms and kept the half with teeth: the referenced script or directory must exist on disk (the measured failure its comment records).

### Final

68 SUCCESS / 3 skipped / 0 failures. Locally: infra-coverage-notice **127/0** (was 126/1), orphan-linter suite **67/0** (was 64/3), webplat-gate 16/16, containment guard 10/10, `repo-wide` 990 tests, `unit` 10 017 tests.

PR #7673, opened 2026-08-20
fix(git-worktree): cd succeeding is not containment — pin the fixture's git toplevel

deruelle:
## `cd` succeeding is not containment

#7546 closed the arm where the fixture directory is **absent** — `cd "$X" || … exit 90` at seven sites. That is correct and it holds. It does **not** close the arm where the directory **exists but is not a repository**: `cd` returns 0, the guard never fires, and git walks **up** to the enclosing repository.

## Measured on main, not inferred

A shim where `git worktree add` creates the target path and then fails reproduces it end-to-end against a disposable repo used as CWD, on `58486e1ae`:

```
canary HEAD 10b74df57 -> f5fd37bd6
f5fd37b v12 change
fd06a90 v9 change
5f9b610 victim2 change
afa4a26 victim change
10b74df baseline
```

The full fixture quartet, committed onto a live branch, with **every existing guard passing**. Same signature that hit three worktrees on 2026-08-20.

## The fix

`cdx()` asserts containment on the **resolved git toplevel** rather than on `cd`'s exit status: `git rev-parse --show-toplevel` must land on `$TMP_REAL` or below, else refuse and name the repository it declined to write to.

Three details that are load-bearing, each learned the hard way in this incident:

- **`pwd -P`, resolved once.** `TMPDIR` is routinely a symlink; comparing an unresolved `$TMP` against a resolved toplevel false-FAILs every call.
- **`mktemp -d` is guarded.** An empty `$TMP` resolves every fixture path against `/` — which via `TMP_REAL` would whitelist the entire filesystem.
- **A `.cdx-failed` marker reds the suite.** `exit 90` inside `( cdx … )` ends only the subshell. The marker is checked **before** the anti-vacuity floor, because a run that refused is not a run with too few assertions and must not be reported as one.

## Verified in three arms — by HEAD-unchanged, not by guard-present

Guard-present is exactly what let the previous round be declared fixed while it escaped again.

| Arm | Case | Result |
|---|---|---|
| (a) | control, unmutated, invoked normally | `rc=0`, **PASS 40 / FAIL 0** |
| (b) | dir exists, not a repo — **the new arm** | refuses, names the toplevel it would have written to; HEAD byte-identical, uncommitted work intact, no fixture commits |
| (c) | dir absent — the #7546 arm | still refuses; HEAD byte-identical |

Arm (b) is the case that wrote the quartet on main. A control run from a CWD *outside* the repo reds identically on main's version and on this one (PASS 13 / FAIL 21), so that shape is environmental, not a regression from this change.

Sibling suites green: `fanout-suite-scope`, `worktree-manager-porcelain-sigpipe`, `guard-vacuity-floor` — all `rc=0`.

## Scope and coordination

Helper shape is `soleur-fa`'s from #7616, used with their agreement rather than writing a second independent implementation. #7616 additionally carries a logical-line self-grep that flags a regression at any of the seven sites; that is **not** included here. This PR is deliberately scoped to the live hole so it can land ahead of that 12-commit branch, which is currently BEHIND with a gate mid-flight. They will carry the merge and re-apply their extras.

Refs #7546.

### DIFF
--- a/plugins/soleur/skills/git-worktree/test/lease-protects-active.test.sh
+++ b/plugins/soleur/skills/git-worktree/test/lease-protects-active.test.sh
@@ -17,8 +17,40 @@ PASS=0; FAIL=0
 fail() { echo "  FAIL: $1"; FAIL=$((FAIL+1)); }
 pass() { echo "  pass: $1"; PASS=$((PASS+1)); }
 
-TMP=$(mktemp -d)
+TMP=$(mktemp -d) || { echo "FATAL: mktemp -d failed — refusing to run with an empty \$TMP, every fixture path would resolve against /" >&2; exit 2; }
 trap 'rm -rf "$TMP"' EXIT
+
+# Resolved ONCE, with `pwd -P`: TMPDIR is routinely a symlink, and comparing an
+# unresolved $TMP against a resolved toplevel false-FAILS every call.
+TMP_REAL="$(cd "$TMP" && pwd -P)"
+
+# cd into a fixture, or refuse. Two arms, and the second is why `cd X || exit` is
+# not enough:
+#   (a) the directory is absent  -> cd fails.
+#   (b) the directory EXISTS but is not a repository -> cd SUCCEEDS, and git then
+#       walks UP to the enclosing repository. On 2026-08-20 that enclosing repo was
+#       a live worktree: `git worktree add` had created the path and then failed, so
+#       every guard keyed on cd's exit status passed and the fixture's commits landed
+#       on a real branch. Measured on main at 58486e1ae: HEAD 10b74df57 -> f5fd37bd6,
+#       carrying "victim", "victim2", "v9 change", "v12 change".
+# So containment is asserted on the RESOLVED git toplevel, not on cd's exit status.
+cdx() {
+  local target="$1" top
+  if ! cd "$target" 2>/dev/null; then
+    printf '  FAIL: fixture directory absent: %s\n' "$target" >&2
+    printf '        refusing to run fixture commands in %s — that is a LIVE repository.\n' "$PWD" >&2
+    : > "$TMP/.cdx-failed"
+    exit 90
+  fi
+  top="$(git rev-parse --show-toplevel 2>/dev/null)" || top=""
+  case "${top:-<none>}" in
+    "$TMP_REAL"|"$TMP_REAL"/*) return 0 ;;
+  esac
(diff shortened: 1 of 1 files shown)


deruelle:
Reviewed independently — not by reading the diff, but by running your branch **in place** in a throwaway worktree and driving the two failure modes. It holds.

| arm | result |
|---|---|
| control, unmutated | `PASS: 40 / FAIL: 0` |
| `worktree add` silently fails (dir absent) | refuses, names the directory, `verdicts are not evidence` |
| `worktree add` **creates the dir then fails** (dir exists, not a repo) | refuses, names the toplevel and the fixture root — **this is the arm that wrote the quartet on main** |
| your empty-root assertion, forced | `rc=2`, refuses at setup |

Live-repo canary byte-identical after every arm. Static: shellcheck clean, 7 `cdx` sites, `readonly TMP_REAL` present, zero unguarded `cd` outside the `&&`-continued forms.

**Merge-compatibility, checked so you don't have to:** I ran the logical-line residual-`cd` detector from my #7616 against your file — it passes clean. So when #7616 lands after you, its self-check will not red on your resolution.

**One methodology note on my own review, since this thread has been keeping score.** My first control run reported `FAIL: 21 / 34 assertions` and I nearly filed it. It was void: I had copied your file to `/var/tmp` and run it there, so its `BASH_SOURCE`-resolved `REPO_ROOT` pointed outside the repo. A relocated script loses whatever it resolves relative to itself — the same class as everything else here, in my review instrument. Re-run in place, it is 40/0. Your own warning about the outside-CWD red is the same phenomenon, and you were right to check it before believing it.

**On the fail-open you found in the helper I sent you:** confirmed here and fixed in #7616 as `3a6ebadd8`. Reproduced standalone — with `TMP_REAL` empty and the CWD a live repo, the guard printed `*** ACCEPTED — guard returned 0 for /var/tmp/failopen/live ***`. Worse, inside the suite the same state runs **40 passed, 0 failed**: a guard that accepts everything is indistinguishable from a healthy run. That is the vacuity class my branch exists to delete, sitting in the containment check my branch added. It had passed review here and shipped in three commits before you caught it.

Your method is the one that works and the one I had not applied to my own diff: read it asking *how does this fail open*, then **mutate the guard and watch it accept**. I have taken your `readonly` too.

**Noted that this carries a self-review, not a panel** — thank you for saying so plainly rather than letting the trailer imply coverage. Treat this comment as the second pair of eyes: I ran the arms myself and the evidence above is mine, not a restatement of yours.

Approving. Ship it — the hole is live on main and this closes it.

Issue #6813, opened 2026-07-21
fix(ship): Incident-PIR gate fires on every single-user-incident plan — its regex matches the threshold label, not an outage


deruelle:
## Summary

`/ship` Phase 5.5's Incident-PIR gate fires on essentially every plan that declares `brand_survival_threshold: single-user incident` — a large share of them — because its outage regex matches the **threshold label itself**, not an outage.

## Evidence (PR #6782, a preventive-hardening change with no incident)

The four lines that tripped `OUTAGE_RE`:

```
7:brand_survival_threshold: single-user incident        <- the threshold label
70:**If this lands broken, the user experiences:**      <- an explicit hypothetical
81:**Brand-survival threshold:** `single-user incident` <- the label again
225:  return type is load-bearing, not incidental.       <- the word "incidENTAL"
```

Meanwhile the linked issue states plainly that no incident occurred: *"Current double-fire posture is clean (dedicated host measured dark: zero runs, zero registrations, backend not wired to prod), so this is not presently firing."*

Three distinct defects in one regex:

1. **`incident` matches the threshold literal.** `brand_survival_threshold: single-user incident` is in the frontmatter of every such plan, so the gate's outage half is satisfied unconditionally for that whole class.
2. **No word boundary.** `incident` matches inside `incidental`.
3. **No hypothetical/conditional exclusion.** A plan's `## User-Brand Impact` section exists to describe what breaks *if this lands broken* — describing a hypothetical failure is that section's job, and it reads as an outage report to the regex.

The production half (`PROD_RE`) is trivially satisfied by any plan that mentions `prod` once, so the AND provides little discrimination in practice.

## Why it matters

The gate is fail-toward-PIR, so the failure mode is noise rather than a miss — but a gate that cries wolf on every `single-user incident` plan trains the operator to dismiss it, which is how it stops catching the real case it exists for. Same erosion dynamic the repo documents for muted alert channels.

## Possible fix (not a prescription)

- Strip `brand_survival_threshold:` / `Brand-survival threshold:` lines from the haystack before scanning.
- Word-boundary the outage verbs; drop bare `incident` in favour of `incident report|post-incident|outage`.
- Exclude conditional framing (`if this lands`, `would`, `could`) — a hypothetical is not an incident.
- Consider requiring a **past-tense** production-failure signal, since a PIR is owed for something that happened.

## Acceptance

[...]

PR #7753, opened 2026-09-02
fix(infra): guard the ssl=full rule holding the apex up, and fix a removal condition that could never be met

### LINKED ISSUE #7749: cert: the ssl=full rule holding the apex up is unguarded, and its removal condition can never be satisfied
The GitHub Pages origin certificate for `soleur.ai` / `www.soleur.ai` currently has **no expiry detection**. Both the producer and its monitor were turned off, and the change that was supposed to make that safe has not landed.

## Measured 2026-09-02, against `origin/main` and live infra

| Fact | Evidence |
|---|---|
| The daily cert poll is disarmed | `apps/web-platform/server/inngest/functions/cron-gh-pages-cert-state.ts` — `grep -cE '\{ *cron:'` returns **0**. Event/manual-trigger only. It previously ran 03:00 UTC daily and filed a `[cert-poll]` issue at `CERT_WARN_DAYS = 21` and `CERT_CRITICAL_DAYS = 7`. |
| Its Sentry monitor is off | `apps/web-platform/infra/sentry/cron-monitors.tf` sets `enabled = false` on `sentry_cron_monitor.scheduled_gh_pages_cert_state`, applied live 2026-08-26 (`a05ae1f77`). |
| The justifying cutover has NOT landed | `apps/web-platform/infra/cf-pages.tf` contains **zero** `cloudflare_pages_domain` resources. `dns.tf` still points the apex at GitHub Pages `185.199.108-111.153` A-records and www at `jikig-ai.github.io`, both `proxied = true`. |
| The disarmed code is live | Deployed `b5ef7b91` on 2026-08-27, which ended the #7395 build freeze and carried the code half of #7640 with it. |

## Why this matters

`cf-pages.tf`'s own header records that this certificate **"expires every 90 days BY CONSTRUCTION and cannot self-heal"** while proxied. The last failure of this class was an **8h15m apex outage on 2026-08-16**. The poll existed to give 21 days of warning; nothing provides it now.

The disarm was not wrong when written — ADR-194's premise is simply **not yet true**. The two halves of #7640 were designed to land together; the build freeze split them, and only the infra half auto-applied (`apply-sentry-infra.yml` fires on merge, independent of the image release).

## The obvious fix does not work

Setting `ssl_expiration` on `betteruptime_monitor.soleur_apex` (`apps/web-platform/infra/uptime-alerts.tf`) would **not** detect this. Because the apex is `proxied = true`, an external HTTPS probe terminates TLS at the Cloudflare edge:

```
$ echo | openssl s_client -servername soleur.ai -connect soleur.ai:443 | openssl x509 -noout -dates -issuer
notBefore=Aug 12 14:29:09 2026 GMT
notAfter=Nov 10 15:29:06 2026 GMT
issuer=C=US, O=Google Trust Services, CN=WE1
```

That is Cloudflare's edge certificate, which Cloudflare auto-renews. The GitHub Pages origin certificate is not observable from outside the proxy, so a monitor keyed on it would stay green indefinitely while the at-risk certificate expired.

Note also that the existing comment in `uptime-alerts.tf` justifying the omission of `ssl_expiration` has gone stale: it reasons that "Sentry + this monitor's regular failure would already be firing by then", and the advance-warning poll it leans on is exactly what got disarmed.

## Disposition

The correct control is an architecture decision (re-arm the poll until the cutover lands, vs. probe the origin directly, vs. something else) and has been routed to the CTO agent. This issue tracks the exposure until that control is in place.

## How this was found

Surfaced by the `user-impact-reviewer` seat during review of PR #7704, which predicted that the first post-freeze deploy would remove the last armed cert-expiry producer. That deploy has since happened, and all four conditions above were then verified directly rather than inferred.

deruelle:
## What this protects

`soleur.ai` and `www.soleur.ai` are currently served on an **expired** origin certificate. Measured from outside the proxy, identical on all four GitHub Pages anycast IPs:

```
$ echo | openssl s_client -servername soleur.ai -connect 185.199.108.153:443 \
    | openssl x509 -noout -subject -issuer -dates
subject=CN=soleur.ai
issuer=C=US, O=Let's Encrypt, CN=R13
notAfter=Aug 16 13:53:34 2026 GMT
```

That timestamp is the exact moment of the 8h15m HTTP 526 apex outage in ADR-194's own timeline. The certificate cannot renew while the records are proxied, and ADR-194 abandons rather than renews it — so it never will.

The site serves 200/301 anyway because **one config rule** overrides the zone default (Full **strict**, which validates the origin cert and refuses it) to `full`, which encrypts the CF→origin leg without validating. Removing that rule re-arms the 526 immediately, on an HSTS-preloaded apex.

Before this PR, nothing in CI protected it — and its own removal condition actively pointed at deleting it.

## The removal condition was unsatisfiable

It said: delete this block once `gh api repos/jikig-ai/soleur/pages` reports `https_certificate.state` in `{approved, issued}` with a future `expires_at`. Under ADR-194 that can never be true. A reader following it had no exit; a reader who deleted the block anyway took the apex down.

Replaced with two measurable exits — the cutover landed, **or** ADR-194 is rolled back and the cert is valid again. The first is **enforced, not remembered**: the guard resolves its stage from `dns.tf` and stops requiring the rule once the apex leaves GitHub Pages, so it self-retires at cutover rather than needing deletion.

## Falsifiability is committed, not claimed

A config-file assertion is exactly the class that rots into vacuity, so the mutation battery ships alongside: **11 rows across three axes**, plus two GREEN rows proving the guard is not simply always-red.

| Axis | Rows |
|---|---|
| SUT — the rule's fields, ordering, absence | `enabled→false`, `ssl→strict`, `ssl→flexible`, drop `www` from the host set, delete the block, insert a shadowing `ssl` rule above it |
| Fixture — the cutover stage | post-cutover `dns.tf` with the rule removed must be **accepted**; comment reflow must be **accepted** |
| Dispatch — the guard's own accounting | neuter `fail()`, neuter both helpers, delete a case |

The dispatch rows are the ones an implementation-only battery structurally cannot reach: every SUT mutation is observed *through* `pass()`/`fail()`.

**Running it found two real defects before this shipped**, neither visible by reading:

1. The shadow check compared rule indices against a mitigation selected as the *first* matching rule — so a lower index could not exist by construction and the check **could never fire**. It now asserts cardinality (exactly one `ssl` rule may target these hosts), which is non-vacuous and is the property that actually matters.
2. Neutering `fail()` alone left the suite **green**: with every case passing, `PASS+FAIL` still equalled `CASES`. Closed with a positive control that drives both helpers and requires each counter to move.

A third was caught by `shellcheck` rather than by mutation — a dead `m_idx` capture left over from the replaced check.

## Also corrected

- `uptime-alerts.tf` justified omitting `ssl_expiration` by leaning on the daily poll that ADR-194 has since disarmed. The conclusion still holds for a stronger reason: that monitor terminates TLS at the Cloudflare **edge**, so `ssl_expiration` there would watch a cert that is never at risk and stay green forever — coverage theater on the exact monitor someone would later cite as proof of coverage.
- `cron-gh-pages-cert-state.ts` claimed its `scheduleLabel` "still reads Daily 03:00 UTC". It reads `"Disarmed pending ADR-194 cutover (event-only)"`. Verified before correcting.
- ADR-194 amended to record that the pre-cutover interval is held by `ssl = "full"` rather than by cert renewal — including the accepted cost, which was nowhere in the record: origin validation is **off** for apex and www, so a MITM between Cloudflare and the Pages anycast range would go undetected. Low severity (static public docs, no auth, no credentials) but it is why the interval should not be extended indefinitely.

## Why not the other options

Ruled by the CTO agent, which falsified my own starting premise by measurement — I had assumed the origin cert was unobservable behind the proxy and at risk of expiring. It is observable, and it already expired.

- **Re-arm the daily cert poll** — the worst option. The cert is *already* expired, so it trips on its first run and every run after, filing a daily countdown issue whose remediation instruction fires a routine that de-proxies live www one-way. That reconstructs the #6691 unread-countdown pathology with the escalation path permanently hot.
- **`ssl_expiration` on the apex monitor** — watches the Cloudflare edge cert (Google Trust Services, expires Nov 10, auto-renewed). Green regardless of origin state.
- **A direct origin probe** — technically works, but would sit permanently RED beside a permanently healthy site, certifying a property that no longer predicts user impact.

## Verification

- `ssl-full-mitigation.test.sh` — 6/6 on the real tree, stage correctly resolved as pre-cutover
- `ssl-full-mitigation-mutation.test.sh` — 11/11 rows behaved as specified
- `terraform fmt -check -recursive` and `terraform validate` — clean
- `shellcheck -S warning` on both new suites — clean
- Both suites registered in `infra-validation.yml` and confirmed derived by `run-registered-suites.sh` (which builds its list from that workflow, so registration is checked rather than assumed)

Closes #7749.


deruelle:
## Incident-PIR gate — assessed, treated as a false positive

The gate fires on this PR because the body describes the 2026-08-16 apex 526 outage at length. Recording the assessment rather than dismissing it silently.

**This PR fixes no outage.** The outage was fixed on 2026-08-16 by the `ssl = "full"` rule (#7584). This PR adds the CI guard that rule never had, and replaces a removal condition that could never be satisfied. Nothing was broken by, or during, this change.

**The gate's own documentation names this class.** #6813 records it firing on the preventive-hardening PR #6782, and notes that the resulting noise "trained the operator to dismiss it". The signals require a past-tense outage verb AND a production context — both of which appear here as *history*, which is exactly the shape that over-detects.

**The referenced outage is already documented**, though not as a dedicated post-mortem:
- `knowledge-base/engineering/operations/runbooks/gh-pages-cert-renewal.md:300` — timeline entry ("Cert expired 13:53 UTC; apex served 526 for ~8h15m")
- `ADR-194` — the same event in its decision timeline, now extended by this PR's amendment

A dedicated `-postmortem.md` for 2026-08-16 does not exist. That is a pre-existing documentation gap, not one this PR introduces, and it was surfaced to the operator rather than absorbed.

Assessed and overridden deliberately, per the operator's decision.

Issue #7695, opened 2026-08-25
infra(inngest): build the gated inngest-volume-recut apply_target — nothing clears a standing /mnt/data flush latch today


deruelle:
Deferred build of the `inngest-volume-recut` `apply_target`. The **design is complete** below and in the #7674 plan (`knowledge-base/project/plans/2026-08-25-fix-inngest-host-not-serving-and-latch-gate-blindness-plan.md`, §Phase 3 + §Guard Contract); this issue carries the build.

## Why it was designed but not built (operator decision, 2026-08-25)

1. **H5/H6 are UNKNOWN.** Nobody has established that a FLUSHALL ever happened on this volume. Building the most destructive target in the inngest surface to clear a latch whose existence is undetermined inverts the evidence ordering the rest of that work insists on.
2. **It would ship inert.** Its only consumer is the cutover window, which #7674 defers.
3. **Its guards could not be graded.** The fixtures are necessarily synthesized, so the first real exercise of the guard would also be the first destructive one. Building it alongside the live plan output makes it gradeable.

## Why this capability has to exist at all

There is **no in-repo mechanism that clears a standing flush latch today**, and the runbook claimed there was until 2026-08-25:

- An `inngest-host-replace` **re-attaches** the same `hcloud_volume`; `/mnt/data` and the latch survive. Measured: volume `106261946` created 2026-07-07, attached to a host created 2026-08-20 — six weeks and one replace later, latch intact.
- `op=verify-wiped-volume` wipes `/var/lib/inngest`, **not** `/mnt/data`.

So while a latch stands, `op=arm` is refused on-host into terminal `aborted` and there is no supported way forward.

## Naming — `-volume-`, not `-luks-`

`hcloud_volume.inngest_redis` is `format = "ext4"` with **zero** LUKS apparatus (`inngest-host.tf`). Modelling this on `workspaces-luks-recut` and inheriting its name would assert an encryption posture the volume does not have. The plaintext posture itself is tracked separately by #6894.

## Guard layers (five — one more than the template)

The four copied from `workspaces-luks-recut` all check **what** is destroyed. None checks **whether the host is currently serving**, which is the failure mode that matters here: recutting a volume under a LIVE scheduler destroys in-flight `step.sleep` state and queued jobs.

1. `environment: inngest-cutover` — required-reviewer gate. Verified non-empty (`required_reviewers=[deruelle]`); a zero-reviewer environment auto-approves (DP-11 F8).
[...]

Issue #7674, opened 2026-08-20
infra: the dedicated inngest host DID serve, then the flip FSM's rollback arm stopped it — steps 2-6 of the restoration (successor)


deruelle:
Successor to the dedicated-inngest-host restoration, which was auto-closed 2026-08-20T18:30:56Z by the follow-through sweeper with **all six sequence steps unchecked**. The remaining work is currently untracked. This issue carries steps 2-6.

## Why a successor and not a reopen

The probe hosted on the closed tracker (`scripts/followthroughs/inngest-zot-boot-7462.sh`) verifies the **zot-primary boot + `stage=bootstrap-done`** and says so in its own header. It returned PASS, which is correct — the host replace fired 2026-08-20T00:52:13Z and the image did bootstrap. The predecessor's body, however, was a six-step restoration. Probe scope != issue scope, and that mismatch is what closed it.

Reopening would re-arm the same probe, which would PASS again and re-close. The probe's contract is **satisfied and should be retired**; this issue takes a probe that measures the thing that is actually outstanding — that the host *serves*.

Note the probe's own header: "sweep-followthroughs.sh lists `--state open`, so hosting a probe on a closed issue is a permanent silent no-op". That is the current state.

## Measured state (2026-08-20 ~21:00Z — re-measure before acting)

| Fact | Value | How measured |
|---|---|---|
| `INNGEST_CUTOVER_FLIP` | `rolled-back` | `doppler secrets get INNGEST_CUTOVER_FLIP -p soleur-inngest -c prd --plain` |
| `INNGEST_DIAGNOSTIC_BOOT` | `1` | same |
| Host | id `162809678`, cpx22, created 2026-08-20T00:52:13Z, running | Hetzner API |
| Serving | `server_active=inactive`, `http_code=000`, `bind=[]`, `priv8288=000` | Better Stack `SOLEUR_INNGEST_SERVER_PROBE` |
| Duration | **12 consecutive hourly probes** 08:55Z-19:57Z, identical, same `boot_id` | same |
| web-1 scheduler | healthy, `inngest/scheduled.timer` + `function.finished` flowing | Better Stack |

The step-3 `net-health` measurement **has already been taken**. It says the work is step 4.

## Two corrections to the prior read

[...]

PR #7754, opened 2026-09-02
fix(inngest): the probe could not tell which host it was on, and four fields could still render 0

deruelle:
## Summary

Merge A of #7695: **the read channel only.** `SOLEUR_INNGEST_SERVER_PROBE` now carries the store facts a future Guard 2 will read before clearing a destructive volume recut. There is deliberately **no `.tf`, no cloud-init, no LUKS and no gated apply_target** here — the plan mandates the two-merge split in its own words ("Merge A's delivery vehicle aborts on Merge B's `.tf` edits, so combining them makes Merge A undeliverable"), and Merge B is deferred by operator decision.

The degradation direction is inverted against every other probe in this repo. Guard 2 clears the destroy on a **measured-empty** store, so `0` is the *clearance condition* — a field that degrades to `0` authorizes the destroy it exists to withhold. That asymmetry is why the two-token vocabulary exists: `n/a` (does not apply on this host) and `__UNREADABLE__` (applied, unanswerable), never `0`.

## What review changed

A 6-agent panel plus a mutation audit found four P1s and a battery that could not see them. All fixed inline.

- **The discriminator did not discriminate.** The store fields were gated on *"is `/mnt/data` a mountpoint"*, with the comment asserting the co-located web host has no `/mnt/data`. `cloud-init.yml` mounts the **workspaces** volume there. On web-1 the test was true, so the probe would have walked every user's repository tree hourly, shipped the aggregate byte count off-box, and put a store measurement from the wrong host into the clearance condition. The role test is now `DOPPLER_PROJECT == soleur-inngest`, and `host_role` is emitted so the consumer can require a self-identifying row.
- **`redis_keys` is not delivered.** It could never authenticate in production (`inngest-redis.service` runs with `--requirepass`; the probe unit carries no credential), so it would have read `__UNREADABLE__` forever while every test drove the fixture seam. The fix is unavailable to this merge: supplying the credential means `doppler run`, but `/usr/bin/doppler` is a symlink created only by `cloud-init-inngest.yml` — the *shared* unit would fail `203/EXEC` on the web host and silently end the hourly liveness marker on the host that actually serves. **Merge B must decide Guard 2 condition 5 explicitly**: drop it and rest the emptiness claim on `data_bytes` + `data_mount_src` + `host_role`, or solve the credential and re-add the field at `probe_schema=3`. It must not read a missing `redis_keys` as satisfied — the plan's own case 12 says absence must not parse to `0`. Reasoning is in the plan's `## Addendum — 2026-09-02`.
- **`latch_flushed_at` and `latch_lines` are not delivered** — no consumer, and `latch_lines` degraded to `0` on the not-latched path, 41 lines below the comment saying neither token is ever `0`.
- **`inngest-cutover-flip.sh` is byte-identical to `main`.** Its mirrored `flush_latched` bit was computed by a *different* predicate than the FSM's own `flush_already_performed` (`-f` not `-e`, no mount gate, missing the documented "completed before this change shipped" arm), so a host in terminal `done` would have emitted `{"flag":"done","flush_latched":false}` every 30 seconds — a self-contradictory row from the source of truth.
- **`flush_latched` inherits `data_bytes`' unreadability.** `[ -f ]` cannot report failure — it returns false for "absent" and for "cannot read the directory" alike. `findmnt` reads `/proc/self/mountinfo`, not the device, so a Hetzner volume detached while still mounted leaves the mount entry standing and every read gives EIO; a bare test would have emitted `false`, a positive claim about a store never read. `du` *does* report failure, so the silent field is keyed off the loud one.
- **Every pre-emit call is bounded** (`timeout 5 findmnt`, `timeout 15 du`), with `TimeoutStartSec` raised to 120 and the budget written out — a bound that fires degrades one field, a unit timeout loses the whole row and blocks `OnUnitActiveSec` from re-firing.

## The tests were the likeliest defect site, and were

16 of 24 mutants survived the previous 250-assertion battery. Every one is now red:

| Mutation | Before | After |
|---|---|---|
| `data_bytes` `__UNREADABLE__` → `0` | 250/250 green | 2 failures |
| `du` normaliser degrades to `0` | green | 3 failures |
| literal-valued field appended to the `logger` site only | green | 1 failure |
| field names bound to swapped variables on the phone-home site | green | 1 failure |
| `data_bytes` hardcoded, `du` never called | green | 3 failures |
| discriminator reverted to the mountpoint test | n/a | 11 failures |
| `flush_latched` tested independently of `du` | n/a | 1 failure |
| `assert()` → `if true` | 250/250 green | FATAL, exit 1 |
| delete the entire `#7695` block | `232/232 passed, 0 failed`, exit 0 | floor fires, exit 1 |
| latch suite: `fail()` stops incrementing | `45 passed, 0 failed`, exit 0 | FATAL, exit 1 |

Five host-shape arms now drive the probe (web host with the mount, a byte count **and** a latch all present so only `DOPPLER_PROJECT` differs; dedicated latched; dedicated unlatched; mounted-but-EIO; no mount). The never-zero claim is asserted as **negative space** over every arm × every field rather than as positive token expectations. Site parity compares the whole payload byte-for-byte instead of field names. Both suites carry an instrument self-test upstream of every assertion, and both report their assertion floors with `printf` + `exit 1` rather than through the helper they backstop (ADR-193).

### Web Platform infra

- `SOLEUR_INNGEST_SERVER_PROBE` gains `probe_schema=2`, `host_role`, `flush_latched`, `data_mount_src`, `data_bytes` at both emit sites, scoped to the dedicated host by `DOPPLER_PROJECT`.
- The probe unit reads `/etc/default/inngest-server` for the role discriminator; `TimeoutStartSec` raised to 120 to cover the bounded-call budget.
- `inngest.test.sh` 250 → 289 assertions, floor pinned at the exact dispatched count.
- `inngest-cutover-latch.test.sh` gains an instrument self-test; its floor no longer dispatches through `fail()`.

### Knowledge base

- Plan addendum recording the four delivery deviations and what Merge B must decide.
- Compound learning: a discriminator built from an absence, and a fixture that agreed with it.

## Test plan

- [x] `apps/web-platform/infra/inngest.test.sh` — 289/289
- [x] `apps/web-platform/infra/inngest-cutover-latch.test.sh` — 45/45
- [x] `apps/web-platform/infra/cutover-inngest-workflow.test.sh` — 497/0
- [x] `apps/web-platform/infra/inngest-dedicated-host-classify.test.sh` — 48/0
- [x] `apps/web-platform/infra/journald-config.test.sh` — 81/0 (Vector Source 4 allowlist unchanged; the logger tag did not change)
- [x] Mutation re-run: all 10 previously-surviving mutants red
- [x] `TEST_GROUP=all bash scripts/test-all.sh`

## Correction on the record

Four commits, the compound learning, the review trailer and issue #7761 all stated that `inngest-cutover-flip.sh` was byte-identical to `main` before it actually was. The revert used `git checkout -- <path>`, which restores from the index — and the change was already committed, so it restored the very thing it was meant to remove. `git diff --stat origin/main...HEAD` at ship time is what contradicted the claim. It is genuinely reverted now, verified by the diff printing nothing rather than by the command having been run.

## User-Brand Impact

Plan: `knowledge-base/project/plans/2026-09-02-infra-inngest-volume-recut-luks-plan.md` (see its `## User-Brand Impact`, `## Observability`, and the `## Addendum — 2026-09-02` recording what Merge A actually delivers).

**If this lands broken, the user experiences:** nothing directly, and that is exactly why it is worth stating carefully. This merge ships no runtime behaviour on any serving path — it changes an hourly observability probe. But `inngest-bootstrap.sh` is the **shared** renderer for the dedicated inngest host *and* the co-located web host that is currently the fleet's only live scheduler. A probe that stalls before its unconditional emit holds its unit `activating`, which blocks `OnUnitActiveSec` from re-firing and silently ends the hourly liveness marker on the host that actually serves — the one signal that would tell us scheduling had stopped. That is why every pre-emit call is bounded and the unit timeout was re-budgeted rather than a bound trimmed. Downstream, the fields this merge emits are the inputs to a gate that will authorize a `luksFormat` of the scheduler's Redis AOF volume; a field that renders `0` when it means "I could not read this" would clear that destroy on a store holding queued jobs and armed reminders.

**If this leaks, the user's data is exposed via:** the byte count of a volume, shipped hourly to Better Stack. On the dedicated host that measures the inngest Redis store and is the intended signal. On the co-located web host `/mnt/data` is the **workspaces** volume — every user's repository tree — and the earlier revision of this change would have walked it hourly with an unbounded `du` and shipped the aggregate off-box. That is the leak this merge exists to not ship; the web-host arm of the test suite is the assertion that it cannot.

- **Brand-survival threshold:** `single-user incident` — a single user losing queued jobs and armed reminders to a wrongly-cleared volume destroy is unrecoverable for that user, and the AOF holds in-flight prompts and agent output (ADR-142, the encryption-posture ledger's highest-sensitivity row).

## Net-issue-flow

- **#7761** — a discovered defect in a *different* subsystem that must stay its own issue and its own PR. `inngest-cutover-flip.service` runs `doppler run` as root with no `User=`, and its fixture seams (`CUTOVER_REDIS_CLI_CMD` and siblings) read straight from the injected environment, so a Doppler secret *name* is an arbitrary-command-as-root path on the dedicated host at the next 30s poll. It is pre-existing on `main` — `inngest-cutover-flip.sh` is byte-identical in this PR — and the fix touches the cutover FSM, which this merge deliberately does not. Fixing it inline would violate the scope discipline that keeps Merge A deliverable.

Tracks #7695
Tracks #7674
Ref #7761


deruelle:
**Correction to this PR's body and commit message.** One of the three reasons given for dropping `redis_keys` is false, and it was the load-bearing one.

The body says supplying the probe's Redis credential "means `doppler run`, but `/usr/bin/doppler` is a symlink created only by `cloud-init-inngest.yml`, so the *shared* unit would fail `203/EXEC` on the web host". **The probe unit's `ExecStart` never needed to change.** As rendered by `inngest-bootstrap.sh` it already carries:

```
EnvironmentFile=-/etc/default/inngest-doppler
EnvironmentFile=-/etc/default/inngest-server
ExecStart=/usr/local/bin/inngest-server-probe.sh
```

A third `EnvironmentFile=-/etc/default/inngest-probe` carrying only `INNGEST_REDIS_PASSWORD` is shape-identical to what is already there. No `ExecStart` change, no `/usr/bin/doppler` dependency, no `203/EXEC` — the web host simply has no such file, the `-` prefix tolerates it, and the variable stays unset, which is honest for a host with no inngest Redis. `web-probe-envwrite.sh` had already established the single-secret-env-file pattern (`/etc/default/inngest-consumer-probe`, `umask 0137` + `chmod 600`, consumed by `EnvironmentFile=` in `inngest-consumer-probe.service`).

I reasoned from "the unit needs a secret" to "the unit must run under `doppler run`" without reading the unit I had edited in the same PR, and the claim then propagated to four sites before anything tested it.

**What this changes:** nothing that merged here is wrong — the discriminator fix, the never-zero assertions, the bounded calls and the flip revert all stand. What changes is the *scope conclusion*. Merge B re-adds `redis_keys` at `probe_schema=3` behind that env file, and Guard 2 rests its emptiness claim on `redis_keys == 0` conjoined with `data_mount_src` pinning rather than on a human reading a byte count.

Corrections are appended (not edited) to the plan as `## Addendum — 2026-09-03` and to the compound learning. Tracked in #7695.

Issue #7650, opened 2026-08-20
sentry: the 0.15.4 provider still reads the deprecated alert-rule path — a required gate wedges on Sentry's brownout schedule


deruelle:
## The claim that closed #6636 is measured false

#6636 was closed by bumping `jianyuan/sentry` `0.15.0-beta2` → `0.15.4`, on the rationale that
provider **v0.15.3** (`jianyuan/terraform-provider-sentry#885`, "fix: Update reads from GET
endpoint") moved `sentry_issue_alert` reads **off** the deprecated
`projects/{org}/{proj}/rules/{id}/` path, so the root "no longer depends on the retired read path".

That datum was **changelog-sourced and never plan-measured** — and could not be at the time. The
2026-07-17 Phase 0 plan ran *outside* a brownout window, so a clean plan was consistent with both
"the bump fixed it" and "we probed between windows". The Phase 0 evidence file says so explicitly:
"with the endpoint restored, `terraform plan` cannot observe the read-path rework, so it is
changelog-sourced, not plan-measured."

## Measured, in CI, with `v0.15.4` installed

| Run | Time (UTC) | Result |
|---|---|---|
| `32362320701` | 2026-08-20 11:08:09 | **pass** |
| `32362401543` | 2026-08-20 11:09:07 | **fail** — `410 "This API no longer exists"` on **29 of 29** `sentry_issue_alert` reads, `terraform plan` exit 1 |

Same branch, same pin, **one minute apart**. `- Installing jianyuan/sentry v0.15.4...` appears in
the failing run's own init log. The alternation repeats: 2026-08-19 17:43 pass / 18:26 fail, and
21:21 pass / 21:30 fail.

So `0.15.4` **still reads the deprecated alert-rule path**, and Sentry serves that family under
scheduled brownouts (deprecated 2026-05-14; see ADR-031 §Amendment 2026-08-19).

## Why this matters beyond the docs

`sentry-destroy-required` is a **required** status check (confirmed against the live ruleset), and
`apply-sentry-infra.yml` runs the same full-root plan. Both therefore fail **on the vendor's
schedule**, not on anything about the change under review — merge control over this repo is
partially held by Sentry's brownout calendar. This is the "detector blocks its own cure" deadlock
ADR-031 describes, now load-bearing rather than hypothetical: the gate that must pass to deploy a
fix is itself wedged by the thing being fixed.

## What is NOT the fix

[...]

Issue #7634, opened 2026-08-19
Sentry alert-rule WRITE path has no known shape on the workflows/ replacement (2 scripts blocked)


deruelle:
## Summary

Two scripts still read **and write** Sentry's deprecated `projects/{org}/{proj}/rules/` endpoint. Unlike the three siblings migrated so far, they cannot be fixed by swapping the URL: the replacement's payload has no `conditions`, `filters` or `actions` keys, so every jq path and every write body in them is structurally dead against it.

Discovered while fixing #7590. Deferred deliberately, with CONCUR co-sign — see "Why this is not inlined" below.

## The two scripts

| Script | Deprecated calls | Automated caller |
|---|---|---|
| `apps/web-platform/scripts/audit-sentry-extra-text-references.sh` | `:278` GET, `:425` GET, `:464` **PUT** | none |
| `apps/web-platform/scripts/configure-sentry-alerts.sh` | `:89` GET, `:123` **PUT**, `:136` **POST** | none |

Neither is wired into a workflow (`grep` over `.github/`, `scripts/`, `package.json`, `apps/` finds no invocation). Both are operator-run. So **no brownout can abort a workflow through these** — that is why this is not urgent. It is *not* because a manual run "fails loudly": a brownout fails *intermittently*, which is the exact misreading that cost this repo months on the audit gate (ADR-031 §Amendment 2026-08-19).

The already-migrated siblings, for contrast: `sentry-monitors-audit.sh` (#7590) and `assert-byok-rules-exist.sh` (same PR — that one WAS wired into `apply-sentry-infra.yml` as a fail-closed post-apply step, so it got fixed inline).

## Why it is not a URL swap

Measured live against the prod org on 2026-08-19 (`SENTRY_IAC_AUTH_TOKEN`, Doppler `prd`):

```
GET /api/0/projects/{org}/{proj}/rules/   -> 410
  x-sentry-deprecation-date: 2026-05-14T00:00:00+00:00
  x-sentry-replacement-endpoint: /api/0/organizations/{org}/workflows/

GET /api/0/organizations/{org}/workflows/?per_page=100 -> 200, array of 30, no deprecation headers
  keys: actionFilters, config, createdBy, dateCreated, dateUpdated, detectorIds,
        enabled, environment, id, lastTriggered, name, organizationId, owner, triggers
```

No `conditions`. No `filters`. No `actions`. The mapping, from a live `auth-callback-no-code-burst`:

[...]

Issue #4781, opened 2026-06-02
Sentry auth alert rules drift to empty filters → email on every issue (recurrence guard)


deruelle:
## Problem

The 4 auth issue-alert rules in live Sentry (`auth-callback-no-code-burst`, `auth-exchange-code-burst`, `auth-signout-burst`, `auth-per-user-loop`) had **drifted to empty `conditions: []` and `filters: []`**. A Sentry issue alert with `actionMatch=all`/`filterMatch=all` and *zero* filters matches **every** issue in the project, so all four were emailing IssueOwners/ActiveMembers on unrelated events.

This surfaced on 2026-06-02 when a `feature=feature-flags`, `op=flagsmith.getIdentityFlags` WARNING on `/login` (Sentry event `8563c7e88cc240c1a44d1427d4fdf33e`, the expected #4571 warn-mirror) generated an email **"triggered by auth-callback-no-code-burst"** — a rule whose intended filter is `feature=auth AND op=callback_no_code`. The repo's existing learnings repeatedly classified this "triggered by auth-callback-no-code-burst" pattern as a *coincidental red herring* (see the 5 learnings under `knowledge-base/project/learnings/` referencing it); in fact the rule literally fires on everything because its filters were blanked.

### Root cause

`apps/web-platform/infra/sentry/issue-alerts.tf` creates these 4 rules with `conditions_v2 = []` / `filters_v2 = []` **placeholders** and `lifecycle.ignore_changes = [conditions_v2, filters_v2, ...]`. The design intent (ADR-031, "import-only") is that the real filters are managed out-of-band (the legacy `apps/web-platform/scripts/configure-sentry-alerts.sh` upsert, run once, then imported). But the rules ended up **created by terraform/CI** (`createdBy: web-platform-ci-…@proxy-user.sentry.io`, `dateCreated: 2026-05-17`) with the empty placeholders, and the intended filters from `configure-sentry-alerts.sh` were never (re)applied. `ignore_changes` then froze them empty, and nothing in the audit pipeline asserts non-empty filters.

### Immediate remediation (already done — live only, NOT in git)

All 4 rules were restored via the Sentry API to their `configure-sentry-alerts.sh` definitions:

| Rule | conditions | filters |
| --- | --- | --- |
| auth-exchange-code-burst | EventFrequency ≥5 / 15m | feature=auth AND op=exchangeCodeForSession |
[...]

Issue #943, opened 2026-03-20
sec: tc_accepted metadata in raw_user_meta_data is client-writable


deruelle:
## Summary

The `tc_accepted` flag used to record T&C consent is stored in `auth.users.raw_user_meta_data`, which is client-writable by design in Supabase. A malicious user could call the Supabase auth API directly, passing `tc_accepted: 'true'` without interacting with the consent checkbox.

## Impact

GDPR Article 7(1) requires the controller to demonstrate consent was given. If the evidence of consent is client-forgeable, the audit trail is weakened.

## Current State

- Signup page passes `tc_accepted` as `options.data` in `signInWithOtp` (client-side)
- Trigger in migration 005 reads `raw_user_meta_data->>'tc_accepted'` as ground truth
- Migration 007 uses the same metadata as discriminator for remediation

## Recommended Fix

Record consent acceptance server-side. After `exchangeCodeForSession` succeeds in the callback route, set `tc_accepted_at` via the service client based on a server-verified condition (e.g., signed token or server-side form submission record).

## Context

Found during security review of #934.

## Priority

P3 — Pre-existing architectural concern, not a regression.

deruelle:
All acceptance criteria from this issue are already addressed on main by the following PRs:

- **#938** (`fix(security): move T&C acceptance to server-side action`) — core fix: signup page no longer passes `tc_accepted` in metadata; new `/api/accept-terms` server action records consent via service client
- **#940** (`fix(sec): enforce tc_accepted_at — middleware, WebSocket, and acceptance page`) — enforcement layer: middleware redirects users without `tc_accepted_at` to `/accept-terms`
- **#941** (`fix(data): remediate fabricated tc_accepted_at timestamps`) — data cleanup for any rows created via the old client-metadata path

Verified all 11 acceptance criteria are met:
- [x] `signInWithOtp` no longer passes `tc_accepted` in `options.data`
- [x] Consent recorded server-side via `POST /api/accept-terms` (service role client)
- [x] Migration 007 removes `raw_user_meta_data` trust from `handle_new_user()` trigger
- [x] `ensureWorkspaceProvisioned` no longer accepts `tcAccepted` parameter
- [x] `ignoreDuplicates: true` trap resolved
- [x] `User` interface includes `tc_accepted_at: string | null`
- [x] `supabase.auth.updateUser({ data: { tc_accepted: true } })` has no effect
- [x] Existing users' timestamps preserved
- [x] Login flow unaffected
[...]

PR #7820, opened 2026-09-03
docs(7650): resume prompt for Phase 2 state surgery

deruelle:
Hands the next session what it needs to run the Phase 2 state surgery, written to be verified rather than trusted.

## The set is 24, not 27 and not 29

Three counts are all defensible-looking here and I have already published two of them, so the prompt carries the **regenerating jq** rather than the list:

```
30 live workflows
-1 Sentry's own default (566201) — never import, not ours to manage
-2 auth-per-user-loop (566671), sandbox-startup-failure (669246)
   event_unique_user_frequency_count, still unexpressible as a trigger at 0.15.7 — upstream #950
-3 auth-signout-burst (566672), auth-exchange-code-burst (566682), auth-callback-no-code-burst (566683)
   provider CAN express these, but their .tf blocks are empty shells
 = 24
```

## The plan is now stale in one place, deliberately flagged

Phase 2.1 says frequency conditions are "NOT MIGRATABLE at v0.15.5". True when written; #943 has since made `event_frequency_count` expressible as a trigger at 0.15.7. The line was version-scoped so it is not *wrong*, but following it would now under-migrate. The prompt says so rather than silently contradicting the plan.

## The auth four are a decision, not a blocker

Three are expressible at 0.15.7, and their real definitions are now readable from the API (per Phase 0). Authoring them properly would retire `configure-sentry-alerts.sh`'s ownership and address #4781's drift-to-empty recurrence — but that changes what owns live paging config, so the prompt routes it to the operator instead of folding it into "move 24 rules".

## Carried forward

The sequencing hazard (surgery **before** the merge that lands the blocks, or 24 duplicate live rules), `state rm` being refresh-free and therefore brownout-safe, the two-Doppler-config credential split, and the three traps this issue has already sprung — changelog-over-source, the boolean-comparison fallback that still loses a threshold at 0.15.7, and `ignore_changes` marking two disjoint sets in one file.

Docs-only.

Refs #7650. Refs #950. Refs #7634. Refs #4781.


deruelle:
Closing as superseded — the Phase 2 surgery this prompt was written to hand off has already run and verified in production.

What actually shipped (Refs #7821, merged 2026-09-06): 27 rules adopted as `sentry_alert` — the 24 this prompt scoped PLUS the 3 expressible auth-* rules it flagged as an operator decision, which was approved and folded in. The post-merge apply ran `Apply complete! Resources: 27 imported, 0 added, 0 changed, 0 destroyed`; its only red was a post-apply AC17 catch on a `sentry_issue_alert.kb_db_error` state orphan, reconciled by 2026-09-08 (auto-filed tracker Refs #7865, closed). Refs #7988 then migrated `git_data_boot_warning` and enrolled the two that cannot move.

Latest apply (2026-09-15): `0 added, 0 changed, 0 destroyed` with `AC17: state holds 29 sentry_alert and 2 sentry_issue_alert; the .tf declares 29 and 2`.

Two notes for the record, since this doc's numbers predate the merge:

- The set that shipped was **27**, not this prompt's 24 — the difference is the auth-* three, migrated with their API-read live definitions (which also retired the drift vector where `configure-sentry-alerts.sh` wrote cadence `60` over live `61`/`62`).
- Its Phase 1 prerequisite check greps for `sentry_alert_blocks_count`, which never existed — the merged identifier is `sentry_alert_shrink`. Anyone regenerating checks from this file should read the filter, not the literal string.

The residual (2 rules pinned by `event_unique_user_frequency_count`, provider fix merged upstream but unreleased) is tracked by Refs #7985. Refs #7650 is closed.

Issue #6286, opened 2026-07-09
chore(infra): bump cloud-init inngest-bootstrap pin v1.1.18 → v1.1.19 (infra-validation AC6 drift)


deruelle:
## Pre-existing infra-validation failure: inngest-bootstrap pin lags published tag

`apps/web-platform/infra/cloud-init.yml` pins `soleur-inngest-bootstrap:v1.1.18`, but
`vinngest-v1.1.19` has been published as a git tag. The AC6 drift-guard in
`apps/web-platform/infra/cloud-init-inngest-bootstrap.test.sh` (run by
`.github/workflows/infra-validation.yml`) asserts the pin equals the semver-max published
`vinngest-v*` tag, so it now FAILS:

```
FAIL: cloud-init pin (v1.1.18) matches latest published vinngest-v* tag (v1.1.19)
DRIFT: cloud-init.yml pins v1.1.18 but the latest published tag is v1.1.19.
Fix: bump every 'soleur-inngest-bootstrap:<tag>' ref in
apps/web-platform/infra/cloud-init.yml to v1.1.19.
```

**Impact:** fresh web-host boots install the older `v1.1.18` inngest-bootstrap image. `infra-validation`
is NOT a required merge check, so this does not block PRs — but it leaves the infra-validation workflow
red on any PR that touches `apps/web-platform/infra/`.

**Provenance:** pre-existing on `origin/main` (pins `v1.1.18`); surfaced during PR #6281 (#6278 zot
mirror-staleness alarm), which touches `cloud-init.yml` but does NOT own the inngest pin. Bumping a
version pin in an unrelated feature PR would violate `wg-never-bump-version-files-in-feature`, so it is
tracked here as a standalone infra chore.

**Fix:** bump both `soleur-inngest-bootstrap:v1.1.18` refs (IREF + ZIREF, `cloud-init.yml:~647/649`) to
`v1.1.19` in a dedicated infra PR, and re-run `cloud-init-inngest-bootstrap.test.sh` (AC6 + AC6b green).
Verify `v1.1.19` was actually built/published (not a stray tag) before bumping.

deruelle:
## Still drifting — now v1.1.21 → v1.1.22 (surfaced again by PR #6528)

This recurred with a newer version pair. Not filing a duplicate; updating here.

[...]

PR #7866, opened 2026-09-06
fix(sentry-infra): AC17 derives its expected counts — a hardcoded 2 cried wolf on a healthy apply

deruelle:
## Summary

The #7650 Phase 2 adoption merge applied cleanly — 27 imports landed, `Terraform apply` succeeded, BYOK Art. 33 liveness passed. **AC17 then failed, and it was wrong.**

```
AC17: state holds 27 sentry_alert and 3 sentry_issue_alert address(es)
AC17 FAILED: expected 27 sentry_alert and 2 sentry_issue_alert ...
```

`27 + 3` is correct. The step hardcoded `2`, and it was **wrong on arrival, not stale**: `git_data_boot_warning` landed in `0f39b7aa2` (PR #7805) two days *before* the commit that wrote the `2` — whose own message counts three while its gate asserts two. (An earlier draft of this description blamed post-authoring drift and cited a "T25 prose-count gate"; both were wrong. No such gate exists, and the README correction is in #7868, which landed after this branch's base. The honest version is stronger support for deriving: the number was never right.)

Ref #7865 — deliberately `Ref`, not `Closes`. The workflow **resolves that tracker itself** on a green apply (`Close the apply-failure tracking issue (success)`, gated on `if: success()`), with a comment naming the run. A closing keyword would resolve it at merge, on a promise, before any apply has proven the fix — exactly what ship's Closes-after-verification gate forbids. Let the green apply close it.

## What review changed

Review found the defect class this PR closes, inside the fix itself. Four of the merge-blocking findings were in guard code the fix added.

- **The new guard could not fire.** `grep` exits 1 on no match; Actions runs a bare `run:` as `bash -e {0}` so errexit is inherited, and `set -o pipefail` promotes it out of a pipeline. All four `VAR=$(grep ...)` assignments killed the step whenever a count was legitimately zero — before the summary table, before any annotation. Measured against the shipped bytes: a genuine partial adoption whose state held no `sentry_issue_alert` exited 1 having printed **nothing**, and the filer then opened a p1 with nothing naming why. The non-vacuity floor was unreachable dead code in exactly the case it was written for. Rescues are now inline at each site.
- **Counts, not address sets.** A one-for-one swap netted to zero: config declaring `sentry_alert.a` against a state holding `sentry_alert.zzz_orphan` scored `rc=0`, with the summary printing `1 | 1` as evidence of health. The step is named "no orphan" and the filer promises it; neither was true. It now compares sorted address sets (`LC_ALL=C comm`) and names both sides.
- **The sweep stopped at the gate.** The step `name:`, its header comment, the filer's rationale comment and — worst — the filer's **issue body** all still said "27 + 2". That body is what an operator reads during a suspected paging outage, and it told them a healthy 27/3 is "a partial adoption and the highest-priority thing in the repo". Correcting the gate but not the runbook moves the false alarm rather than removing it. The literals are deleted rather than updated; any literal there re-acquires the same rot.
- **AC19/AC20 and the BYOK Art. 33 liveness probe** were on an implicit `success()`, so an AC17 failure skipped them at the moment they matter. Verified from the API on run `34149741385`: AC17 red, fidelity probe `skipped`. Both now carry `always()`.

## Guards (mutation-proven)

- `tests/scripts/test-sentry-ac17-derived-counts.sh` — 10 rows executing the step's own bytes extracted from the shipped YAML (anchored on `id: ac17`, uniqueness asserted) under `bash -e`. Restoring the hardcoded `2` reds 3 rows. **Before this, reverting the entire change left every suite in the repo green.** The verdict reads an append-only ledger, because gating on the pass/fail counters let a one-token rewrite of `_report` report "8 passed, 0 failed" and exit 0 with a real regression injected.
- **T13** in `test-sentry-full-root-apply.sh` is the repo's existing gate for this class and matched only the bare-command-then-`rc=$?` shape, so it was blind to all four of these lines. Widened to `VAR=$(...)`, to helper definitions, and to backslash-continued lines — the first fix routed the counts through a `_lines()` helper and the widened rule went green because it could no longer *see* the command, not because the command was safe.

## Verified in production

- `terraform state list` → 27 `sentry_alert.` + 3 `sentry_issue_alert.`, no orphan
- AC19/AC20 probe, recovered out-of-band (run `34144160724`) → **PASS**, all 27 match the committed capture field-for-field
- AC21 → all four BYOK rules present and ENABLED

## Post-incident report

`knowledge-base/engineering/operations/post-mortems/2026-09-08-ac17-stale-constant-cried-wolf-postmortem.md` — required by ship's Incident-PIR gate. Not a user-facing outage (live paging verified correct throughout); the damage was to assurance and to trust in the signal. Art. 33/34 both false.

### CI / Sentry infra

- AC17 derives its expected `sentry_alert` / `sentry_issue_alert` counts from the Terraform config instead of hardcoding them, and compares address sets so a compensating rename no longer passes as healthy.
- AC17 no longer dies silently when a count is legitimately zero; its failure and non-vacuity messages actually print.
- The post-apply live-fidelity and BYOK Art. 33 liveness probes are no longer skipped when AC17 fails.
- The auto-filed tracker's body no longer instructs the operator to treat a healthy state as a partial adoption.

## Test plan

- [x] `tests/scripts/test-sentry-ac17-derived-counts.sh` — 10/10, mutation-proven (restoring `2` reds 3 rows)
- [x] `tests/scripts/test-sentry-full-root-apply.sh` — 13/13 with the widened T13; 5-mutation battery all caught, control green either side
- [x] All 9 sentry suites green
- [x] `actionlint` adds no new rule class vs `origin/main` (the 2 `SC2035` this PR had introduced are gone)
- [x] `TEST_GROUP=all bash scripts/test-all.sh`
- [ ] Post-merge: the next `Apply sentry infra` run on `main` goes green and AC19/AC20 stops being skipped — the workflow's own success-path step then resolves that tracker on its own

**Net-issue-flow override.** This PR files one issue and closes none, so `NET` is `+1`.

- **#7941** — while running this PR's own Phase 7 required-check fetch, I measured that `adr-ordinals` IS in the live required set for `main`, while `ship/SKILL.md` asserts twice that it is not and builds its ADR-collision guidance on that claim. A discovered defect in another subsystem (`plugins/soleur/skills/ship/SKILL.md`), which net-issue-flow names as a legitimate override case — folding a ship-skill prose correction into a Sentry-infra CI fix would be unrelated scope. It is the same defect class this PR exists to close: a prose claim nobody re-checks, justifying a mechanism.

This PR does not carry a `Closes` keyword by design (see the `Ref #7865` note above), so the tracker it resolves is not counted as `CLOSING`.


deruelle:
## This is now a blocker, not a cleanup

PR #7868 merged at 16:34 UTC and fired the first apply on this root since the AC17 constant went
stale. It reds, exactly as this PR predicts:

```
AC17: state holds 27 sentry_alert and 3 sentry_issue_alert address(es)
AC17 FAILED: expected 27 sentry_alert and 2 sentry_issue_alert addresses in state, found 27 and 3.
```

Run [34143928574](https://github.com/jikig-ai/soleur/actions/runs/34143928574).

**Three consequences that were not true before today:**

1. **Every future apply on `apps/web-platform/infra/sentry/**` now reds.** The 27→30 migration is
   complete, so state legitimately holds 3 `sentry_issue_alert` (`auth_per_user_loop`,
   `git_data_boot_warning`, `sandbox_startup_failure`) and will keep holding 3. The gate's
   hardcoded `2` cannot become true again on its own.
2. **AC19/AC20 is collateral.** `sentry_alert live fidelity` is step 15, immediately after AC17, so
   the stale constant *skips the live-fidelity check on every apply*. I recovered it out-of-band this
   time by dispatching `scheduled-sentry-alert-drift.yml` (run
   [34144160724](https://github.com/jikig-ai/soleur/actions/runs/34144160724) → `PASS`), but that is
   a manual step standing in for an automatic gate.
3. **The error text misdirects.** It reads *"This is the partial-adoption signal: some of the 27 moved
   and some did not."* That is false — all 27 moved, and the apply succeeds before the assertion runs.
   A reader who trusts it will go looking for a partial adoption that does not exist. Worth correcting
   the message in this PR alongside the derived counts.

Nothing is wrong with live state: `Apply complete! Resources: 1 added, 0 changed, 0 destroyed.` and
the post-apply fidelity probe passes field-for-field. The gate is the only broken thing.

deruelle:
The red `deploy-script-tests` check is **pre-existing and unrelated to this PR** — tracked in #6286.

Measured rather than assumed:

- Failing step: `Run cloud-init inngest-bootstrap pin drift-guard`, 127/129 passed. The two failures are `cloud-init pin (v1.1.25) matches latest published vinngest-v* tag (v1.1.26)` and its dedicated-host twin.
- This PR touches **neither** cloud-init file (`git diff --name-only origin/main...HEAD | grep cloud-init` is empty).
- The `v1.1.25` pin is byte-identical on this branch and `origin/main` (2 occurrences per file, both sides).
- `vinngest-v1.1.26` now exists as a tag; the guard compares the pin against the live semver-max tag at run time, so it reds on any branch.
- `main`'s last Infra Validation runs are all green because they predate the tag.
- `deploy-script-tests` is **not** in the required-context set for `main`, so it does not gate this merge.

#6286 is the live tracker for this recurring drift class (it was last updated one minute before this run failed). Filing a second issue would have been a duplicate, so I did not.

Issue #7823, opened 2026-09-04
chore(runbook): recover-userid-from-pino-stdout prescribes two raw prod SSH logins


deruelle:
## Problem

`knowledge-base/engineering/operations/runbooks/recover-userid-from-pino-stdout.md` prescribes two
raw host logins:

- L93 — `ssh root@<prod-ip> "docker logs soleur-web-platform ..."` (Flow 1, userId recovery)
- L188 — `ssh root@<prod-ip>` (Flow 2, the retention measurement)

Both violate `hr-no-ssh-fallback-in-runbooks` and `hr-no-dashboard-eyeball-pull-data-yourself`: the
Soleur operator is non-technical, and a runbook whose only read path is a root shell is not a read
path they can use.

## Why it is only surfacing now

`lint-infra-no-human-steps` runs **changed-vs-base**, so it lints a file only when a PR touches it.
Both lines are **pre-existing on `main`** — PR #7782 edited an unrelated paragraph in this runbook
(resolving the `__TBD_OBSERVED_VOLUME__` narrative) and that pulled the whole file into scope,
turning `lint-bot-statuses` red. That check is **not** in the required set, so it does not block
merge; it will stay red for any future PR that touches this file.

## Why it was not fixed inline

The fix is not a doc edit. This runbook exists *because* pino stdout is not shipped anywhere — its
own steps confirm there is no off-host copy of the container's json-file buffer to query
(`systemctl list-units | grep -iE 'promtail|vector|fluent|filebeat|rsyslog'` -> "zero matches
expected"). Satisfying the rule means **building** a no-SSH read path for that surface, not
rewording the runbook.

## What done looks like

Either:

1. Ship a no-SSH read path for the web container's pino stdout — the shape used elsewhere in this
   repo is a `SOLEUR_*` marker emitted into journald -> Vector -> Better Stack, queried with
   `scripts/betterstack-query.sh`; or
2. Establish that Flow 1 is obsolete (the `userIdHash` lookup may already be answerable from Sentry
   via `scripts/sentry-issue.sh`) and delete the SSH flow outright.

Then remove the two `ssh` lines so `lint-infra-no-human-steps` passes on the file.

## Not in scope

Suppressing it with the lint's `lint-infra-ignore` region. That escape exists for
deferred-orchestrator prose, not for a genuine standing instruction to SSH into production.

## Scope-Out Justification

Different subsystem (observability read paths) from #7782's statutory-register scope, and the
[...]

Issue #7791, opened 2026-09-03
compliance: CLO attestation for the Art. 33(5) breach register was not obtained (AC14 unmet)


deruelle:
`knowledge-base/legal/breach-register.md` shipped in #7717 without a CLO attestation. The plan's task 1.13 called for one at `knowledge-base/legal/audits/2026-09-03-clo-attestation-7717-art-33-5-register.md`; that path is deliberately absent, and acceptance criterion **AC14 is unmet**.

### Why it is missing

Non-substantive. The `clo` agent was invoked to produce it and terminated on an API 529 (server-side overload) three times:

- `req_011CegcTbK6bQXipPMqhFHVS`
- `req_011CegcmkyX9kXDsZxkqP2fn`
- `req_011Cegd6WcbnQP2vaVHPMcTW`

An attestation is an act of authority. Writing one and signing it `signed_off_by: CLO agent` would have fabricated a signature — the exact defect class #7717 exists to close, a record asserting an authority it does not hold. So the gap was recorded rather than papered over.

### What DID happen, so this is not starting from zero

The CLO's two **binding rulings** were obtained before implementation and applied as given. They are reproduced with their grounds in `knowledge-base/legal/audits/2026-09-03-implementation-record-7717-art-33-5-register.md`:

1. **The determination set is 4 indexed / 3 waived**, not the 7 the plan specified — the plan had derived its set from a regex two paragraphs after warning that no regex makes a legal judgement.
2. **`controller:`, not `processor:`** — AC1's field list overruled, on the ground that the identity field names the capacity the register is kept in and Art. 33(5) binds the controller on its face.

What is missing is the attestation artifact: per-artifact verdicts and a DISCHARGED/BLOCKED disposition under CLO authority.

### Scope

- [ ] Invoke `clo` to produce the attestation at the path above.
- [ ] **Ratify or overrule one correction made without CLO sign-off.** The Sentry row's "evidentiary limbs inconclusive" cell was corrected against its source before landing: the CLO's drafted cell cited a **pre-Phase-9** statement ("no arms-length external-party exposure positively confirmed", post-mortem L150). Phase 9 closed both questions — Sentry support confirmed in writing 2026-05-19 that both orgs are operator-owned and every audit-log action in both was the operator's, and the affected population is enumerated. The cell now reads "Partially" and names the T3 granular-causal residual as an engineering residual rather than an evidentiary limb bearing on Art. 33/34.
[...]

Issue #6474, opened 2026-07-15
legal(gdpr): Art. 30 register states a log-retention bound the container does not use (30 MB daemon default vs journald SystemMaxUse=1G)


deruelle:
> **Draft material — requires professional legal review before it is relied upon.**

## Summary

`knowledge-base/legal/article-30-register.md` (PA8 §(f)) states container log retention is **"30 MB rolling (max-size=10m × max-file=3)"**. That is **only the Docker daemon default** (`cloud-init.yml`, the `daemon.json` write). The `soleur-web-platform` container runs with **`--log-driver journald`**, which **overrides** it. Real retention is bounded by journald: **`SystemMaxUse=1G`** (`journald-soleur.conf`).

The register's Article 30(1)(f) description of the retention bound is therefore **wrong** — not stale prose, a factually incorrect statement in a compliance register.

## Why this was nearly missed (the important part)

The claim sits behind a **rotted line-number anchor**. The obvious "fix" — re-pointing the citation at the current line — would have left the register **wrong AND freshly "verified"**: the anchor would resolve, the sentence would still be false, and the rot would have been laundered into apparent correctness. It was caught only because the anchor's true target was read rather than the line number re-pointed.

This is why it is **not** a citation fix and was split out of the #6436/#6429/#6446/#6447 batch (whose Phase 1 sweep deliberately drops line numbers in favour of grep-able anchors). Fixing the substance is legal work, not mechanical work.

## Scope

1. **Correct PA8 §(f) substance** — journald `SystemMaxUse=1G`, not the 30 MB daemon default. Trace the redirect: the `--log-driver journald` switch and the later `SystemMaxUse=1G` bound landed in separate PRs; state what is true **today**.
2. **Re-anchor to executable evidence, not line numbers** — `journald-config.test.sh:71` already asserts `SystemMaxUse=1G`. A test that fails when the claim stops being true is a far stronger anchor than a citation.
3. **Fill the unfilled placeholders** — `__TBD_OBSERVED_VOLUME__` and `__TBD_BETTERSTACK_RETENTION__` are still literal in the register. An unfilled placeholder in an Art. 30 register is an **independent Art. 30(1)(f) defect**.
4. **Repair the DSAR runbook step** — its `docker inspect … .LogPath` step returns **empty** under the journald driver. That is an Article 15 / DSAR step that **cannot work as written**.
5. **Rebase #3754** — it measures against a cap (the 30 MB rolling bound) that does not apply.
[...]

PR #7875, opened 2026-09-06
compound: the #7791/#7823 session learning, and the citation #7872's deferral was missing

deruelle:
Follow-up to #7838 (#7791) and #7872 (#7823). Two files, no behaviour change.

## 1. The session learning

`knowledge-base/project/learnings/2026-09-06-a-live-positive-validated-the-filter-on-one-row.md`

The headline is a measurement error I made and then nearly shipped: #7872's runbook query was
**run live against production, returned a real record, and was recorded as verified** — while
silently dropping most of the population. `.message` arrives in two shapes (Better Stack
auto-parses a JSON-valued `message` at ingest, so pino records land as OBJECT; fd-2 crash lines
stay STRINGS), the filter carried `select(type == "object")`, and the one matching row happened
to be in the object branch.

**A live positive proves the filter works on that row.** Validating a *filter* needs the
population's shape distribution — and the successful production run is the most persuasive
possible evidence for the wrong conclusion, precisely because it is real.

Also recorded, each measured rather than recalled: resolving agent disagreement by measurement
(two agents shared one unmeasured premise and were therefore one agent); the jq `//`
falsy-default written in hours after quoting its own warning into a review prompt; **five**
instruments that needed checking before their verdicts could be trusted; a sweep indexed by
remembered wording that silently did not apply and was reported as fixed; and a date asserted
from charity inside a PR about unmeasured premises.

## 2. The citation #7872's deferral was missing

One line in the pino runbook. Its Flow 2 correction states that Art. 30 PA-8 §(f) records a
`json-file` retention mechanism that does not apply to a container running `--log-driver
journald` (since #4786), and defers the register edit as out of scope. Accurate, but it pointed
at nothing — the deferral now cites **#6474**, which tracks exactly that.

## Verification

`markdownlint` 0 issues; `lint-infra-no-human-steps.py` OK on the runbook. Net issue flow: closes
0, files 0.

https://claude.ai/code/session_016TzeL4qU1DJfFmDft4zAGM


deruelle:
**Review coverage: 0 of 4 agents — `inline-fallback`, recorded in the trailer.** Stated here explicitly per the review skill's Gate 2a, so it is not inferred from a passing gate.

This is a learning document plus a one-line issue citation, with no executable content. A panel would mostly re-verify the document's factual claims, so I verified them mechanically instead — the more targeted instrument for a claims-dense artifact:

| Claim | Verification | Result |
|---|---|---|
| Vector shipped the path 3 months before #7823 was filed | `git log -S app_container_journald` vs the issue's `createdAt` | 2026-06-02 (`223364c14`) vs 2026-09-04 |
| `--host` is REQUIRED on `betterstack-assert-absence.sh` | grep the guard | present; omitting it exits 64 |
| `formatters` is defined in `server/logger.ts` | grep the definition | confirmed |
| jq `//` short-circuits past a later matching key | ran it | `MISSED` — the P1 claim holds |

`markdownlint` 0 issues; `lint-infra-no-human-steps.py` OK on the runbook; net issue flow 0 (closes 0, files 0).

Gate 2a's block on marking ready with zero agents applies to a `single-user incident` brand-survival threshold; this PR has no plan file and no such threshold.

Issue #7208, opened 2026-08-03
memory backstop (#7166): post-merge hardening from the PR #7169 review panel


deruelle:
Post-merge hardening for the memory backstop shipped in PR #7169 (ADR-158).

The five-agent review panel on that PR produced more findings than belonged in one
PR. Everything that made the shipped feature **wrong** was fixed inline before merge
(re-entry verification was tautological, the slice call was unverified and
all-or-nothing, `ManagedOOMPreference` was armed on the unit oomd never consults, and
two test vacuities that were *proven* by experiment). What remains is hardening —
each item makes an existing guard sharper, none of it changes whether the feature
works today.

Grouped by what a fix would touch. Sources: `security-sentinel`, `silent-failure-hunter`,
`test-design-reviewer`, `architecture-strategist`, `code-simplicity-reviewer`.

## A. AC7 sweep is narrower than its name claims

`.claude/hooks/memory-backstop.test.sh` — the before/after cgroup sweep excludes
everything under `/soleur.slice` and roots at `user@<uid>.service`. Consequences,
each a surviving mutation:

- [ ] A per-run scope **leak** (mutate the scope name to include `$RANDOM`) is invisible —
      and that is the previous attempt's literal failure mode.
- [ ] `SLICE_NAME="soleur.slice"` (fleet cap lands one level up, on a slice the hook does
      not own) is invisible.
- [ ] `SLICE_NAME="user-$(id -u).slice"` — capping the operator's entire login session at
      20 GiB — is *above the sweep root* and not sampled at all.
- [ ] Only `memory.{high,max,swap.max,low,min}` are sampled; `pids.max`, `cpu.max`,
      `memory.oom.group` on a foreign unit escape.
- [ ] The sweep window spans only the first hook run, so out-of-scope writes that occur
      only on the **re-entry refresh** are outside it.

Fix shape: root at `/sys/fs/cgroup`, replace the blanket exclusion with an allowlist of
the exact three paths the hook may create, so a *fourth* entry under `soleur.slice` fails.

## B. Mutation battery coverage gaps

`.claude/hooks/memory-backstop-mutation-battery.sh`:

- [ ] No mutant touches `discover_claude_pid`, `read_ppid`, `read_starttime`,
      `validate_caps`, `collect_descendants`' bounds, `attribution_message`, or the
      `final_cg` membership check — i.e. the identity walk and the membership
      verification, the two mechanisms the design leans on hardest, are unmutated.
[...]

PR #7879, opened 2026-09-07
fix(test-fixtures): fixture-env adoption, ledger isolation, and the ancestry walk (#7849, #7853, #7854)

### LINKED ISSUE #7849: test-fixtures: adopt gitFixtureEnv() at every fixture-creating suite (defence-in-depth behind #7833's process boundary)
Defence-in-depth follow-up behind the process-boundary fix shipped for #7833.

## What is deferred

Per-suite conversion to `gitFixtureEnv()` / `gitFixture()` at every fixture-creating suite. #7833
shipped three layers that make this unnecessary *in any reachable invocation*:

1. the `unset` on lefthook's `plugin-component-test` entry point,
2. the Guard 2 ratchet over the closed hook entry-point set, and
3. the Guard 3 fail-loud tripwire in all three runner runtimes (bun preload, both vitest
   setupFiles, the `test-helpers.sh` shell prelude), which aborts with `rc=97`.

It was deferred, not skipped, because (a) it buys nothing over layers 1-3 in any invocation we can
reach, (b) the beneficiary set is **not reliably greppable** — `git grep -l 'git init'` misses the
`execFileSync("git", ["init", …])` array form and every python list-form spawn — and (c) coupling a
~50-file mechanical diff to three new guards guarantees a rubber-stamp review.

## Two conditions that would make this load-bearing

Either of these turns the deferral from defence-in-depth into a real gap:

1. **A runner that cannot host a tripwire.** Guard 3 assumes one prelude per runtime. A new runner
   with no preload/setup hook has no layer-3 protection, so its suites need the helper directly.
2. **A fixture that must run `git` with a partially-inherited environment.** The helper returns a
   constructed env; a suite that genuinely needs to keep some inherited git state cannot use the
   tripwire's all-or-nothing rule and needs a per-site scrub instead.

## Files that fall between "converted" and "explicitly waived"

Enumerated so nothing sits in the gap. #7833 converted the three Bun suites under
`plugins/soleur/test/` plus the two python suites under `tests/scripts/`, and explicitly waived the
six `apps/web-platform/test/workspace*` suites (already correct for their runtime — vitest runs on
Node, where a `process.env` deletion propagates) and `agent-ready-git-worktree.test.ts` (sets
`GIT_DIR` deliberately as its own subject; the tripwire runs at process start, before the test sets
it). Neither converted nor waived:

- `test/pre-merge-rebase.test.ts` — Bun-run, already isolated via its own `GIT_ENV` destructure.
  Correct today, but a fourth spelling of the same idea.
- `apps/web-platform/test/cc-reprovision-git-discriminator.test.ts` and `worktree-config-seed.ts` —
  vitest, `git init`, no `GIT_DIR`/ceiling handling of their own. Covered by Guard 3's vitest
  prelude, not by any per-file scrub.
- `apps/web-platform/infra/workspaces-luks-loopback.test.sh`
- `apps/web-platform/test/helpers/context-queries-fixture.ts` — a helper, not a `*.test.ts`.
- the four `tests/**` shell suites, and `.github/scripts/test/test-*.sh`.

The earlier single `git grep -l 'git init' -- '*.test.sh' '*.test.ts'` derivation is structurally
blind to every one of these: python list-form spawns, the `test-*.sh` naming convention, and helper
files all escape it.

## Re-evaluation triggers (carried here so the review is scheduled, not remembered)

1. **lefthook or git stops exporting `GIT_DIR` in a linked worktree.** Re-run measurement M-1. If it
   is gone, the `lefthook.yml` prefix becomes redundant and can be retired; Guards 1 and 3 still
   hold, because they set the hostile environment themselves rather than relying on an ambient one.
2. **The fixture helpers are consolidated so scrubbing happens in exactly one place** — e.g. if
   `plugins/soleur/test/*.test.sh` reaches full `test-helpers.sh` adoption (measured 2026-09-04:
   41 of 73 suites source it). At that point Guard 3's shell arm becomes a true chokepoint and this
   issue can be closed as unnecessary.
3. **Bun changes `process.env` deletion propagation.** Measured on Bun 1.3.11: `delete
   process.env.GIT_DIR` takes effect in-process but a child spawned *without* an explicit `env`
   still receives the original **inherited** value; Node v22.22.2 propagates the deletion. Mutation
   row M6 rests on that divergence and must be re-derived if it changes. Note `.bun-version` pins
   **1.3.14** while the measurement was taken on 1.3.11 — unverified against the pin. If Bun starts
   matching Node, an in-process scrub becomes a viable *additional* layer, though the tripwire
   should still fail rather than scrub.

Re-probe caveat for trigger 3: the divergence appears **only for an inherited variable**. A probe
that sets `process.env.GIT_DIR` and then deletes it in the same process shows both runtimes clean
and will wrongly read as "Bun is fine".

These are conditional revisits, not external-dependency soaks, so this is a plain tracking issue —
deliberately **not** `follow-through`, which the daily sweeper reserves for external dependencies
awaiting verification and drives from a verification script.

Background, measurements and guard contracts: #7833 and
`knowledge-base/project/specs/feat-one-shot-7833-git-dir-beats-cwd/measurements.md`.

### LINKED ISSUE #7853: test-fixtures: gdpr-gate-self-test writes fabricated deny events into the operator's real incident ledger
Found while running `/soleur:compound` Phase 1.5 step 3.5 for #7833, which reads
`.claude/.rule-incidents.jsonl` as **deviation evidence**.

## Measured

`plugins/soleur/test/gdpr-gate-self-test.test.sh` invokes the real
`plugins/soleur/skills/gdpr-gate/scripts/gdpr-gate.sh` against synthetic fixtures without setting
`INCIDENTS_REPO_ROOT`, so every run appends to the operator's live ledger.

Before/after around one run (ledger restored from a copy afterwards):

```
suite rc=0  ledger before=47 after=55 delta=8
```

The 8 appended rows:

```
gdpr-gate-staleness            warn        307-days-stale
gdpr-gate-staleness            deny        307-days-stale-posture-fail
hr-gdpr-gate-on-regulated-data-surfaces  applied  regulated-data path touched: apps/web-platform/lib/auth/foo.ts
gdpr-gate-cron-binding         unavailable no-token-or-gh-cli
gdpr-gate-cron-binding         min-wins    cron=215 notice=307
gdpr-gate-staleness            warn        215-days-stale
gdpr-gate-staleness            deny        215-days-stale-posture-fail
hr-gdpr-gate-on-regulated-data-surfaces  applied  regulated-data path touched: apps/web-platform/lib/auth/foo.ts
```

The ledger already holds 47 such rows from earlier runs today (35/95/999-day prefixes, 15 of them
`deny`).

## Why it matters

Three consumers read this file as a record of what actually happened:

1. `compound` Phase 1.5 step 3.5 admits `event_type ∈ {deny, bypass}` as deviation evidence —
   "a hook caught a violation". These denies caught nothing; the staleness is a fixture.
2. `scripts/rule-metrics-aggregate.sh` rolls the ledger into `knowledge-base/project/rule-metrics.json`,
   which **is committed**. Fixture rows inflate the committed hit counts and suppress the
   `rules_unused_over_8w` pruning hint for `gdpr-gate-staleness`.
3. `apps/web-platform/lib/auth/foo.ts` does not exist. The ledger asserts a regulated-data surface
   was touched, on a path that has never existed in this repo.

## Class

`hr-write-boundary-sentinel-sweep-all-write-sites`. This is the same shape as #7833 (a test writing
into the developer's real state because the redirect knob was not passed), and the second miss of
the same sweep: `d0d3b5d68` ("hook suites no longer write the real incident ledger") sandboxed
`.claude/hooks/rule-incident-marker-capture.test.sh` — whose header comment already says *"Inline
per-call `INCIDENTS_REPO_ROOT=… bash "$HOOK"` is what leaked here"* — but did not reach this suite,
which invokes the gate script rather than a hook.

## Suggested fix

1. Set `INCIDENTS_REPO_ROOT="$SANDBOX"` on every `gdpr-gate.sh` invocation in the self-test, matching
   the pattern already used at `rule-incident-marker-capture.test.sh:62`.
2. Sweep for the *general* case rather than this one site: enumerate every test that execs a script
   which sources `.claude/hooks/lib/incidents.sh`, and pin the set. A per-file fix has now been
   applied twice and missed a sibling both times.
3. Consider a tripwire in `incidents.sh` itself — refuse to append when the process looks like a test
   run and `INCIDENTS_REPO_ROOT` is unset — so the ledger cannot be corrupted by a *transitive* spawn
   that no call-site grep can see (the #7833 lesson).
4. Decide whether the 47 already-written rows should be discarded before the next
   `rule-metrics.json` refresh.

## Scope-Out Justification

Different subsystem (gdpr-gate / incidents telemetry) from #7833's branch (test fixture git
environment). Per `compound` triage, a discovered defect in a different subsystem stays its own issue
and PR rather than being bundled into an unrelated feature branch.

https://claude.ai/code/session_019ekjPGkwTTAUNZLZsgieL2

### LINKED ISSUE #7854: memory-backstop.test.sh fails under lefthook: its ancestry walk and the hook's disagree at depth
Blocks `git commit` for any diff whose lefthook `bun-test` step runs the full
`scripts/test-all.sh`. Found while committing #7833's compound artifacts.

## Measured

Standalone:

```
$ bash .claude/hooks/memory-backstop.test.sh
RESULT: PASSED 54 [live: yes]   rc=0
```

The same suite, inside `lefthook run pre-commit` -> `test-all.sh`:

```
✗ T8 real hook did not apply (outcome='skipped' reason='claude_pid_not_found')
[FAIL] .claude/hooks/memory-backstop.test.sh (5347ms)
```

## Mechanism

The suite already anticipates the detached-runner case and gates its end-to-end arm on an
independent 8-hop `/proc` ancestry walk (`memory-backstop.test.sh:683`), with the comment:

> "A user bus exists" does NOT imply "a claude ancestor exists" ... Measured: running
> scripts/test-all.sh under nohup did exactly this.

Under lefthook the tree is deeper than the detached case it was written for:

```
claude -> bash -> git commit -> lefthook -> sh -> bash(test-all.sh) -> bash(suite)
```

The two walks then disagree: the suite's walk reaches `claude` and sets `E2E=yes`, while the
hook's own derivation (`memory-backstop.sh:414`) does not and reports
`outcome=skipped reason=claude_pid_not_found`. The suite therefore asserts an adoption that
correctly never happened, and fails on an environment fact — exactly the outcome its own gate
exists to prevent, one process-tree depth further out.

## Suggested fix

The gate and the thing it gates must derive the same fact. Either:

1. Have the suite ask the **hook** whether it found a Claude PID, rather than re-deriving it
   with a different walk — the disagreement is the bug, and a second independent walk cannot
   fix it because the two can always diverge; or
2. If the independent derivation is deliberate (the comment says it is, so the same code cannot
   satisfy the gate it gates), then match the hook's traversal limit and termination conditions
   exactly, and add a fixture at lefthook depth so the divergence is caught by CI rather than by
   someone's commit.

Option 1 is the smaller change; option 2 preserves the independence the comment is defending.
Whoever picks this up should decide — I have not.

## Impact

`wg-when-tests-fail-and-are-confirmed-pre` applies: confirmed pre-existing and out of scope for
#7833's branch. But it is not harmless — it fails the pre-commit gate, so the practical effect
is pressure to reach for `--no-verify`, which disarms every other hook too.

## Scope-Out Justification

Different subsystem (memory backstop / systemd scope adoption) from #7833's branch (test fixture
git environment). Per `compound` triage, a discovered defect in a different subsystem stays its
own issue and PR.

https://claude.ai/code/session_019ekjPGkwTTAUNZLZsgieL2

### LINKED ISSUE #7886: memory-backstop.test.sh fails under lefthook and passes standalone — pre-commit goes red on unrelated diffs
`.claude/hooks/memory-backstop.test.sh` fails when run by lefthook's `pre-commit` and passes when run directly, so every agent-driven `git commit` that stages enough files to trigger it sees a red gate for a reason unrelated to the diff.

## Measured, both directions, same tree

| invocation | result |
|---|---|
| `bash .claude/hooks/memory-backstop.test.sh` | `RESULT: PASSED 54 [live: yes]`, exit 0 |
| the same suite under `lefthook run pre-commit` | `RESULT: FAILED 1 (passed 46) [live: yes]` |

The single failing arm is:

```
✗ T8 real hook did not apply (outcome='skipped' reason='claude_pid_not_found')
```

Note the case counts also differ (54 vs 47), so the live-arm ledger resolves differently under the hook, not just that one assertion.

## Why it matters

The hook looks for a Claude Code PID, and under lefthook's process tree (`git commit` → `/bin/sh pre-commit` → `lefthook` → suite) that lookup fails. The consequence is not a wrong verdict about memory backstopping — it is that the pre-commit gate goes red on commits whose diff has nothing to do with this hook. That is the "trains the reader to walk past red" shape: a gate that fails for environmental reasons stops being read, and the next real failure in it gets waved through too.

Discovered while shipping #7855 / PR #7856, whose branch touches neither `.claude/hooks/memory-backstop.sh` nor its test (`git log origin/main..HEAD -- <both paths>` is empty).

## Plausible directions, not yet verified

- Make the Claude-PID lookup resolve through the lefthook ancestry, or
- have `T8` declare itself SKIPPED rather than FAILED when no Claude PID is discoverable, since "no Claude session in this process tree" is a legitimate context rather than a defect — the suite already has an explicit skip/ledger mechanism for its live arms.

Either way the fix should keep the assertion meaningful when a Claude PID *is* present.

Ref #7856.

deruelle:
Closes #7849
Closes #7853
Closes #7854
Closes #7886

> **#7886 transferred in** (see the maintainer's comment on this PR). A parallel `/one-shot`
> dispatch re-probed after planning, found this branch already carried the fix, and cut its PR 1
> rather than race it — no file under `.claude/hooks/**` is touched on
> `feat-one-shot-7867-7873-7886-betterstack-credfwd-lefthook` (PR #7894). The commit is
> `8d6d7ec51` here (`4115024f5` before this branch's rebases): the memory-backstop e2e
> precondition now asks the hook for its own verdict instead of re-deriving one, which is what
> made the suite pass standalone and fail under lefthook. Without this line the issue would be
> orphaned when the fix merges.

Three issues in the test-fixture / git-hook area, shipped together because they share one surface: what a test suite is allowed to touch outside its own fixture.

## #7854 — memory-backstop's ancestry walk vs the hook's

The suite failed under lefthook because its walk and the hook's disagreed at depth. Reconciled, and pinned with a mutation battery: **12 killed, 0 survived**.

While fixing it, `newtmp` — the chokepoint rooting **eleven** fixtures in that suite — turned out to accept a relative fixture root. It calls `mktemp -d -t`, which is only absolute when `TMPDIR` is, and the file exports `TMPDIR="${TMPDIR:-/var/tmp}"`, which defaults an *unset* TMPDIR but passes an inherited relative one straight through:

```
$ TMPDIR=reltmp mktemp -d -t membackstop.XXXXXXXX
reltmp/membackstop.DN0hD7to        rc=0
```

`mktemp` does not object, so nothing below it does either, and all 22 `rm -rf`/redirect/`cp` operands then aim at a CWD-relative tree — with `cd "$(git rev-parse --show-toplevel)"` above them. Guarded at the chokepoint (one function, eleven callers) rather than at 22 write sites, because the per-site form is the one that drifts.

## #7849 — adopt the helper at the remaining fixture suites

Converted the remaining fixture-creating suites across vitest, bun, shell and python onto the shared builder, behind a new shell chokepoint (`plugins/soleur/test/lib/git-fixture-env.sh`) that owns one `GIT_LOCATION_VARS` array feeding both the tripwire and the env builder. The sweep is **by prefix, never by name list** — a name list is a claim about which variables git honours, and it is wrong the moment git adds one.

Adoption is enforced by a new guard with four derivations plus non-vacuity floors and driven controls, not by a checklist.

## #7853 — the gdpr-gate self-test wrote fabricated rows into the operator's real ledger

Ledger writes are now contained at five runner entry points. Measured over a 2567-test run: **8 rows/run → 0**.

The aggregator orphan gate from the issue body is in, with a two-clause discriminator: `rc=5` with 23 orphans before, `rc=0` with 0 after, and an injected orphan still reddens it.

### What the review round found underneath it

**The gate that fires most was forging its own telemetry.** Adding `summary.non_corpus_counts` (below) surfaced `post-dispatch-watch-gate` at 2708 — by far the largest emitter and previously invisible. It greps `$HOOK_CMD` **raw**, so any command that merely *quotes* a dispatch armed it, and the label extractor lifted the quoted fixture's PR number back out:

```
grep -oE 'gh pr merge [0-9]+'   ->   "gh pr merge 123"
```

That is where **140 rows** came from — they accumulated in bursts matching this session's own greps for that string, including the commands used to investigate the contamination. Ten sibling hooks already use `strip_command_bodies()` from the library this hook *already sources*. Now it does too, failing toward noise rather than silence.

**A contributor-writable marker could forge keys into a tracked file.** The namespace widening let any shape-valid unprefixed id through, and three of those are read into *named summary fields* of the committed `rule-metrics.json`. Demonstrated end-to-end: a marker in payload markdown, executed on the `gh pr checkout` review path, forges an attacker-chosen key **and** fires a false ADR-156/157/162 alarm. The hook's own security note claimed no consumer reads such rows; the aggregator says otherwise 200 lines away. Rejected before the section-prefix test, with the existing case that *asserted* the hole replaced by one pinning both directions.

**The committed artifact shipped the contamination the issue was filed about.** `rule-metrics.json` was generated 28 minutes *before* the first quarantine, and the quarantine was inverted — it removed 109 exempt rows and left the ones corrupting two real hard rules in `.rule-incidents-2026-09.jsonl.gz`, which the aggregator reads (it merges active + **all** archives).

Cleanup pinned on `rule_id` **and exact snippet equality**, never substring — the archives also hold *real* denials whose payload merely quotes the fixture string. Exact matching removed 30, and the metric delta equals the row count on both rules:

| rule | before | after |
|---|---|---|
| `hr-when-a-command-exits-non-zero-or-prints` | 99 | **79** (20 fabricated, 20%) |
| `rf-never-skip-qa-review-before-merging` | 80 | **72** |

Every removed row was preserved in a quarantine file; both ledgers were backed up before rewriting; the active-ledger rewrite preserves non-JSON lines byte-for-byte rather than reserializing.

`summary.non_corpus_counts` now emits every exempted id with its fire count — the namespace is no longer gated but is still **visible**, so a flood shows up in a diff instead of a ledger that rotates at 5 MB.

## The guards' own blind spots

The review panel's most useful findings were in the checks, not the code:

- **2 of 3 hostile-value assertions were vacuous** — the fixture injected `GIT_CONFIG_KEY0`; git spells it `GIT_CONFIG_KEY_0`, so two keys were compared against a hostile value never set on them. The comment three lines up warns about exactly this failure mode, one key over.
- **The return-check could not see its own subject** — it stripped every double-quoted span, which deletes the call: `eval "$(git_fixture_env "$d")"` → `eval $d`.
- **The waiver rot-check could not tell a 3-name scrub from a 9-name one** — all seven waivers scrub 3/9 (one 1/9). Now measured per name, printed, ratcheted.
- **A guard printed a mitigation for the members that lacked it** — "they source test-helpers.sh, so the tripwire is armed" was false for 7 of 20. Now derived per member and named.
- **`tests/conftest.py` said it was enforced by a parity test that never mentioned it** — now actually compared.
- **Derivation A pooled verbs per line**, the same fail-open its own comment documents one level up: `git status && git commit` classified read-only.

Every one ships with a mutation check showing the guard fails on the defect and only on it.

## What the full gate caught that targeted runs could not

A full-gate run on the rebased tree returned **375/387**. Nine suites failed. Six were real and
all six were mine; three were contention artifacts. Both halves are worth stating, because the
gate's own banner asks for exactly this confirmation before a failure is accepted as real.

**Real, fixed, each re-verified by running the suite:**

| failure | why every targeted run passed it |
|---|---|
| `tests/hooks/incidents` disarmed | this PR's chokepoint exports `INCIDENTS_REPO_ROOT` run-wide, which shadows the `BASH_SOURCE` fallback that suite exists to test. Isolation by lib-copy and by env-var are not additive — the second DEFEATS the first. Passed standalone, RED under the gate. |
| Guard 5's floors scored NO_FIRE | its "load-bearing" floor reported through `fail`, which the ADR-193 mutant stubs to a no-op |
| two floors scored CONSTRUCTION | `readonly MIN_ASSERTIONS=22` is not carried into the mutant by build_mutant's backward widening — an untested floor wearing a passing colour, invisible because CONSTRUCTION is not reported as a failure |
| ADR-129 leak | the derivation-C counters file had no owning trap, in the guard this PR adds |
| repo-wide containment | four converted suites now import from `plugins/soleur/test/lib/`, escaping `apps/web-platform`, and were missing from the manifest — so they would be declined on exactly the diffs that change the helper they depend on |
| P1b baseline stale | +2 sites from this PR's own edits |

**Contention, not defects** — `fanout-suite-scope`, `test-all-runtime-ceiling` and
`test-all-killed-classification` all drive `test-all.sh` as their SUT and saw the parent gate as a
sibling ("refused on an INHERITED sibling count"). In isolation: 36/0, rc=0, rc=0.

## What CI caught that no local run could

Three CI jobs failed on an earlier push — `gdpr-gate-self-test` on both its with-token and
without-token paths, plus advisory `deploy-script-tests`. The cause is worth stating plainly
because it invalidated my own verification.

`trap -p` prints a re-executable command whose body carries bash's **own** quoting, so an embedded
single quote comes back as `'\''`. The sandbox block stripped the outer quotes with `sed` and
re-wrapped the remainder in double quotes, leaving those escapes unbalanced — the composed trap was
a syntax error:

```
test-helpers.sh: exit trap: line 1: unexpected EOF while looking for matching `''
```

`gdpr-gate-self-test.test.sh` therefore exited **2 after printing "ALL TESTS PASSED"**. Three files
carried the defect: the two this PR added the sandbox block to, and
`.claude/hooks/lib/test-incident-sandbox.sh`, which has carried it since before this branch. Fixed
by not re-parsing the quoting at all — `trap -p`'s output round-trips exactly through `eval`.

**Why every local run said it passed.** My sweep harness was
`printf '%s rc=%s' "$(basename "$s")" "$?"` — bash expands arguments left to right, so the command
substitution ran before `$?` was read and every number reported was `basename`'s status, always 0.
An earlier revision of this PR body published "10 affected shell suites: rc=0 each" on that basis.
**That claim was void** and is replaced below by a re-run with the status captured first.

## Verification

- **CI** on the final SHA — the clean-room check; the local full gate has been continuously
  blocked by sibling runs from other worktrees, and `SOLEUR_ALLOW_FULL_GATE=1` was deliberately not
  used because it would corrupt their measurements
- Full local gate on the rebased tree earlier: **375/387**, all six real failures fixed and
  re-verified individually; the other three were contention (36/0, rc=0, rc=0 in isolation)
- Shell suites re-run with `rc=$?` captured before anything else: **rc=0 each**, including
  `gdpr-gate-self-test` (rc=2 before this fix)
- bun **76 pass / 0 fail** (4 files); vitest **104 pass** (8 files)
- Operator ledger delta across the sweep: **+1 row, and it is real** — a
  `hr-monitor-not-run-in-background-for-polling` deny, the hook correctly firing on me. Zero
  fixture rows written by any suite.
- Mutation batteries: memory-backstop **12 killed / 0 survived**; every fix ships a mutant shown
  RED on the defect with a positive control GREEN under the same mutation

The trap regression test's bound is reported honestly: against the old composition only its
prior-trap arm reddens, so the leak arm is a regression guard rather than evidence.

## Deferred

Recorded on #7889 with measurements, not hand-waved: 47 suites outside every guard-5 list, derivation C's `env: process.env` escape, and the telemetry-sandbox gap at the git-fixture-env chokepoint (**measured leak: 0 rows** — latent, and the prerequisite is making `test-incident-sandbox.sh` idempotent, which touches a helper 19 suites share).

Also filed there: **971 leaked `soleur-inc-*` sandbox directories** measured on this machine, so
`test-incident-sandbox.sh`'s cleanup trap is not firing reliably. Pre-existing and in a helper 19
suites share, so not fixed here.


deruelle:
Cross-linking a scope transfer: **#7886 is yours**, please add `Closes #7886` to this PR's body.

#7886 ("memory-backstop.test.sh fails under lefthook and passes standalone") was routed into a parallel `/one-shot` dispatch today. Its post-plan collision re-probe found your local commit `4115024f5` already implements the fix, with the same verdict-gate mechanism that dispatch had independently arrived at — and a fuller reason classification than it had planned. So that dispatch **cut its PR 1**: no file under `.claude/hooks/**` is edited on branch `feat-one-shot-7867-7873-7886-betterstack-credfwd-lefthook` (PR #7894), and #7886 is transferred here rather than raced.

Without `Closes #7886` in the body, the issue is orphaned when this merges — the fix lands and the issue stays open.

Full transfer record, including the 54-vs-47 count reading and two findings that travel with it (#7208 `MAX_WALK_HOPS` not raisable-for-benefit; the `guard-vacuity-floor.test.sh` `MAX_DEFERRED=47` shrink-only trap for any new `.claude/hooks/` suite): https://github.com/jikig-ai/soleur/issues/7886#issuecomment-5571317094

deruelle:
Added `Closes #7886` to the body — thanks for the cross-link, and for cutting the parallel PR 1 rather than racing it.

Verified before adding it rather than taking the SHA on trust, since this branch has been rebased seven times since that commit was written:

- `4115024f5` replayed to **`8d6d7ec51`** here — "fix(test): ask the hook for the e2e precondition instead of re-deriving it"
- the mechanism is live in `.claude/hooks/memory-backstop.test.sh` (the suite asks the hook for its own verdict instead of re-deriving one, which is exactly the standalone-vs-lefthook split #7886 describes)
- #7886 confirmed still OPEN, so it would indeed have been orphaned on merge

The body now carries a short transfer note pointing at your comment and naming both SHAs, so the reason this PR closes an issue it never planned for is legible from the PR itself rather than only from this thread.

One thing that travels with it, from your linked transfer record: the `guard-vacuity-floor.test.sh` `MAX_DEFERRED=47` shrink-only trap for any new `.claude/hooks/` suite. This PR adds no new suite under `.claude/hooks/` — it modifies existing ones (`incident-sandbox-coverage.test.sh`, `memory-backstop.test.sh`, `post-dispatch-watch-gate.test.sh`, `rule-incident-marker-capture.test.sh`) — so that trap is not tripped here. `guard-vacuity-floor` is green on this branch (23/23), which is consistent with that reading.

Issue #7844, opened 2026-09-04
legal: side-letter-register publishes a Co-Member's full legal name in a public repo (same class as #3210 B1)


deruelle:
`knowledge-base/legal/side-letter-register.md` publishes a **`Counterparty | Full legal name`** column in a repository whose visibility is `PUBLIC`. Every row committed to it would be world-readable, copied into every clone and fork, and **not erasable**.

It has gone unnoticed only because the table is empty (`| (none yet) | | |`). Nothing currently keeps it empty.

## Why this is worse than the case that surfaced it

This is the same defect class as Blocking Finding B1 of #3210, where the Corporate CLA register was specified with identity columns on the same surface. The CLO ruled those columns out permanently (`knowledge-base/legal/audits/2026-09-04-clo-ruling-ccla-register-siting-and-coverage-map-basis-3210.md`, option B1-c).

But the side-letter case is **strictly worse**, not equivalent:

- The CCLA register's organisation legal name is only *conditionally* personal — it is a natural person's name for a sole trader, which is why the ruling's amendment B1-c-2 publishes an opaque `Record ref` alone in that case.
- `side-letter-register.md`'s `Counterparty` is defined as *"Full legal name of the Co-Member"* — **always** a natural person, published alongside a workspace UUID and a signature timestamp. That is a directly-identifying triple, permanently, on a surface from which erasure is impossible.

The CLO's reasoning transfers directly, because it is **surface-based, not instrument-based**: *"tracked in a public repository ... every row committed to it is world-readable ... cannot afterwards be erased."*

## Why this was not fixed in the PR that found it

Not because it is a different instrument or a different custodian — both registers carry `custodian: clo`, so that distinction does not exist and should not be offered.

The real reason: `side-letter-register.md` records that `knowledge-base/legal/audits/2026-05-counsel-review-4289.md` carries **operator-attested counsel sign-off on this register's schema**, and that the same audit file holds its external-counsel re-review triggers. Editing the schema in place invalidates a recorded counsel attestation. That is a counsel re-review event, not a documentation edit.

## Interim mitigation already landed

PR #7828 adds a write-side guard note to the register's `## Notes` block: do not append a `Counterparty` legal name until this is resolved. That is a note, not a control — it protects nothing if someone does not read it.

## What resolving this looks like

[...]

Issue #7632, opened 2026-08-19
supply-chain: the web host pins the inngest bootstrap image by mutable tag while the dedicated host pins by digest


deruelle:
## What

`apps/web-platform/infra/cloud-init.yml` (web host) pins the inngest bootstrap image **by tag only**:

```
IREF=ghcr.io/jikig-ai/soleur-inngest-bootstrap:v1.1.25
ZIREF="$ZURL/jikig-ai/soleur-inngest-bootstrap:v1.1.25"
```

`cloud-init-inngest.yml` (dedicated host) pins the **same image** by tag **and digest**:

```
...soleur-inngest-bootstrap:v1.1.25@sha256:f23a2a0d730d…faba49
```

## Why it matters

The dedicated host's own comment states the hazard, and it applies equally to the web host:

> MUTABLE tag: anything able to re-point `:v1.1.24` between the build and a host replace … the digest and ignores the tag, so these are immutable bytes.

A tag is mutable. Between the build and a web-host replace, anything able to re-point that tag changes what the web host boots, and nothing in the boot path would notice. The digest-pinned host is immune to that; the web host is not. Same image, same registry, two different guarantees.

## Scope

Append `@sha256:<digest>` to both references in `cloud-init.yml`, and keep it in step with the dedicated host on every re-pin — they are already re-pinned together, since AC6 fails if either drifts from the max tag (see #7630).

## Why it was deferred rather than folded into #7630

#7630 fixes a dark host. Widening the digest pin changes the boot path of **web-1**, currently the healthy production scheduler, and that deserves its own diff and review rather than riding along in an unrelated fix.

## Acceptance

- [ ] Both `cloud-init.yml` inngest-bootstrap references carry `@sha256:`
- [ ] A guard assertion fails if either web reference is tag-only (mirror the dedicated-host guard, which already asserts its own digest pin)
- [ ] Verified inert: web-1 is not replaced by the change landing

Ref #7462 · Ref #4699 (adjacent: the drift guard compares tag *names*, not content or digests)

PR #7954, opened 2026-09-08
fix(7500): redact zot_last_err at the producer and the sink — and close the public egress the issue never named

### LINKED ISSUE #7500: arch(registry): decide whether zot_last_err should route through redact() on the SOLEUR_ZOT_DISK path
Referred out of the legal domain by the CLO counsel-review gate on PR #7444 (audit: `knowledge-base/legal/audits/2026-08-counsel-review-7440.md`, §3 / blocker B2). The CLO's own gate is discharged by accurate disclosure either way — this is an architecture call, not a compliance one.

## The path

`soleur-registry`'s pre-existing `SOLEUR_ZOT_DISK` reporter (#6122/#6244) `curl`-POSTs directly to Better Stack Logs source 2457081, with no Vector agent in the path. Its payload carries `zot_last_err`: a ≤300-character sample of zot's own log output, selected in tiers —

1. `panic:` / `fatal error` / signal / `runtime error`
2. `level:error|fatal`
3. a `cannot|failed to|unable to` sweep
4. a `docker logs --tail 3` fallback

That sample passes through the payload-integrity sanitizer **only**, and through **no `redact()` at all**.

## Why it matters

On tier 4 the sample is routinely an `info`-level `HTTP API` row carrying the full `headers` object. On the pinned zot image, zot self-masks `Authorization` — but it does **not** mask `Cookie`, `X-Api-Key`, or arbitrary headers. So the header-object allowlist built for the zot-log-shipper (`redact()`, fail-closed, `jq walk`, case-folded keys) does not protect this sibling path, which reaches the same warehouse source.

This is **pre-existing** — the log-shipper work did not create it, and that emitter is verified sound (150/150, allowlist not denylist, fails closed). It surfaced only because the Art. 30 amendment initially wrote a *host*-level assurance ("its own `redact()` is the whole boundary") that is true of the emitter and false of the host. PR #7444 corrects the disclosure to be emitter-scoped and records this path as a known, bounded residual.

## The decision

**Should `zot_last_err` route through the same `redact()` the zot-log-shipper uses?**

Arguments to weigh, not a foregone conclusion:

- **For:** one host, one warehouse source, two redaction postures is exactly the asymmetry that produces a leak nobody predicted. `redact()` already exists on the host and is tested.
- **Against:** `redact()` is jq-based and `SOLEUR_ZOT_DISK` is a fixed-schema `key=value` reporter with a different failure model. Routing a diagnostic sample through a fail-closed redactor means a row that cannot be redacted is *dropped* — and on a disk-pressure reporter that may destroy the signal you most need.
- **Third option:** a narrower scrub on `zot_last_err` specifically (header-shaped substrings), rather than the full structural redactor.

Both `user_data` cost and the ForceNew cap constrain any change here (ADR-185: 8,000 B headroom, derived from the owning budget constant) — the registry payload is currently 13,136 B stored.

## Bounding context

`hcloud_firewall.registry` carries zero inbound rules; ingress is intra-`10.0.1.0/24` plus a Cloudflare tunnel whose origin is `tcp://10.0.1.30:5000`. Any change admitting public ingress raises the severity of this issue **and** converts `clientIP` into Art. 4(1) personal data on a path with no redaction — recorded as the strongest re-evaluation trigger in the audit frontmatter.

Refs #7444

deruelle:
## Summary

#7500 asked whether `zot_last_err` should route through `redact()` on the `SOLEUR_ZOT_DISK` path. The answer is **yes, at two layers** — recorded in **ADR-211** (`status: adopting`), which is the primary deliverable per `wg-architecture-decision-is-a-plan-deliverable`.

The work also found a **live public egress the issue never mentions, and it is the worse one.** The restart-loop alarm splices a container's own log tail into a comment on a **public** GitHub issue. Measured on #7272: 100 comments, 36 carrying a zot `headers` object, 13 carrying a `clientIP`. A live Better Stack query the same day showed **18 of 20 rows shipping headers+clientIP every 5 minutes while naming a cause 0 times**. Nothing sensitive has leaked — but the bound was a vendor default (zot masks `Authorization` upstream) plus a private ingress, neither of which this repo controls.

**Two layers, because they fail independently:**

- **Layer 1 — producer** (`cloud-init-registry.yml`): `redact()` applied per line, with a tier gate that suppresses tier 4 (the tier that produced ~100% of measured header exposure while naming a cause zero times). **Inert until the next `registry-host-replace`** — the host is cloud-init-only (ADR-096), so merging applies nothing to it. Delivery is watched by #7960, not assumed.
- **Layer 2 — sink** (`zot-restart-loop-alarm.sh`): `scrub_public()` on every published field, via a single `emit_field` chokepoint. **Live at merge**, and it is what covers the public egress in the meantime.

## Plan and brand-survival framing

Plan (archived at compound time): `knowledge-base/project/plans/archive/20260908-215821-2026-09-08-arch-zot-last-err-redact-decision-plan.md` — it carries the `## User-Brand Impact` section (threshold: `single-user incident`, ruled by the CPO) and the `## Observability` block, including a `discoverability_test` that runs unauthenticated and was measured green inside the preflight sandbox.

## The defect class this PR is really about

Four consumers of a `SOLEUR_ZOT_DISK` row parsed **trusted fields out of the untrusted region** — a leading greedy `.*` binds to the *last* occurrence, and the attacker-influenceable `zot_last_err` tail is emitted last (`user-agent` is on the allowlist and preserved verbatim). Forgeable that way: the tier label built to enforce ADR-166, the `REDACTION_FAILED` sentinel, a follow-through's `boot_id` (a forged value **closes the tracker**), and `NIC_ALARM_VERDICT` (a forged `GREEN` **silences the alarm**). This repo already had the discipline in `zot_trusted_region()`; none of the four used it. All four are fixed and fixtured.

A **P0 was introduced and caught inside this PR**: the fix hardening the verdict extraction doubled its backslashes inside a YAML `run: |` block, which applies no escape processing — so both verdicts extracted **empty on every run** and all four filing conditions went false. That is the #7242 "dark on exactly the verdicts it exists to raise" defect, reintroduced by its own remedy. Found by a reviewer re-reading the uncommitted tree; both forms are now measured (single → `FIRE`, doubled → empty).

### Registry host (producer, inert until replace)
- `redact()` applied per line to `zot_last_err`, degrading only that field rather than dropping all 27
- Tier gate suppresses tier 4, emitting `zot_last_err_src=suppressed` so the withholding is visible
- `redact()` now **fails closed without jq** — previously the allowlist silently degraded to a denylist and an unanticipated credential header shipped in the clear at rc=0

### Restart-loop alarm (sink, live at merge)
- `scrub_public()` on every published field through one `emit_field` primitive
- Trusted-region parsing for `boot_id`, `exit_code`, `zot_last_err_src` and the `REDACTION_FAILED` sentinel
- The `REDACTION_FAILED` branch was **dead code** (exact `==` never matched past envelope residue); now prefix-matched and driven

### Records
- **ADR-211** minted (`status: adopting`), recording the declined terraform-interpolation option and both design positions
- ADR-184 scope marker; Article 30 register amended additively (5 rows, incl. Art. 30(1)(d) Recipients); C4 `publicReader` actor

### Guards
- 2 new suites (scrub 31/31, heartbeat redaction 33/33); boot-guard grew 12 drift assertions to 102/102
- Follow-through probe with a real FAIL branch — its baseline is now **baked in**, because `env -i` stripped the env var and made the delivered-and-leaking verdict unreachable

## Test plan

- [x] `zot-restart-loop-alarm-scrub.test.sh` — 31/31
- [x] `zot-disk-heartbeat-redaction.test.sh` — 33/33 (14 measured RED before the fix)
- [x] `registry-boot-guard.test.sh` — 102/102
- [x] `zot-log-shipper.test.sh` — 174/174
- [x] `sweep-followthroughs.test.sh` — 55/55
- [x] `guard-vacuity-floor.test.sh`, c4 parity, infra templates, legal registers — green
- [x] shellcheck clean; `user_data` 14,180 B against the 20,000 B budget
- [x] Follow-through probe driven both ways against live data: baked baseline → TRANSIENT rc=2; differing baseline + leaking rows → FAIL rc=1
- [x] Verdict extraction measured in the shape the runtime receives it: single backslash → `FIRE`, doubled → empty

Closes #7500
Filed: #7959 #7960

Net is +1 (closes 1, files 2). Justification per filed issue:

- **#7959** — a capability gap in a **different subsystem**, not a defect in this PR: no gate anywhere covers runtime publication of third-party or system output to a public artifact by agent-authored automation. This PR fixes the one instance; the class is ungoverned and must stay its own issue (it also subsumes #7844, a second surface). Folding a new workflow lint into a redaction PR is exactly the scope creep the Guard Contract exists to expose.
- **#7960** — the delivery watch for Layer 1, which is **inert by construction** until a `registry-host-replace` fires. The failure mode for an inert change is silence, so the interval is watched by an enrolled follow-through probe rather than assumed. It cannot be "fixed inline" — there is nothing to fix until a host replace happens.


deruelle:
**Pre-existing non-required red: `deploy-script-tests` — inngest bootstrap pin drift. Not caused by this PR.**

`cloud-init-inngest-bootstrap.test.sh` AC6 reports the cloud-init pin `v1.1.27` against a newly published `v1.1.28`. Recording the evidence, because this job is the non-required check that covers `apps/web-platform/infra/` — the directory this PR changes most — so waving it through on "not required" alone would be exactly the gap `ship/SKILL.md` warns about.

Measured:

- The pin lives in `cloud-init.yml` and `cloud-init-inngest.yml`. **This PR touches neither** — its only cloud-init file is `cloud-init-registry.yml`, and `git diff origin/main...HEAD -- apps/web-platform/infra/` changes **zero** pin lines.
- Both files are **byte-identical between `origin/main` and this branch** (md5 `5b1c143b…` and `98b87469…` on each side), so the guard fails the same way on main.
- Main's own recent `infra-validation` runs were green (`5c8f42a3`, `8c3c2336`), which dates the drift to a publish that landed after them — time-dependent, not diff-dependent.

Root cause is already tracked at **#7632** (the host pins the bootstrap image by mutable tag). The recurring bump chore is the #6286 class. Deliberately not bumping the pin here: it is an unrelated infra change in a redaction PR, and it would put a fleet-wide image change in a diff no one reviewed for that.

Every other check is green, and the two failures that *were* mine (a missing xtrace refusal in this PR's own follow-through probe, and the `fixture-relative-assert` baseline row for its new test file) are fixed in `4ab105e16`.

Issue #7957, opened 2026-09-08
work/SKILL.md: the wrapper-as-guard paragraph's notification clause is false, and 12 pre-existing markdownlint errors block editing the file


deruelle:
### 1. `work/SKILL.md`'s wrapper-as-guard paragraph contains a measurably false clause

The paragraph added by #7828 — *"THE WRAPPER THAT REPORTS A GATE IS ITSELF A GUARD"* — says:

> never take a verdict from a PARTIAL artifact (**a backgrounded command emits a completion notification precisely so you need not infer from a tail**, and a per-step line is not the runner's verdict)

The bolded clause is false, and it caused a recurrence of the exact class the paragraph exists to prevent.

**A background task's exit code is the LAST command in the backgrounded string.** A trailing convenience line becomes the verdict. Measured three times in one session (PR #7912):

```
<task-notification> ... completed (exit code 0)
$ grep COMMIT_RC= /var/tmp/c5.log
COMMIT_RC=1
```

The backgrounded string was `git commit … > log; echo "COMMIT_RC=$?" >> log; git log --oneline -1`. The notification reported `git log`'s status. A fourth run was killed mid-`tsc` by the memory reaper — log ends in the bare word `Terminated` — and the notification also said `completed`, i.e. could-not-measure rendered as measured-good, which is the parent class exactly.

I caught it only because I had *also* applied the paragraph's other rule (`echo "RC=$?" >> log` and read that). Anyone following the notification clause alone would have pushed an uncommitted tree while reporting a green gate.

**Proposed correction** (the exact diff, ready to apply — saved from the session):

```diff
-Three rules: never take a verdict from a PARTIAL artifact (a backgrounded command emits a completion notification precisely so you need not infer from a tail, and a per-step line is not the runner's verdict); never pipe a command whose exit code IS the result
+Three rules: never take a verdict from a PARTIAL artifact, and never from the completion NOTIFICATION either — its exit code is the LAST command in the backgrounded string, so a trailing `git log` for convenience becomes the verdict (measured #7912: three commits notified `exit code 0` while `git commit` returned 1, and a gate killed mid-`tsc` notified `completed`). Write the real code yourself (`echo "RC=$?" >> log`) and read that; a per-step line is not the runner's verdict either. Never pipe a command whose exit code IS the result
```

And, in the same paragraph's `**Why:**`, one addition — #7828's two-writers-on-one-index error recurred in #7912 by a **different route**, which does not follow from the original wording:

```diff
[...]

Issue #7935, opened 2026-09-08
fix(kb): knowledge-base/INDEX.md has no merge driver — it conflicts on every sync and silently drops entries on resolve


deruelle:
`knowledge-base/INDEX.md` is generated by `scripts/generate-kb-index.sh` and committed. Its own
header says `Auto-generated ... Do not edit manually`. There is **no `.gitattributes` file anywhere
in the repo** (verified: `git ls-tree -r main --name-only | grep -i gitattributes` returns nothing),
so git merges it line-by-line like ordinary prose.

## Why this is a correctness bug, not only latency

Every branch that adds a KB file appends a row to `INDEX.md`, so every such branch conflicts with
every other one. The reflex resolution is `git checkout --theirs knowledge-base/INDEX.md`, which
takes `main`'s copy — and `main`'s copy **never contains the entries your branch adds**. The
resolve therefore succeeds, CI goes green, and the branch's own index rows are gone.

Observed on PR #7896: this **silently dropped the ADR-206 index entry three separate times**. It
was caught only because an explicit `grep -c 'ADR-206' knowledge-base/INDEX.md` was run after each
resolve. Nothing in the pipeline asserts that invariant, so the same resolve on a branch without
that manual grep ships a KB file that is unreachable from the index.

## Latency cost

`INDEX.md` conflicted on **every** sync during PR #7896 (4+ times in one session). Each conflict
forces a merge commit, and each merge commit restarts the whole CI cycle. `test-scripts` alone is
a measured 27 min on the most recent `main` push (run 34163673236) against a 3 min next-longest
job, so each avoidable conflict costs roughly a full cycle.

## Proposed fix

Create `.gitattributes` at the repo root registering a merge driver for the generated index that
resolves by **regenerating** rather than merging — i.e. the driver runs
`scripts/generate-kb-index.sh` against the merged tree and takes its output, which is correct by
construction because the generator is a pure function of the KB file set.

Acceptance:

- Two branches that each add a distinct KB file, merged together, produce an `INDEX.md`
  containing **both** entries. This must be verified by actually constructing the merge, not by
  inspecting the driver definition.
- Decide and record whether `knowledge-base/kb-tags.txt` and `knowledge-base/kb-categories.txt`
  need the same treatment. Both are emitted by the same generator (`generate-kb-index.sh:29-30`,
  facet extraction at :144) and both are committed, so they are the same defect class. Note there
  is also a mirrored copy under `plugins/soleur/knowledge-base/`.
[...]

Issue #7927, opened 2026-09-07
markdown-lint: main is red on 49 errors, and only the local pre-commit hook can see it


deruelle:
`main` currently fails the repo's own `markdown-lint` pre-commit hook on **49 errors
across ~10 files**. Nothing in CI catches this, so it accumulated silently and now
blocks a specific, common operation.

## Why it was invisible

`markdownlint` exists **only** as a lefthook `pre-commit` command
(`lefthook.yml`, `markdown-lint`, `npx --yes markdownlint-cli {staged_files}`).
`grep -rln markdownlint .github/workflows/` returns nothing. PRs land on `main` by
GitHub **squash merge**, where lefthook never runs — so a red file reaches `main`
whenever its PR was merged without a local commit that staged it.

## Why it matters now

The hook lints `{staged_files}`. A **merge commit** stages every file the merge
brings in, so any contributor who runs `git merge origin/main` locally — the
documented `/ship` Phase 6.5 conflict path — is blocked by errors they did not
write. Measured on PR #7894: 68 staged `.md`, 49 errors, **zero** attributable to
that branch (`git diff --quiet origin/main -- <file>` is clean for every offender).

## The errors

| File | Rule | Note |
|---|---|---|
| `plugins/soleur/skills/work/SKILL.md` | MD038 x9, MD052 x1 | see root cause below |
| `docs/legal/{privacy-policy,gdpr-policy,data-protection-disclosure}.md` | MD034 | bare `ops@soleur.ai` / `legal@jikigai.com` — published legal texts, so the fix is a content decision, not a lint decision |
| `docs/legal/{individual,corporate}-cla.md` | MD032 | list inside a blockquote needs surrounding blanks |
| `apps/web-platform/infra/sentry/README.md` | MD012 | double blank line |

## Root cause of the `work/SKILL.md` cluster (diagnosed, not guessed)

Seven of its ten errors trace to **one** construct. Line 688 writes a code span with
backslash-escaped backticks inside it:

```text
`.select(\`…\`)`
```

CommonMark does not honour a backslash escape as a code-span delimiter, so the span
closes at the first escaped tick. Every later backtick on that long line then pairs
one position off, which is why `markdownlint` reports spans like `, assert no ` and
why `[i][0]` lands outside a span and trips MD052. The line renders wrong today. The
fix is a longer delimiter run and no backslashes. Line 639 carries the identical
defect (a `grep` for a literal backtick inside a single-tick span).

[...]

Issue #7837, opened 2026-09-04
39 pre-existing markdownlint errors block any commit that stages the legal docs or expenses.md


deruelle:
## What

`markdownlint-cli` reports **39 errors across 8 files** that are already present on `main`. They are invisible day to day because lefthook's `markdown-lint` step runs against `{staged_files}` only — so they surface the moment a PR happens to stage one of these files together, and then block an unrelated commit.

Measured 2026-09-04 by linting the `HEAD` blobs of the affected files with the repo's own `.markdownlint.json`:

| count | rule | where |
|---|---|---|
| 24 | `MD034/no-bare-urls` | `docs/legal/privacy-policy.md`, `docs/legal/data-protection-disclosure.md` + both `plugins/soleur/docs/pages/legal/` mirrors, `knowledge-base/legal/article-30-register.md`, `knowledge-base/legal/compliance-posture.md`, `knowledge-base/operations/expenses.md` |
| 5 | `MD055/table-pipe-style` | `knowledge-base/operations/expenses.md` (4), `knowledge-base/legal/article-30-register.md` (1) |
| 4 | `MD038/no-space-in-code` | `knowledge-base/legal/compliance-posture.md` (3), `knowledge-base/legal/article-30-register.md` (1) |
| 2 | `MD050/strong-style` | `knowledge-base/legal/article-30-register.md` |
| 2 | `MD012/no-multiple-blanks` | `apps/web-platform/infra/sentry/README.md`, `knowledge-base/legal/compliance-posture.md` |
| 1 | `MD032/blanks-around-lists` | `knowledge-base/engineering/operations/runbooks/betterstack-log-query.md` |
| 1 | `MD018/no-missing-space-atx` | `knowledge-base/engineering/architecture/decisions/ADR-198-*.md` |

Most of the `MD034` hits are contact addresses (`ops@soleur.ai`, `legal@jikigai.com`) that need autolink brackets.

## How it was found

PR #7805 (#7772, git-data pre-birth hardening) stages several of these files. Its `pre-commit` `markdown-lint` step failed with 42 errors. Diffing the working tree against the `HEAD` blobs of the same files under the same config isolated **3 new** (introduced by that PR, all fixed there: an `MD012` in the ADR addendum, an `MD032` at a task-annotation boundary, an `MD037` from an appended bold opener) from **39 pre-existing**. After the three fixes the count is exactly the HEAD baseline.

## Why it was not fixed in that PR

[...]

PR #7962, opened 2026-09-08
chore(ship): batch commits before --auto, and trigger the settle-then-admin-merge hatch at sync 2

### LINKED ISSUE #7937: chore(ship): two merge-latency rules — batch commits before queueing auto-merge, and give the settle-then-admin-merge hatch a concrete early trigger
Two convergent ship-discipline gaps, both measured on PR #7896 (merged 2026-09-07, 21478828),
which burned ~6 full CI cycles. Both fixes live in `plugins/soleur/skills/ship/SKILL.md`. Bounded
surface: one bullet or one <=3-line clarification each — this is not a restructuring of the skill.

Context that makes both load-bearing: CI cycle time is gated almost entirely by one job.
Re-verified on the most recent `main` push (run 34163673236): `test-scripts` = 27 min, next-longest
job = 3 min, then 3m / 3m / 2m / 1m. Every sync or commit restarts that whole cycle. `ship/SKILL.md`
already carries this figure in its Phase 7 poll rationale ("`test-scripts` alone is a median 28").

## (a) Batch every commit BEFORE queueing `gh pr merge --auto`

On #7896 a correct-but-late ADR commit landed when CI was already at 65/67 checks. Pushing it reset
the head ref and cost a **full 27-35 minute cycle** for a change that was known-needed earlier in
the session.

Phase 6 should state that late verification work belongs in the **review-agent wait**, not after
the PR is marked ready: any commit you can foresee needing must be in the tree before
`gh pr merge --auto` is queued. The review-agent wait is dead time that is already being spent, so
moving the work there is free; moving it after `--auto` costs a cycle at the worst possible moment,
when the PR is one check from merging.

## (b) Give the settle-then-admin-merge hatch a concrete trigger

The hatch is already documented in `ship/SKILL.md` under **"Settle-then-admin-merge escape hatch
(zero-conflict-surface changes only)"**, beneath "Auto-sync on BEHIND". Today it is reachable only
at the `MAX_BEHIND_SYNCS` cap — the poll loop's `behind_exhausted` branch recommends it after
**6** auto-syncs — and otherwise left to judgment. #7896 rode the normal path through ~4 cycles
before anyone reached for it.

Two things are wrong with the current trigger:

1. **The stated premise is stale.** The hatch's rationale reads "when `main` is merging PRs faster
   than this PR's **~8-minute CI cycle**". The cycle is now 27-35 minutes. At that duration against
   a busy `main`, the livelock the hatch exists to break is close to **structural**, not
   exceptional — every sync push re-triggers the full required-check set and `main` moves again
   before it settles.
2. **Six syncs is far past the point of diagnosis.** The livelock is identifiable much earlier.

Proposal: give the hatch a concrete early trigger — e.g. **2 consecutive DIRTY/BEHIND syncs on a
change whose only conflict is a generated file** — instead of leaving it to judgment or waiting for
the 6-sync cap. A generated-file-only conflict surface is exactly the "zero conflict surface"
precondition the hatch already requires, so the trigger does not widen the hatch's scope; it makes
the existing precondition detectable at sync 2 rather than sync 6.

Note the interaction with #7935 (the `INDEX.md` merge driver): once generated-file conflicts stop
happening, this trigger fires less often — but it should still exist, because the trigger names a
*class* (generated-file-only conflict) that is not exhausted by that one file.

## Constraints

- Bounded edit. One bullet or one <=3-line clarification per item; do not restructure Phase 6 or
  the Phase 7 poll block.
- The Phase 7 poll block is fenced (`<!-- phase-7-poll-block:start -->`) and **mirrored** in
  `plugins/soleur/skills/merge-pr/SKILL.md` §5.2, with a fixture at
  `plugins/soleur/test/ship-phase-7-poll-fixtures.test.sh` whose awk extractor anchors on that
  fence and a variable-set fingerprint. If (b) touches anything inside the fence, all three must
  move together.
- `plugins/soleur/test/fullsuite-merge-gate.test.ts` slices ship Phase 4 between the literal
  headings `## Phase 4: Run Tests` and `\n## Phase 5: Final Checklist`, and asserts the literals
  `Then run the project's full test suite.` and `no required status check runs that shard`. Edits
  near those anchors break the suite loudly — that is intended, but keep this change clear of them.
- Update the stale "~8-minute CI cycle" figure to the measured one while in the file, with its
  derivation named so the next reader can re-derive rather than trust it.

Evidence re-verified on `main` @ 214788282 / CI run 34163673236 (2026-09-07).

deruelle:
Closes #7937

Two ship-discipline rules, both measured on PR #7896 (merged 2026-09-07, `21478828`), which burned ~6 full CI cycles.

**Scope, stated honestly: (b) overran the issue's bound.** #7937 asks for "one bullet or one ≤3-line clarification each". Hunk (a) is one bold-lead paragraph and complies. Hunk (b) is not one clarification — it is a rewritten paragraph, a fenced derivation command, a trigger paragraph and two correction paragraphs. The issue's *hard* constraint held (both edits sit outside `<!-- phase-7-poll-block:start/end -->`; the mirror and its fixture are untouched, 16 pass / 0 fail); the soft one did not, and the overrun is almost entirely review findings — the trigger had to be re-keyed, and the hatch's central safety claim turned out to be false.

## (a) Pushed is not the same as finished

Phase 6.4 already asks whether what you have is **pushed**. It cannot ask whether what you have is **all you will need**, and that is the gap #7896 fell through: a correct-but-late ADR commit landed at 65 of 67 checks, reset the head ref, and restarted the entire required-check set at the moment the PR was one check from merging.

The new bullet says any foreseeable commit belongs in the tree *before* `gh pr merge --squash --auto` is queued, and names the review-agent wait as where that work goes — dead time already being spent, so moving it there is free.

## (b) The hatch's premise was stale by 4×, and its trigger was three hours late

The settle-then-admin-merge hatch was reachable only at the 6-sync cap, and its rationale read *"main is merging PRs faster than this PR's ~8-minute CI cycle."*

**Measured rather than inherited, because the issue's own figure was already a day old.** Over the five most recent *completed* `main` CI runs on 2026-09-08, `test-scripts` took **34 / 36 / 36 / 35 / 36 min** while the next-longest job took 4 min (7 min once). The file claimed ~8; #7937 cited 27 from run `34163673236` (2026-09-07). Both are wrong now — which is exactly why the **derivation command** is written into the file instead of another bare number, and why the text tells the next reader to re-derive: #7907 shards this job, so the figure moves again on its merge.

At a ~35-minute cycle the livelock the hatch exists to break is close to **structural** rather than exceptional, and six syncs is roughly **three hours** — far past the point of diagnosis. The new trigger is **2 consecutive DIRTY/BEHIND syncs whose only conflict surface is a generated file**.

That does not widen the hatch. A generated-file-only conflict surface already *is* the "zero conflict surface" precondition the hatch requires, so reading it at sync 2 only makes the existing precondition detectable before the budget is spent. It names a **class**, so #7935's merge driver removes `knowledge-base/INDEX.md` from it without emptying it.

## Blast radius

Both edits sit **outside** `<!-- phase-7-poll-block:start/end -->`, so the mirror in `merge-pr/SKILL.md` §5.2 and its fixture are untouched — verified, 16 pass / 0 fail. Both are clear of the literals `fullsuite-merge-gate.test.ts` anchors on (10 pass / 0 fail). `components.test.ts`: 1297 pass / 0 fail.

## Eleven of the thirteen hunks are pre-existing lint, not scope creep

lefthook lints **staged** files, and `ship/SKILL.md` carries **16 markdownlint errors on `main` today** — measured identical on `origin/main` and on this branch before any of my edits. The file cannot be committed at all without clearing them, so clearing them is the cost of touching it.

- **3 × MD038** — two were the escaped-backtick-inside-a-code-span construct (markdown has no backslash escape there, so the spans mispair); rewritten as a double-backtick fence. One carried a trailing space *inside* the span to show a regex has one; moved into prose so the meaning survives the fix.
- **3 × MD031, 1 × MD032** — fences and a list given their blank lines.
- **9 × MD007** — scoped out with a comment rather than "fixed". Those bullets are continuation content aligned to a numbered step's own 3-space indent; MD007 derives the expected indent from list *nesting depth*, so it reports 2 at one site and 0 at the other, and satisfying either would misalign them from the block they belong to.

**One correction to an earlier version of this description.** It said "2 × MD032", summing to 17 against a stated 16. The second MD032 was **self-induced**: the `markdownlint-disable` comments were at column 0 inside 3-space continuation content, and a column-0 HTML block terminates the enclosing list item in CommonMark — so the comment protecting the nesting silently destroyed it, and the resulting blank-line error read as somebody else's debt. Indenting the comments into the list item removes both the de-nesting and the induced error. The partition is 16 = 9 scoped out + 7 repaired.

**This is #7927, not a one-off.** That issue records `main` red on 49 markdownlint errors visible only to the local pre-commit hook; #7817, #7832, #7837 and #7957 are four more per-file instances of the same block. `ship/SKILL.md` is the next one, now cleared — nothing is filed here.

The drift is datable rather than guessed. The escaped-backtick MD038 line was introduced **2026-04-22** (`9113c559d`) and the MD007 bullets **2026-07-11** (`09e9a3e82`); `ship/SKILL.md` has been staged and committed **12 times in the last 30 days** with both present, passing the same hook each time. The file did not change in those regions, so the linter's verdict did — consistent with lefthook running `npx --yes markdownlint-cli`, which resolves the **latest** version on every invocation. Pinning it is the real fix and belongs to #7927.

## Review

Four agents (non-code class: git-history, pattern-recognition, security-sentinel, code-quality). Every blocking finding was re-derived before acting on it, and every one held. The three that changed the deliverable:

1. **The trigger keyed on something the loop cannot print.** The only BEHIND sync that reaches 2 is a *clean* one — a conflicting `git merge origin/main` aborts and breaks on the first occurrence — so there is no conflict surface to classify when the trigger fires. Re-keyed on the branch's own diff, with the classifier command inline and verified in both directions.
2. **`DIRTY` had to go.** The poll block evaluates DIRTY before BEHIND and `break`s on the first occurrence, so two consecutive DIRTY syncs are unobservable; and `--admin` bypasses branch protection, never an actual conflict, so the hatch cannot execute on a DIRTY PR at all. #7937 proposed `DIRTY/BEHIND`; that half of the proposal is wrong.
3. **The hatch's safety claim was false, and this PR was making it reachable three times earlier.** `--admin` bypasses the entire `required_status_checks` rule — all 22 contexts, not only the up-to-date gate — because the strict policy and the contexts are sibling parameters of one rule and ruleset bypass is granted per rule. Corrected at both sites rather than propagated.

Also fixed from review: `--slurp` is rejected by `gh` alongside `--jq` (my first fix didn't run); `$GH_REPO` was unset in every local shell, so the derivation snippet 404'd silently; step 1 asserted a state that does not exist at sync 2; "six syncs is roughly three hours" is impossible under `MAX_POLL_MIN=60`; `merge-pr` §5.2's prose still advertised the 6-sync cap; and the poll block's own `median 28` figure now carries a cross-reference instead of contradicting the new one in silence.


deruelle:
Heads-up: this branch now conflicts with `main` on `plugins/soleur/skills/ship/SKILL.md`, and the cause is mine.

PR #7955 (the repo-wide markdownlint sweep, merged as `ae44051a8`) was explicitly scoped to **exclude** this file precisely so it could not fight with #7962. I never added the exclusion to `.markdownlintignore`, so the sweep reached it. Verified: `git merge-tree origin/main <this branch>` now reports `CONFLICT (content) in plugins/soleur/skills/ship/SKILL.md`.

**The conflict is cheap to resolve — nothing semantic changed on the `main` side.** Measured on the merge commit:

- 4 blank lines added (blanks-around-headings / blanks-around-fences)
- 1 `<!-- markdownlint-disable-line MD038 -->` comment added
- Alphanumeric prose token delta is exactly 4 tokens, all of them inside that HTML comment: `markdownlint disable line MD038`. No rendered text changed.

**Suggested resolution:** take *your* side wholesale for this file, then run:

```bash
bash scripts/markdown-lint.sh plugins/soleur/skills/ship/SKILL.md
```

That reproduces whatever formatting `main` now expects, and it is the same binary CI runs — `markdown-lint` became a **required** status check on ruleset 14145388 as of that merge, so the file must be lint-clean to merge regardless of how the conflict is resolved.

Two notes that may save you time:

- The linter is now pinned exactly (`markdownlint-cli` 0.49.1) and `scripts/markdown-lint.sh` refuses to run against anything else, so the verdict no longer drifts between runs — which was the root cause #7962 and #7927 independently diagnosed.
- If you were relying on this file's 16 pre-existing errors as your work item, they are already cleared on `main`; your PR's remaining value is the discipline-rule content, not the lint fix.

Apologies for the collision — the exclusion was specified and I did not implement it.

Ref #7955

Issue #7710, opened 2026-08-26
gdpr-gate refuses before scanning: 'rules 108 days stale' / POSTURE_FAIL


deruelle:
`scripts/gdpr-gate.sh` refuses before scanning anything, on a pre-existing repo condition unrelated to any one branch:

```
rules 108 days stale
POSTURE_FAIL
```

Surfaced while running the compliance gate during #7652's planning phase. It is not caused by that branch and has no open tracker, so it is filed on its own rather than absorbed into an unrelated PR.

## Why it matters

The gate is invoked at `plan` Phase 2.7 and `work` Phase 2 exit. In its current state every invocation returns the same refusal, which means:

- no diff is actually scanned for GDPR/CCPA/HIPAA exposure at either gate;
- the refusal is uniform, so it reads as background noise rather than as a finding, and gets skipped past;
- a genuine regulated-data regression would produce **the same output** as today's healthy-but-stale state.

That last point is the reason to fix it rather than raise the staleness window: a gate whose failing output is indistinguishable from its passing output is not a gate.

## What to check

Whether the rule set needs a refresh, or whether the staleness threshold itself is the wrong instrument — a rule corpus that is 108 days old is not necessarily wrong, and a date-based refusal that fires on correct rules trains the reader to ignore it.

Mandated-By: wg-when-deferring-a-capability-create-a

**Source:** PR #7702 (filed as deferred scope during that work).

deruelle:
**Automated Triage**

Priority: p2-medium
Type: bug
Domain: engineering

Reasoning: The GDPR compliance gate (`scripts/gdpr-gate.sh`) is non-functional — it refuses before scanning and produces identical output for healthy vs broken states. This is a compliance tooling regression: regulated-data regressions now go undetected at both `plan` Phase 2.7 and `work` Phase 2 exit. However, it does not affect production directly and is filed on a Post-MVP milestone, so medium priority is appropriate. The gate needs a fix rather than a window increase, but this is not urgent enough to block current work.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #7255, opened 2026-08-04
gdpr-gate: cmd_cron_run_stale returns 999 unconditionally — its workflow became an Inngest cron


deruelle:
`cmd_cron_run_stale` in `plugins/soleur/skills/gdpr-gate/scripts/notice-frontmatter.sh`
returns **999 on every invocation**, and has since the content-vendor-drift job moved off
GitHub Actions.

## What it does

The function reports "days since the content-vendor-drift check last succeeded". The caller
uses that number to decide whether to show the operator-attested-mode banner. It resolves
the number with:

```bash
gh run list --workflow=scheduled-content-vendor-drift.yml --status=success --limit=1 ...
```

`scheduled-content-vendor-drift.yml` **does not exist**. The job is now an Inngest cron:
`apps/web-platform/server/inngest/functions/cron-content-vendor-drift.ts`. An Inngest
function produces no GitHub Actions run, so the query matches nothing, the empty result
falls through the `// empty` guard, and the function returns its 999 sentinel
unconditionally.

## Why it went unnoticed

999 is the **fail-safe** direction. It forces the operator-attested-mode banner rather than
falsely asserting the check is fresh, so the visible behaviour is conservative and looks
deliberate. The function's own comment said only that "renaming the workflow silently breaks
this binding" — accurate about the mechanism, and it did not say the binding was already
broken.

## Why it was not fixed in #7234

Found by the Phase-2.7 `gdpr-gate` run while planning #7234, and initially scoped there as a
one-line workflow-name correction. It is not one line: **repointing at a different workflow
filename cannot work when the job is not a workflow at all.** A real fix needs a liveness
source that can see an Inngest cron — the Inngest API, a Sentry cron monitor check-in, or a
`SOLEUR_*` marker the function emits that Better Stack already ingests — which is a
different subsystem from the credential work #7234 does.

#7234 corrected the false reference in place (so the next reader is not hunting a workflow
that does not exist) and left the behaviour **exactly** as-is, since 999 fails safe.

## Done when

- [ ] `cmd_cron_run_stale` resolves a real timestamp for the Inngest-hosted
      content-vendor-drift check, or the frontmatter field is retired if the banner no
      longer needs it.
- [ ] A test pins the non-999 path, so a future move of the job reds instead of silently
      reverting to the sentinel. The absence of such a test is why this survived the move.
- [ ] The `⚠️` warning block added by #7234 is removed once the binding actually resolves.

[...]

PR #7984, opened 2026-09-09
chore(infra): the documented Terraform state rollback cannot run — repair the runbook, ADR-006, and the Art. 30 register

### LINKED ISSUE #7836: R2 has no object versioning: the documented Terraform state rollback is inoperable for all six roots
## The documented Terraform rollback procedure cannot be executed

`infra/github/README.md` §"Phase 5 — Rollback" step 1 instructs the operator to list prior
state versions in R2, and step 2 restores one by `versionId`. **Step 1 returns
`NotImplemented`**, so step 2 has no input and the whole procedure is inoperable.

This is not specific to `infra/github` — it is a property of the bucket, so **every Terraform
root in this repo has a rollback runbook that cannot run.**

## Measured 2026-09-04

Credentials: `doppler -p soleur -c prd_terraform`. Exit codes captured on their own line, not
through a pipe.

```
# the runbook's own command, verbatim prefix
aws --endpoint-url=https://4d5ba6f096b2686fbdd404167dd4e125.r2.cloudflarestorage.com \
  s3api list-object-versions --bucket soleur-terraform-state --prefix github/terraform.tfstate
  -> rc=254
     An error occurred (NotImplemented) when calling the ListObjectVersions operation:
     ListObjectVersions not implemented

aws ... s3api get-bucket-versioning --bucket soleur-terraform-state
  -> rc=254  (AccessDenied — token scope; ambiguous on its own, which is why the control below matters)

# CONTROL — the same credential, same endpoint, a call R2 does implement
aws ... s3api list-objects-v2 --bucket soleur-terraform-state
  -> rc=0
```

The control is the point: this is **not** an auth or endpoint problem that a wider token would
fix. Cloudflare R2 does not implement the S3 object-versioning API at all.

## Blast radius — all six roots on the bucket

`list-objects-v2` enumerates:

```
cla-evidence/terraform.tfstate
github/terraform.tfstate
telegram-bridge/terraform.tfstate
web-platform/rung2-rehearsal/terraform.tfstate
web-platform/sentry/terraform.tfstate
web-platform/terraform.tfstate
```

None of them has point-in-time state recovery. Any runbook, ADR, or plan that assumes
"restore the prior state version from R2" is assuming a capability that has never existed here.

## Why it surfaced now

#7650 Phase 2 moves 27 live Sentry paging rules — including `byok-art-33-breach`, the anchor
of the GDPR Art. 33 72-hour clock — between Terraform addresses. Its Encryption Posture
section listed "object VERSIONING is unverified" as a known unknown and scheduled the probe.
The probe ran; the answer is that it does not exist. That PR's recovery design is
**roll-forward only** and names state-restore explicitly as a trap, so it is not blocked by
this — but it is the reason the claim finally got checked.

## Done when

One of:

1. **Make the claim true** — enable versioning on the bucket if Cloudflare has since shipped
   it (re-probe first; this issue records a measurement, not a permanent vendor fact), or add
   a scheduled copy of each state object to a versioned store, and update the runbooks to the
   gesture that actually works.
2. **Make the docs true** — delete the Phase 5 rollback steps from `infra/github/README.md`
   and replace them with the roll-forward gesture, then grep every other root's README, ADR,
   and plan for the same assumption and correct each. A runbook that names an impossible
   command is worse than no runbook: it is read during an incident, by someone who will spend
   their first minutes discovering it does not work.

Option 2 is the smaller change and does not depend on a vendor roadmap; option 1 is the one
that restores the capability. They are not mutually exclusive.

## Owner

@deruelle — infrastructure/CTO domain. Raised from #7650 Phase 2 task 7.3, which required a
named owner rather than a PR-body line precisely so this does not become another documented
capability nobody re-checks.

deruelle:
`infra/github/README.md` §"Phase 5 — Rollback" opened by listing prior state versions in R2 and restoring one by `versionId`. Step 1 returns `NotImplemented`, so step 2 had no input: the documented recovery has never been executable. The claim did not originate there — it originates in **ADR-006**, which recorded the decision as *"R2 as remote backend **with bucket versioning**"* and its consequence as *"State loss eliminated via bucket versioning."* Both were false on the day they were written, and over ~18 months the claim propagated into a runbook and into a GDPR Article 30 register as an **Article 32 technical measure**.

This corrects the record. The capability limb — an automatic pre-apply state snapshot — is deferred and tracked at **#7992**.

## Why this is a documentation fix and not a capability fix

The issue's "Done when" offered two non-exclusive options and left the choice open. That is a technical fork, decided here rather than escalated (`hr-technical-fork-is-not-an-operator-question`):

| Option | Verdict |
|---|---|
| Enable R2 bucket versioning | **Refuted by measurement** — re-probed three times; there is no toggle to flip |
| Scheduled copy to a versioned store | **Wrong trigger.** The damage is written *by an apply*, so a schedule's recovery point is the last tick. Read as a property rather than adopted as written, the correct gesture is a **pre-apply** snapshot — which is #7992, not this PR |
| Make the docs true | **Adopted** — discharges the issue, removes a live incident-time hazard, zero apply risk |

The deferral's coherence rests on a measured claim: **none of the four real recovery paths for this root needs a snapshot.** If that stops holding, the deferral stops holding with it — stated in #7992 as its own re-evaluation trigger.

## Presence-with-provenance

Absence sweeps prove the false claim is gone; they never prove the replacement is true. Every claim this diff **adds**:

| Added claim | Source |
|---|---|
| R2 has no point-in-time recovery | Probe pair, re-run at implementation time: `list-object-versions` rc=254 `NotImplemented` vs control `list-objects-v2` rc=0, same credential and endpoint |
| No second recovery substrate | Zero `.backup` objects in the bucket |
| State is not locked; applies serialized by CI | `use_lockfile = false` in all five backend blocks; concurrency groups enumerated per root |
| Serialization is **partial** | sentry's group is ref-keyed; rung2-rehearsal uses a third group; `apps/cla-evidence/infra` has **no apply workflow** |
| Config-revert is the primary recovery | `apply-github-infra.yml` runs `terraform apply -auto-approve` on merge |
| The revert can fail closed | Destroy-guard fixture returns `nested_deletes: 1`; guard at `apply-github-infra.yml:360` requires `[ack-destroy]` |
| A self-lock still has a way through | `bypass_actors` at `ruleset-ci-required.tf:73` (`OrganizationAdmin`, `RepositoryRole 5`) |
| Auto-apply restores **3 of 6** resources | Three `import_*` lines in the workflow; `marketplace_pr_required` (live `20802604`) is not among them |
| The DR script reissues the ruleset id | `create-ci-required-ruleset.sh:107` is `-X POST`; `14145388` is hardcoded in 12+ sites |
| `state push`: equal serial pushes **without** `-force` | **Measured**, Terraform v1.10.5: equal-serial rc=0; lower-serial rc=1 `cannot import state with serial 1 over newer state with serial 7` |
| `state push` has no `-backup` | `state push -help` lists only `-force`, `-lock`, `-lock-timeout`, `-ignore-remote-version` |
| AWS keys must sit outside `--name-transformer tf-var` | The transformer yields only `TF_VAR_aws_*`; matches the shape in `apply-github-infra.yml` |
| PA12 §(g) items (1)(2)(3)(5)(8) were false/stale | App auth per #4384 (`main.tf:33`); `GH_RULESET_PAT` absent from Doppler (rc=1); no `§Rotation` heading; Phase 2 has 0 sub-steps; auto-apply exists |
| 23 required contexts | `grep -c 'context *=' ruleset-ci-required.tf` → 23, and the live ruleset agrees |
| 27 state-only secrets; 18 `terraform_data` | Counted in `apps/web-platform/infra/*.tf` |

## Measured scope correction to the issue

#7836 says *"every Terraform root in this repo has a rollback runbook that cannot run."* Measured: **exactly one root has a rollback section at all.** Four live roots have none and two have no README. Its title also says "all six roots" — the roots are **five**; the bucket holds **six** objects, the sixth an orphan of a root deleted in `dccc56dee` (#1586). That distinction is load-bearing for one thing only: a `-force` push can clobber any of the six.

## GDPR gate

Run, not waived (the plan's earlier waiver was removed — a hard-rule waiver cannot rest on an agent's paraphrase).

```
gdpr-gate: path scan complete — 10 examined, 0 matched
```

No regulated **code** path is touched; the change is document-layer, so substantive review belongs to `clo` / `legal-compliance-auditor` rather than a code-level gate. The run also measured the detection corpus at **122 days** stale (`POSTURE_FAIL`); that is an already-tracked condition — recorded on #7255 rather than filed anew.

## Review

Five seats: legal cross-artifact, runbook operability, ADR coherence, structural enumeration, git-history provenance. **18 findings, all fixed inline, 0 filed.**

**The structural cause of four P1s was one gap:** the recovery taxonomy was written from the *resource list and the plan*, not from the *workflow and DR script that must deliver the recovery* — "the runbook asserts a capability nobody measured against its implementer", the exact defect class this PR exists to remove, reproduced inside the replacement text. Fixed at the class: every automatic-behaviour claim in §Phase 5 is now checked line-by-line against `apply-github-infra.yml` and `create-ci-required-ruleset.sh`.

Findings that were corrections *to my own correction*, worth naming:

- The §(g) note asserted the cited `§Rotation` heading **does not exist**. It does — `## Phase 4 -- Rotation` — and my own item (2) cites it three sentences later. A false statement newly added to a statutory register.
- The Art. 33/34 precedent pointed at the 2026-08-06 row, which **is** a row in `breach-register.md` and is recorded as *contested* — cited to support declining a row. Repointed at the Excluded-records determinations.
- The Art. 4(12) reasoning rested on *"a control that never existed cannot have failed"*, which returns the same answer **even if a state loss had occurred**. Replaced with two grounds that are sensitive to it.
- PA12 §(c)/§(e) still defined the data-subject population by possession of the PAT. This diff is what made §(g) contradict them in the same block.
- I asserted *"no writer is advancing `last-verified`"* on the #7710 row. False — #7841 merged the writer on 2026-09-07. Corrected here and on #7255.

**The sweep proved something narrower than its name.** The enumeration seat established four unmarked live sites remained, two carrying *runnable* commands for a capability that does not exist. Applying this PR's own carve-out criterion consistently — correct live commitments, append to dated records — the unmeetable AC is closed out and two dated records carry a supersede banner.

## Deviations, recorded rather than dropped silently

- **AC21 not done.** Ticking `specs/feat-remove-telegram-bridge/tasks.md` 3.5 is correct on the merits, but that file carries a **pre-existing** `lint-infra-no-human-steps` violation (`ssh root@<ip>` teardown) failing identically on `origin/main`. AC21's rationale was that the tick is free; it is not. The deeper staleness: the decommission completed in `dccc56dee`, yet the spec still carries 45 open tasks for finished work.
- **One unmarked site left unmarked, deliberately.** `plans/2026-05-16-...` still carries the runnable `list-object-versions` recipe this section was transcribed from. Staging it trips a pre-existing violation at its line 908 that also fails on `main`, and both ways through — rewriting a dated record, or wrapping an unrelated finding in a suppression nobody reviews — are worse than the residual.
- **AC10b resolved by editing, not deferring.** The plan carved out `apply-sentry-infra.yml` as owned by PR #7866; that merged 2026-09-08, so the carve-out expired and the falsified citation is corrected in place. Verified comment-only: not hashed into any `triggers_replace`/`filesha256`, YAML parses, and `actionlint` reports the same 14 pre-existing findings as `main`.
- **Three plan-quoted counts were stale and re-derived**: the unsafe credential form appears **6** times across **4** fenced blocks (not "four … two broken, one safe"); the root manages **6** resources (not 3); `errored.tfstate` is zero-hit only once the plan file itself is excluded.
- **A no-op suppression was written and then removed.** An initial `lint-infra-ignore` region on ADR-006 turned out under mutation to suppress nothing — shipping it would have falsely advertised a waived finding.

## Verification

R2 probe pair re-run at implementation time; every command in the rewritten section executed as written (`init` rc=0, `state list` rc=0, `state pull` rc=0 at mode `0600`/dir `0700`, `jq` rc=0, cleanup rc=0, `gh run list` rc=0) with **`terraform state push` documented and deliberately never exercised**; `state push` serial semantics measured in a scratch root; AC1 non-vacuous (208-line section, 43 extracted command lines, 0 violations, 6/6 positive anchors); every `--name-transformer` fence self-contained; `markdownlint`, `lint-infra-no-human-steps`, `lint-legal-registers` (11 assertions), `check-adr-ordinals` all clean; legal-doc suites 41 passed; 7 sentry op-contract suites 56 passed; `gdpr-gate` + repo-scan + `terraform-target-parity` 257 pass / 0 fail; `actionlint` zero introduced.

Not run locally: the full battery. `test-all.sh --capacity` reported `CAPACITY_CONTENDED` (a sibling worktree held a full-gate run), so the suites whose inputs this diff changes were run individually. CI's required `test` context is the merge gate and is running on this SHA.

Closes #7836
Filed: #7992 #8014

## Net issue flow

`NET = FILED(2) - EXEMPT(0) - CLOSING(1) = +1`, declared rather than left for the gate to find.

Neither filing qualifies for the corpus-derived exemption — `wg-when-deferring-a-capability-create-a`
does not carry `[mandates-filing]` (only `wg-block-pr-ready-on-undeferred-operator-steps` does), so
the override is the honest route rather than a `Mandated-By:` trailer that would not match.

| Issue | Why it could not be inlined or consolidated |
|---|---|
| **#7992** — automatic pre-apply state snapshot | The capability limb. Real infrastructure work across four apply workflows with its own retention policy, per-root public-repo disclosure analysis, and an ADR — and it is **cited by number** in both ADR-006's amendment and the Art. 30 register cell, so it had to exist before those documents were written (filed in Phase 0 for exactly that reason). Folding it in would keep an impossible command in the runbook for the duration of the infra work. |
| **#8014** — Doppler absent from the Art. 30 sub-processor mapping | A **discovered defect in a different subsystem**, which the review rules require stay its own issue and PR. It is also a different severity: a US processor with no recorded location, Art. 28(3) instrument or Chapter V mechanism, holding the pepper that decides whether stored hashes are personal data. Filing it as a line item inside an aggregate would have been severity laundering — the CONCUR gate dissented on precisely that. |

Two cheaper instances of the same class were **fixed inline instead of filed** on the dissent's
reading: Cloudflare activity 12 and Resend activity 28, both of which needed no new determination
because the location, instrument and transfer mechanism were already recorded in their own rows.


deruelle:
## Review handoff — design-validity pass on #7836

Ran a design-validity review of this branch from a second session. Partway through I detected that **this worktree is being actively edited right now** (a file written one second before my clock read), so I stopped and am handing findings over rather than editing alongside you. Everything below is verified against the tree, with the command named.

### Housekeeping — two things I did to this branch

1. **Force-pushed the branch** (lease-checked, `9a28e30f3` → `2b82e0949`). The remote was still the 3-commit pre-rebase state, so this PR had never displayed the actual work. Nothing was lost; the push published commits that already existed locally.
2. **Committed one fix, `b24940854`** — please don't revert it as unexplained drift. `compliance-posture.md` moved the #7710 row `IN-PROGRESS` → `OPEN`, arguing the control is unremediated because "no writer has advanced `last-verified`". The writer exists: #7841 (`eeec56893`) merged 2026-09-07T11:39Z and restored it (`cron-content-vendor-drift.ts` carries *"Whether a run may advance NOTICE `last-verified` (#7710)"* on `main`). The field reads stale only because that merge landed **after** the same Monday's 11:17Z tick, so the first post-merge run is 2026-09-14. Restored to `IN-PROGRESS` with the merged PR named.

### P1 — blocks merge

**F1 · The DR path hands the operator a dead ruleset id** (`infra/github/README.md`)
Routing row *"A ruleset was deleted outright | `scripts/create-ci-required-ruleset.sh`, then import"*. That script POSTs a **new** ruleset (`gh api "repos/${REPO}/rulesets" -X POST`), so GitHub assigns a fresh id and `14145388` ceases to exist. "Then import" 404s, and `apply-github-infra.yml`'s `import_ruleset github_repository_ruleset.ci_required 14145388` aborts under `set -euo pipefail`. The two prescribed steps cannot compose.

**F2 · Art. 30 PA12 §(g)(4) certifies that broken limb as an Art. 32 measure** (`article-30-register.md`)
The cell lists *"re-import from the hardcoded resource ids"* and the DR script as coordinate measures, when running the script invalidates those ids. The same cell says it is *"certified as reviewed in full rather than patched at the one item the issue named"* — so a statutory register now certifies in full a recovery model whose steps don't compose. That is the defect class this PR exists to remove.

**F3 · The state-loss limb is create-shaped, and the destroy guard is blind to it** (ADR-006)
This one is mine, verified end-to-end. The amendment's per-root warning is scoped to *"Where a snapshot **is** taken"* and describes the hazard as *"destroys and replacements"*. The other limb — state lost, config reverted, apply-on-merge reconciles — is create-shaped:

- a resource absent from state plans as a **create**, so `destroy_count == 0` and `[ack-destroy]` is never demanded;
- `luks_passphrase_rotations` counts only update/delete/forget — `tests/scripts/lib/destroy-guard-filter-web-platform.jq` states verbatim *"A first CREATE is legal and expected"*;
- `host_creates` counts `hcloud_server` only.

So a **total** state loss is blocked (by the `host_creates` HALT), but a **partial** loss confined to secret material is blocked by nothing — and those addresses are `-target`-reachable from the per-merge apply (`apply-web-platform-infra.yml`, default allow-list, `:683` `tls_private_key.ci_ssh`, `:725` `random_password.inngest_redis_luks`). A create there mints a new passphrase against a live LUKS header cut from the old one: exactly what the LUKS HALT exists to prevent, reached by the create path instead of the rotate path. `tls_private_key.ci_ssh` has no HALT at all — regenerating it rotates `DEPLOY_SSH_PRIVATE_KEY` in Doppler while the new public half never reaches `authorized_keys` (`ci-ssh-key.tf`: *"Operator-local apply only"*), locking CI out of the deploy path.

Suggested paragraph for the amendment, after the per-root snapshot paragraph:

> **The state-loss limb is create-shaped, and the destroy guard is blind to it.** The paragraph above scopes its warning to *restoring a snapshot*. The other limb — state lost, config reverted, apply-on-merge reconciles — carries a different hazard on `apps/web-platform/infra/`, and it is the limb the recovery model reaches for first. A resource absent from state plans as a **create**, not a destroy, so `destroy_count` is 0 and `[ack-destroy]` is never demanded. The two tripwires outside that sum do not close it: `luks_passphrase_rotations` counts only update/delete/forget (*"A first CREATE is legal and expected"*), and `host_creates` counts `hcloud_server` only. A TOTAL state loss is therefore blocked by the `host_creates` HALT; a PARTIAL loss confined to secret material is not blocked by anything. **So config-revert is NOT the safe first move on this root while state is missing** — re-import every `random_password`, `random_id` and `tls_private_key` from its Doppler escrow before any apply. The `[ack-destroy]` prompt will not appear, and its absence is not evidence the plan is safe.

### P2 / P3

| id | file | finding |
|---|---|---|
| F4 | README | `marketplace_pr_required` row says "re-apply" — but it lives on `jikig-ai/soleur-marketplace`, `import_ruleset` hardcodes `soleur:$id` and cannot express it, so a state loss makes the apply compute a CREATE against the live `"Marketplace PR Required"` → name collision → the apply fails **as a unit**, blocking all five other resources. Import it (`gh api repos/jikig-ai/soleur-marketplace/rulesets --jq '.[]\|{id,name}'`); do not rename the live ruleset. |
| F5 | README | Four classes resolve to two remedies; rows 1+2 share one operator action and row 3 is literally 1+2. Suggest collapsing to two rows. |
| F6 | README | §5d is ~45 lines (longer than 5a+5b+5c) and is reachable only from a snapshot the non-technical operator can never have created. Move into §"Authorization model: apply-on-merge", beside the instruction that creates it. |
| F7 | README | The removed GitHub-UI pointer was the only zero-terminal gesture. Dropping "5 baseline checks" was right (stale against 23); dropping the URL was not. |
| F8 | README | Row 2's discriminator ("state lost, resources intact") names no operator-visible symptom — it needs `terraform state list` + R2 creds. Dead branch for the intended reader. |
| F9 | `plans/2026-05-16-...ruleset-widening...md` | Still carries `s3api list-object-versions` / `?versionId=` under a heading that calls itself *"canonical form for infra/github/README.md"*. Regenerating from it reintroduces exactly what #7836 removes. Needs a SUPERSEDED banner. (Mechanical cause: AC10's sweep is path-scoped and excludes `knowledge-base/project/plans/`.) |
| F10 | ADR-006 | The `## Context` serialization sentence named 2 of 5 roots — **appears already fixed in your working tree** by the "Where serialization does and does not reach" table. Worth confirming `cla-evidence/` is called out as having no CI writer at all. |
| F11 | `apply-sentry-infra.yml` | *"NEW EGRESS SURFACE… takes the WHOLE root's state out from behind the R2 credential boundary"* reads as every-run; the only `upload-artifact` is `if: failure()`-gated. Add "on failure". |
| F12 | `apply-sentry-infra.yml` | Counts rot: *"89 resources"* was already one behind at authoring (HEAD declares 90), and *"the 27 addresses"* is branch-local — `a866e1f29` is on `main`, **is not an ancestor of this branch**, moves `sentry_alert` 27→28 and edits this same file. Suggest non-numeric phrasing. **This branch is behind `main` on a file it edits — rebase before merge.** |
| F13 | `plans/2026-05-29-...apex-canonicalizer...md` | `logs:` block cites *"R2 terraform state history"* as an evidence store and *"R2 state versioning"* as retention. Neither exists — points an incident responder at a store that has never held anything. |
| F14 | `ruleset-ci-required.tf` | Header says *"the 20 `context` strings below"*; the file holds 23 — newly conspicuous now that §5b cites the true count. |

### Verified sound — please don't re-derive

- **`use_lockfile = false` on all five backends** — enumerated file-by-file, exact.
- **No `cancel-in-progress: true` anywhere** in `.github/workflows/`; every terraform writer sets it explicitly `false`. No state-corruption-by-cancellation surface.
- **GitHub App auth** — `infra/github/main.tf:31-38` `app_auth` block, App 3261325 / installation 122213433, credentials in Doppler `prd_terraform`, migrated #4384. PA12 §(g)'s rewrite is accurate.
- **R2 at-rest encryption** — substantiated by `scripts/encryption-posture-ledger.json` → `r2.terraform_state_backend`, `provider-managed:Cloudflare-R2-SOC2-Type-II`, AES-256-GCM, attestation retrieved 2026-07-24.
- **23 required contexts** is correct, and the canonical JSON carries 23, so the DR script does restore the full set.
- **"Five roots" in §5d is correct.** I initially raised this as an off-by-one against #7836's "six"; that was **my** error, from a `grep --include=*.tf` that silently didn't filter and pulled in markdown. `git grep -l 'soleur-terraform-state' -- '*.tf'` resolves to five. No committed artifact asserts "six roots" anywhere. Finding withdrawn.

---

https://claude.ai/code/session_01J9EPUQaD9e8pqGxCPUaYGg

Issue #7867, opened 2026-09-06
New Better Stack sources 202-ack and store nothing — only the May-created source ingests (blocks git-data birth)


deruelle:
Follow-through for #7855. The fix there established the *invariant* — an HTTP 202 from Better Stack is not evidence of storage — and shipped the instruments. This tracker holds the one claim that could not be settled at merge time: **what the real POST→queryable latency for source `2734275` actually is**, measured rather than assumed.

## Why this is a tracker and not a constant in the PR

The plan deliberately refused to write a latency constant it had not measured. At plan time the warehouse was storing nothing from any producer (#7811, since closed), so the measurement was impossible and any number in the diff would have been fabricated. The poll budget in the probe is therefore a *stated multiple* of the only measured figure that exists — ADR-172's 17 s POST→queryable latency, recorded 2026-08-06 against a **different** source (`2457081`, `eu-fsn-3`, `vector` platform). Source `2734275` is `eu-central-1a` and `http` platform, so that figure is a borrowed floor, not a measurement of this source.

## Close criteria

The sweeper closes this when `scripts/followthroughs/betterstack-roundtrip-latency-7855.sh` exits `0` (`ROUNDTRIP_STORED`) — i.e. a marker written to source `2734275` was read back out of the warehouse. Its output records both the wall-clock latency and the vendor's own `ingest_time - dt`, which answer different questions.

The probe has four verdicts over the sweeper's three actions, so the stdout token separates the two that share one:

| verdict | exit | sweeper | meaning |
|---|---|---|---|
| `ROUNDTRIP_STORED` | 0 | closes | a marker was written and read back |
| `ROUNDTRIP_NOT_STORED` | 1 | comments, stays open | **the finding.** The warehouse is demonstrably storing other producers' rows and did not store ours — Better Stack acknowledged a write it did not store |
| `ROUNDTRIP_DARK` | 2 | retries | the warehouse is storing nothing from anyone; not a fact about this source |
| `ROUNDTRIP_UNKNOWN` | 3 | retries | not established — missing credential, failed read, or a poll budget below the 17 s floor |

A `ROUNDTRIP_NOT_STORED` is **not** a reason to close this issue. It is the substantive finding, and it should be escalated to Better Stack with the marker and timestamps from the sweeper comment.

## Known limitations, recorded rather than discovered later

[...]

Issue #7855, opened 2026-09-04
Better Stack returns 202 but stores nothing for git-data source 2734275 — rung-2 rehearsal cannot reach a verdict, birth blocked


deruelle:
The first real rung-2 rehearsal since 2026-07-31 ran today as [33888071954](https://github.com/jikig-ai/soleur/actions/runs/33888071954). It returned `RUNG2_CAPTURE_VERDICT=2` (TRANSIENT) on all 20 capture attempts, wrote no evidence, and the birth gate remains correctly HELD. Teardown verified against the Hetzner API — no surviving host.

**The good news first: #7204 is fixed.** The host booted cleanly past the stage that killed the previous attempt.

## What blocked the verdict

Every capture attempt failed with `the Better Stack query transport exited 22` — a 500 `CLUSTER_DOESNT_EXIST` for `t520508_soleur_git_data_prd_logs`. Better Stack creates the ClickHouse table lazily on first stored row, and **no row has ever been stored** in source `2734275`. The capture behaved correctly: it refused to read a broken instrument as a dark boot.

## What is measured, and what is not

Sentry shows the host emitting through `bootstrap complete`:

```
15:15:34  [info]    git-data boot stage
15:15:55  [info]    git-data cloud-init reached runcmd
15:15:56  [warning] git-data sshd restart failed (config valid)
15:16:06  [info]    git-data bootstrap complete
```

Those come from `/usr/local/bin/git-data-emit` itself (`cloud-init-git-data.yml:643`), so the emitter ran.

Ruled out by measurement, each independently:

| hypothesis | result |
|---|---|
| token empty on host | **no** — 24 chars, threads correctly through `--name-transformer tf-var` |
| wrong token for the source | **no** — Doppler value and the source's API token are sha256-identical (both sources checked) |
| wrong ingest URL | **no** — rendered `user_data` posts to `https://s2734275.eu-central-1a.betterstackdata.com/`, matching `ingesting_host` |
| template didn't render the block | **no** — rendered bytes contain the POST with a non-empty token |
| wrong ClickHouse table name | **no** — convention verified against the working source (`soleur_inngest_vector_prd_3` → `t520508_soleur_inngest_vector_prd_3_logs`) |
| ingest failing and being reported | **no** — `_bs_mirror` emits a distinctly-titled *"git-data Better Stack ingest unavailable"* on **both** the POST-failure and no-token branches; no such event exists in the full issue list for the window |

[...]

PR #7999, opened 2026-09-09
fix(observability): a stale SQL API connection returns the same 701 as a source that never stored

deruelle:
## What this fixes

`CLUSTER_DOESNT_EXIST` has **two** causes. The capture asserted only one of them, and that cost five days of investigation on a source that was working correctly.

**Measured 2026-09-09.** Our Better Stack SQL API connection was provisioned **2026-06-01**; source `2734275` was created **2026-09-03**. A connection does not cover sources created after it, so every read against the newer source returned:

```
Code: 701. DB::Exception: Requested cluster 't520508_soleur_git_data_prd_logs' not found. (CLUSTER_DOESNT_EXIST)
```

…while the source was storing normally the entire time. The message says **cluster**, not **table**.

## Why the control read does not discriminate

This is the part worth understanding. #7856 added a control read to separate "this source is empty" from "the warehouse is down", and it is correct as far as it goes. But **the control table predates the connection**, so it answers in *both* cases:

| state | target read | control read |
|---|---|---|
| nothing ever stored to this source | fails 701 | answers |
| **connection older than the source** | fails 701 | answers |

Indistinguishable. The classifier resolved it to the first and printed **"NEVER STORED A ROW"** as fact.

Everything downstream was self-consistent and false: `202`-acknowledged writes *were* being stored; a disposable `platform=vector` test source read empty for the same connection-scope reason; and a rung-2 rehearsal that had **actually passed** reported `TRANSIENT` twenty times. All 16 rows were in `s3Cluster(primary, t520508_soleur_git_data_prd_s3)` the whole time — including that rehearsal's `boot_complete` with every assertion `yes`.

## What changes

- **The capture no longer asserts a cause it did not measure** (AP-021). It names both and carries the one-call diagnostic that settles them — compare the connection's `created_at` against the source's. The `NEVER STORED A ROW` **marker string is retained deliberately**: the rehearsal workflow and two suite arms branch on it, and renaming it would turn a message fix into a silent behaviour change.
- **The workflow's operator summary** drops the false heading and reorders the next actions: check connection age *first*, table name only if the connection is newer. It keeps "do NOT re-dispatch" — no paid host can fix a read.
- **`betterstack-log-query.md`** records the durable rule: *adding a Better Stack source is not complete until the SQL API connection is re-provisioned* — with the failure it produces, the diagnostic, and the **additive** create-then-rotate order (`POST /api/v1/connections` is non-destructive; delete-first is not).
- Also recorded: the connection was re-provisioned 2026-09-09 (id `424104`, host `eu-central-1a-connect`), `BETTERSTACK_QUERY_{HOST,USERNAME,PASSWORD}` rotated in `soleur/prd_terraform`, old connection left in place so consumers holding the previous credentials keep working. Plus the hot-vs-archive split: a fresh row is in `remote()` within **~75 s**, rows five days old are **only** in `s3Cluster(primary, <table>_s3)` — measured.

## Verification

| check | result |
|---|---|
| `test-git-data-rung2-evidence-capture.sh` | **78 / 0** |
| markdownlint on changed docs | clean (file was clean at HEAD; the one error I introduced is fixed) |
| gitleaks (staged) | no leaks |
| workflow YAML parses | yes |

## Not in this PR

The rung-2 evidence file that releases the birth gate. That is the gate-releasing change and gets its own review — the evidence file's own header says it must not be auto-committed, because "a route that writes its own gate-releasing evidence is self-approving".

Refs #7867, Refs #7855


deruelle:
### Heads-up: this branch will hit a modify/delete conflict on its first sync after #8384 merges

**Why you are seeing this:** #8384 (ADR-235) untracks four generated knowledge-base caches and regenerates them on demand. Your branch still tracks all four, so the first `git merge origin/main` (or rebase) after it lands will conflict — `deleted in origin/main and modified in HEAD` — on any of these your branch has touched:

```
knowledge-base/INDEX.md
knowledge-base/kb-tags.txt
knowledge-base/kb-categories.txt
knowledge-base/project/rule-metrics.json
```

This is a one-time transition, not a defect. These files are caches of the tree with no product or human consumer; nothing is lost by dropping them, and `scripts/ensure-kb-index.sh` regenerates them on demand.

**Remedy — run this on your branch, then sync:**

```bash
git rm --cached -- knowledge-base/INDEX.md knowledge-base/kb-tags.txt \
  knowledge-base/kb-categories.txt knowledge-base/project/rule-metrics.json
git commit -m "chore: untrack the generated kb caches (ADR-235 transition)"
```

If you have already started the merge and are sitting in the conflict, `git rm` the four paths to accept the deletion, then continue.

**Note the automated resolver will NOT do this for you.** `plugins/soleur/scripts/resolve-regenerable-conflicts.sh` deliberately refuses `modify/delete` conflicts: it cannot know which side of a delete-vs-modify is intended, and guessing would be the dangerous behaviour. Refusing is correct — the remedy above is the manual step it is refusing in favour of.

**Why you are being told here rather than in the docs:** the remedy is documented on the *new* `main` (`plugins/soleur/skills/merge-pr/SKILL.md`), but your branch reads its own copy of that file, so it cannot reach you until after you have already hit the conflict.

Posted to every open PR whose branch still tracks these files. No action needed until #8384 merges.

Issue #7772, opened 2026-09-03
git-data pre-birth hardening: per-source ingest token, metadata egress rule, render-arg evidence binding (deferred from #7460)


deruelle:
Three pre-birth hardening items deferred from #7460 (PR B). They are one tracker because they share a subsystem, a threshold, and — critically — the same re-evaluation trigger.

**Re-evaluate: BEFORE the git-data host is born.**

That trigger is not boilerplate. `user_data` is ForceNew with no `ignore_changes` (ADR-149/ADR-152), so every one of these is free to change today and costs a destructive host replace the moment the host exists. The host has never been born — live Hetzner holds no `soleur-git-data`.

---

## 1. A per-source Better Stack ingest token

#7460 bakes the SHARED ingest token into git-data's `user_data`. Measured 2026-09-03 against the Better Stack API: exactly **one** Logs source exists (`soleur-inngest-vector-prd`, id 2457081), and `var.betterstack_logs_token` fans out to the Inngest host's bake, the zot registry's Doppler secret, git-data's Doppler secret, and the web host's Vector sink.

So a git-data metadata leak forces a rotation that darkens the web host's shipper and the registry's, and requires **two** host replaces. Under a `single-user incident` threshold, baking the lowest-trust host's copy of a credential shared with three others is the wrong direction — the host serving untrusted user pushes is the last one that should hold the shared value.

A dedicated git-data source shrinks forged-row blast radius to git-data's own stream and satisfies leg (3) of ADR-198's capability test ("single-purpose to this host"), which is currently the open residual.

**Why deferred, measured rather than preferred:** the Better Stack provider in this repo is `betterstackhq/better-uptime` — an *uptime* provider. Logs **sources** are not expressible in it, and `apps/web-platform/infra/inngest.tf` already records that as an **"IaC gap"**. `apps/web-platform/infra/variables.tf` states the ingest token is minted at source and only *copied* into Doppler. Adopting per-source therefore needs either a provider that does not exist here or an operator mint that `hr-all-infrastructure-provisioning-servers` forbids doing ad hoc. That is a work-stream, not a variant of #7460.

## 2. An egress rule blocking the metadata endpoint for non-root UIDs

`hcloud_firewall.git_data` declares **zero rules**, which under Hetzner semantics is inbound-denied and **egress fully open**. Any code execution as the `git` account can `curl` the metadata endpoint and read the entire `user_data` — `doppler_token`, `sentry_dsn` and the newly-baked ingest token alike.

[...]

Issue #7204, opened 2026-08-03
fix(infra): git-data rung-2 rehearsal dies at stage:luks_open — the birth mkfs sets an ext4 quota feature the target image's kernel cannot mount


deruelle:
## What happened

The rung-2 boot rehearsal (`.github/workflows/git-data-rung2-rehearsal.yml`) run **[30649892865](https://github.com/jikig-ai/soleur/actions/runs/30649892865)** (2026-07-31T17:06:57Z, `workflow_dispatch`, branch `main`) failed at step *Capture rung-2 evidence (bounded poll)* with a REAL host verdict — `RUNG2_CAPTURE_VERDICT=1` (FAIL), not TRANSIENT and not a wrapper/toolchain fault.

The `git-data-rung2-capture-log` artifact carries the verdict but **not the cause**:

```
FAIL: soleur-git-data-rehearsal-30649892865 reported a FATAL. The rehearsal found the failure class this route exists to catch — a boot that would have looked green from the apply.
{"dt":"2026-07-31 17:11:22.303685","stage":"luks_open","level":"fatal","host":"soleur-git-data-rehearsal-30649892865","luks_mounted":"","repo_root":"","hooks_path":"","provision":""}

NO EVIDENCE FILE WRITTEN. Fix the cause, then re-run the rehearsal against the corrected template.
RUNG2_CAPTURE_VERDICT=1
```

## The cause, self-pulled from the observability layer

The artifact row omits the `detail` field because `HOST_SQL` in `scripts/followthroughs/git-data-rung2-evidence-capture.sh` does not `SELECT` it. Re-querying the same Better Stack source with `detail` included (`doppler run -p soleur -c prd_terraform -- scripts/betterstack-query.sh`) returns:

```json
{"dt":"2026-07-31 17:11:22.303685","raw":"{\"message\":\"git-data LUKS stage FAILED\",\"stage\":\"luks_open\",\"level\":\"fatal\",\"host_name\":\"soleur-git-data-rehearsal-30649892865\",\"detail\":\" disable this, use --no-read-env. mount: /mnt/git-data-luks: mount(2) system call failed: No such process.        dmesg(1) may have more information after failed mount system call.\",\"rc\":\"32\"}"}
```

Sentry (org `jikigai-eu`, project `web-platform`) carries the full stage timeline for the same host:

| time (UTC) | stage | level | note |
|---|---|---|---|
| 17:10:46 | `bootcmd_start` | info | |
[...]

Issue #7025, opened 2026-07-27
git-data: clear the DO-NOT-DISPATCH banner once the emitter is proven to emit (W12 rehearsal)


deruelle:
## Clear the git-data DO-NOT-DISPATCH banner, once the emitter has been seen to emit

**Precondition — this is the whole issue.** #6982 released the *mechanical* interlock:
`git-data-birth-readiness-gate.sh` no longer refuses, because the `${sentry_dsn}` sentinel
now threads through `cloud-init-git-data.yml`. The DO-NOT-DISPATCH banner in
`knowledge-base/engineering/operations/runbooks/git-data-birth.md` is deliberately
**retained** and is now the only thing holding the route.

That state is a real, accepted regression in hold-strength — "held by convention", which
ADR-149's own Alternatives table rejects as *"held until the first person who reads the
runbook and not the plan."* It is accepted because the alternative is worse: clearing the
banner over an emitter nobody has ever seen emit.

### Why the banner could not be cleared inside #6982

A PR merges **atomically**. A banner cleared in its final commit clears at the same instant
as all the untested code it is supposed to be downstream of. Sequencing cannot fix that
within one PR, so the clear moves out.

### The release condition (do not clear on anything weaker)

Every gate #6982 ships is **static**, and the failure class it defends against — *green
apply, dark host* — is only observable at **runtime**. Mutation arms prove the code CAN go
red when neutered; they never prove an event ARRIVES when it is intact. The interlock is
also a one-bit latch: the bit flips on **threading**, not on **emitting**, so a
`grep -c '${sentry_dsn}'` proves nothing about whether anything was received.

Required evidence, all three artifacts, each with the query that retrieved it:

- [ ] a **Sentry event from the fatal channel** (the baked-DSN path)
- [ ] **≥1 Better Stack stage marker**
- [ ] one **`stage:boot_complete` row** carrying its four assertion booleans
      (`luks_mounted`, `repo_root`, `hooks_path`, `provision`)

**Obtained from a real boot**, i.e. the rendered template booted **once on a throwaway
host** outside the `hcloud_server.git_data` address, with stub volume ids and a scratch
Doppler config, then destroyed.

**The container-harness rung is NOT sufficient.** It cannot exercise `doppler run` against
real Doppler, `luksOpen` against a real volume, the private NIC, or whether an event
actually lands off-box. #6982's Phase-0 W0 probe is precisely the defect class that passes
every static check and every mutation arm and then dies on first boot — it was found by
[...]

PR #8002, opened 2026-09-09
feat(git-data): commit the rung-2 boot evidence that releases the birth hold

deruelle:
Commits `apps/web-platform/infra/git-data-rung2-boot-evidence.env` and nothing else.

## Merging this file RELEASES the birth hold

`git_data_rung2_rehearsal_gate` HOLDs while this file is absent from `main`. It is the binding mechanical hold on the birth of the host that will store every connected user's source code.

The file is deliberately **not** auto-committed by any workflow. Its own header states why: *a route that writes its own gate-releasing evidence is self-approving.* **This PR is the human approval.** Nothing else about the merge is special — the release happens because the file lands.

Verified locally, both directions:

| | gate verdict |
|---|---|
| evidence present | `RELEASED` (rc=0) |
| evidence moved aside | `HOLD` (rc=1) |

So the gate is not vacuous.

## What the evidence is

Produced by `scripts/followthroughs/git-data-rung2-evidence-capture.sh` against the 2026-09-04 rung-2 rehearsal (run [33888071954](https://github.com/jikig-ai/soleur/actions/runs/33888071954), host `soleur-git-data-rehearsal-33888071954`). Not hand-written.

```
RUNG2_CAPTURE_VERDICT=0
RUNG2_BOOT_REHEARSAL=PASS
RUNG2_SENTRY_CROSSCHECK=CLEAN
```

**The evidence is dated 2026-09-04 and is REUSED, not re-run, because the hash binds it.** `RUNG2_TEMPLATE_SHA256` is `3a2392fb5b0d4fae9d4abeaf5ca10ee682430e473dd9eac1ca3d340ef4ce1725`, recomputed with the gate's own `git_data_rung2_user_data_sha256` at this branch tip *and* against `origin/main`'s copies — identical. The hash binds **13 files** — the cloud-init template, the three `modules/git-data-userdata/*.tf` files, and nine payloads (`git-data-bootstrap.sh`, `git-data-gc.{sh,service,timer}`, `git-data-gc-failure.service`, `git-data-pre-receive-placeholder.sh`, `git-data-provision.sh`, `git-data-remove.sh`, `git-data-transport-wrapper.sh`). All 13 are byte-identical between the rehearsal commit `0f39b7aa2` and current `main`. Had any of them moved, the evidence would be void and a fresh paid rehearsal would be required; that is the whole point of the hash.

`RUNG2_VAR_DIVERGENCE` is taken from `REHEARSAL_DIVERGENCE` in `.github/workflows/git-data-rung2-rehearsal.yml` **at `0f39b7aa2`** — the commit the rehearsal actually ran against. It is identical on `main`, but the run's own commit is the correct provenance.

## Why it reported TRANSIENT at the time

The rehearsal always passed. It read `TRANSIENT` x20 on the day because our Better Stack SQL API connection was provisioned 2026-06-01 and could not see source 2734275, created 2026-09-03 — **a stale connection returns the same `701 CLUSTER_DOESNT_EXIST` as a source that never stored a row.** The connection was re-provisioned 2026-09-09; the rows read back cleanly from the S3 archive, which is what the capture's `s3Cluster` UNION arm is for.

PR #7999 corrects the capture *message* that produced the false "NEVER STORED A ROW" reading. Merging it first is preferred but not required — it changes what the capture says, not what this evidence is.

## The capture's own caveat, stated plainly

`boot_complete`'s four booleans are **hardcoded string literals** in `git-data-bootstrap.sh` — the emit call passes `luks_mounted=yes repo_root=yes hooks_path=yes provision=yes` verbatim. So this attests that **the final stage was REACHED and nothing reported a fatal** — NOT that four invariants were independently measured. The capture says so itself, in its PASS message and in the committed file.

`nft_metadata_drop` is the exception: it is computed just above the emit by grepping the live nftables chain (`nft list chain inet soleur_git_data output` for `169.254.169.254`), anchored on the metadata-service address rather than the table name, so a table that exists with its rule flushed reads `no`.

**What "reached" is worth here.** Reaching that emit is not free — each literal has a named upstream `FATAL: …; exit 1` gate, and the script's log router emits a `level:fatal` event on every one:

| literal | upstream fatal gate |
|---|---|
| `luks_mounted` | volume not mounted; empty LUKS key; no LUKS volume found; `luksOpen` failed; LUKS root not mounted; **mounted from a non-LUKS device** (#7204's "mountedness is not identity" assertion); plus post-bootstrap re-checks |
| `repo_root` | `$GIT_HOME/repositories` does not resolve to `$REPO_ROOT` |
| `hooks_path` | pre-receive hook missing/not executable; `core.hooksPath` not set |
| `provision` | `git-data-provision.sh` missing/not executable |

That is the correct reading: the literals are not measurements, but the boot could not have got to them with any of those conditions false.

> [!IMPORTANT]
> **`luks_mounted` is about the DEVICE. It does NOT mean user source code is encrypted at rest.**
> `git-data-bootstrap.sh` pins this wording itself (AC30): *"they are NOT, REPO_ROOT is the PLAINTEXT volume until the cutover, and duplicating that false claim into telemetry would be the Art. 30 defect in a second artifact."*
> `REPO_ROOT` being plaintext pre-cutover is **by design**, not a defect this evidence missed. Read "#7204's LUKS fix holds" as *the LUKS device mounts and is the device we think it is* — nothing more.

Two narrower corrections to the same table, so it is not read as stronger than it is:

- **The zero-fatal read is not independently load-bearing.** Twelve `log "FATAL:"` sites route a `level:fatal` event, but the script is `set -euo pipefail`, so a failed `mkdir`/`chown`/`install`/`git config` aborts with **no** `log` call and therefore no fatal row — and the emitter is explicitly fail-open (`[ -x "$GIT_DATA_EMIT" ] && … || true`). What excludes a silent abort is not the absence of fatals; it is the **arrival of `boot_complete`**.
- **`hooks_path=yes` attests the fail-closed PLACEHOLDER, not the real CAS fence.** Only `git-data-pre-receive-placeholder.sh` is among the nine `file()`-bound payloads — `git-data-pre-receive.sh` is not, so `RUNG2_TEMPLATE_SHA256` does not bind it. The real fence arrives later via the deploy pipeline, outside this hash. Lost-update/write-skew on a user's repo is outside everything this evidence attests.

## What the rehearsal proved

Pulled directly from the git-data source (hot ∪ `s3Cluster` archive, `INTERVAL 7 DAY`, filtered to the rehearsal host):

```json
{"stage":"boot_complete","level":"info","luks_mounted":"yes","repo_root":"yes",
 "hooks_path":"yes","provision":"yes","nft_metadata_drop":"yes"}
```

Zero `level:fatal` rows. So **#7204's LUKS fix holds and #7772's metadata-egress control loaded on a real Hetzner host.**

> [!WARNING]
> **The transcribed queries have a ~7-day shelf life.** Every ARTIFACT query is bounded `dt > now() - INTERVAL 7 DAY`, and the rehearsal ran 2026-09-04 — so from roughly **2026-09-11** they return zero rows, which is byte-identical to the reading a never-booted host gives. The committed file records **queries, not rows**, so the row pasted above is the durable record; re-running the queries after that date proves nothing either way.

**Provenance of that row.** `nft_metadata_drop` is in *neither* committed artifact — the `.env` records queries rather than rows, and the capture's `__HOSTROWS__` SELECT projects only the four hardcoded booleans (the same four its false-assertion arm greps). It was read directly from Better Stack, with this query, so a reviewer can re-run it:

```sql
SELECT JSONExtractString(raw,'stage')             AS stage,
       JSONExtractString(raw,'luks_mounted')      AS luks_mounted,
       JSONExtractString(raw,'repo_root')         AS repo_root,
       JSONExtractString(raw,'hooks_path')        AS hooks_path,
       JSONExtractString(raw,'provision')         AS provision,
       JSONExtractString(raw,'nft_metadata_drop') AS nft_metadata_drop
FROM (SELECT dt, raw FROM remote($BS_TABLE)
      UNION ALL
      SELECT dt, raw FROM s3Cluster(primary, $BS_TABLE_S3) WHERE _row_type = 1)
WHERE dt > now() - INTERVAL 7 DAY
  AND JSONExtractString(raw,'host_name') = 'soleur-git-data-rehearsal-33888071954'
ORDER BY dt DESC LIMIT 50 FORMAT JSONEachRow
```

## Disclosure — read this before approving

Three things a reviewer would otherwise get wrong. All measured, not asserted.

**1. The run this file cites is marked `failure`, and that is expected.**

`gh api repos/jikig-ai/soleur/actions/runs/33888071954` returns `conclusion: failure` — the job `rehearse` failed at step *Capture rung-2 evidence (bounded poll)*. **Click the URL and you land on a red run.** The host booted and reported; what failed was the in-run poll's ability to *read* Better Stack, for the stale-connection reason above. The URL is the provenance of **the boot**, not of **this file**.

**2. This file was produced out of band, five days later.**

`# Captured (UTC) : 2026-09-09T19:47:02Z` versus a run created `2026-09-04T15:11:42Z`. The runbook's `## After a PASS` path downloads the run's own artifact; that path does not apply, because the run wrote no artifact. This was captured locally and post-hoc from the S3 archive after the SQL connection was re-provisioned. That is genuinely weaker for auditability than an in-run artifact, and it is why every query is transcribed verbatim into the file — so anyone with a working connection can re-run them.

**3. What the gate actually checks is narrower than its name.**

`git_data_rung2_rehearsal_gate` strips comments, then asserts: four keys each appear exactly once; `RUNG2_BOOT_REHEARSAL=PASS`; the URL matches an Actions-run *shape*; the sha is 64 hex; the divergence tokens are all in an identity-only allowlist; and the sha equals `git_data_rung2_user_data_sha256` of the current template.

Its predicate is therefore **"a well-formed assertion exists that is bound to this template"** — *not* "a rung-2 rehearsal passed". Concretely, measured:

- The gate does **not** resolve the run id, check the run's conclusion, or check which workflow it belongs to. A four-line hand-written file naming a nonexistent run releases identically.
- Everything that reads like evidence — the host name, the four ARTIFACT query blocks, the capture timestamp, the header — is a `#` comment, and the gate strips comments before reading anything.
- **`RUNG2_SENTRY_CROSSCHECK` is read by no gate at all.** The capture writes it; nothing consumes it. `CLEAN` above is a human-facing claim.
- The hash is a **staleness detector, not an authorship proof**: it is a pure function of tracked files, computable by anyone with a checkout.

None of that is a defect being introduced here — the file's own header concedes the point (*"a route that writes its own gate-releasing evidence is self-approving"*). It is the reason this PR exists as a human gate rather than an automated commit. **The provenance control is a person reading this PR, and that person is you.** Hardening the gate (binding the host-name suffix to the URL's run id, resolving the run in CI, making the Sentry key load-bearing) is **tracked at #8010**; it is deliberately not done here, because adding a key would require regenerating the evidence and voiding its capture provenance.

**The one measured boolean has no durable record.** `nft_metadata_drop` is the only value here that was actually measured on the host, it is the entire basis for the metadata-egress claim, and it appears in **neither committed artifact** — not in the `.env` (which records queries) and not in the capture's `__HOSTROWS__` projection. It survives only as the row pasted above. A future capture would have to re-read it directly.

**The rehearsal exercised ONE SSH identity; production has three, and they are capability grants.**
`cloud-init-git-data.yml` writes three `authorized_keys` lines, each pinning a *different* forced
command to a *different* pubkey — `git-data-transport-wrapper.sh`, `git-data-provision.sh`, and
`git-data-remove.sh` (the Art. 17 erasure path). `rung2-rehearsal/rehearsal.tf` sets all three to the
same `tls_private_key.rehearsal`, so **the three-forced-command dispatch was never exercised at all.**

That matters beyond coverage: because the three were already identical in the rehearsal, a future edit
that renders one pubkey into all three slots is a **no-op there** — `boot_complete` still emits, no
fatal appears, the evidence records PASS, and the hash moves so the file looks freshly re-rehearsed —
while in production the transport key (held by the web app for ordinary push/fetch) would silently
acquire the erase capability. sshd matches by key and takes the first match, so it never surfaces as a
failure. Nothing in this evidence would catch it. The gate's allowlist calls these vars
"identity-shaped"; for these three they are a **capability** divergence, not an identity one. A static parity assertion (three distinct pubkeys in the rendered
`authorized_keys`) belongs in the birth gate — **tracked at #8009**, not done here.

**"Two intentional human gates" may be one human twice.** The capture script's stdout (not the committed file — corrected at CPO sign-off) calls merging it the second of two intentional human gates; the first is the `web-platform-infra-apply` environment approval (1 required reviewer, verified non-empty). Measured against the live API, that environment has `prevent_self_review: false` and exactly one reviewer — so **the dispatcher can approve their own deployment.** This is not a two-party control, and if the same person merges this PR it is one human clicking twice. Do not treat the pair as defence in depth.

**And the freshness check is advisory, not blocking.** The `Rung-2 evidence freshness` step lives in job `deploy-script-tests`, which is absent from the `CI Required` ruleset (verified live via `gh api repos/jikig-ai/soleur/rulesets` — 25 required contexts, none of them this one), and `main` carries no required-review rule. So if a future PR edits a git-data payload and turns this step red, `--auto` merge proceeds anyway and the stale evidence lands on `main`. The backstop is the birth job's own call to the same gate at dispatch time — after an environment approval has been spent.

Recorded headless at plan time per ADR-084: points where the review panel and the planning
session agreed the **stated direction** should change. Your brief was kept as the default in both
cases and neither was applied. Surfaced here because you have not seen them.

**DC-1 — `--window '7 DAY'` versus the script's own `30 DAY` default.** Three of five reviewers
flagged it independently. The rehearsal ran 2026-09-04 and the capture ran 2026-09-09, so `7 DAY`
clears it by ~2 days; the capture's own default is `30 DAY`. DHH's framing was *"you manufactured
a problem and then wrote three mitigations for it."* **Settled empirically by the run:** the
capture returned `VERDICT=0 / PASS` on the **first attempt** at `7 DAY` — no widening, no
re-dispatch. The contingency ladder was never exercised. No action needed.

**DC-2 — `--out` is strictly worse than omitting it.** Verified: the capture defaults `OUT` to
`$(dirname "$CLOUD_INIT")/…`, and `CLOUD_INIT` derives from `BASH_SOURCE` — an *absolute* path
that resolves from any working directory. The explicit relative `--out` is the only reason the
plan needs a working-directory precondition at all. **Settled by the run:** the file landed at the
correct path and the gate found it. No action needed.

Neither is filed as an issue. ADR-084's filing step exists so `operator-digest` surfaces dissents
you would otherwise never see — but both of these were answered by the run itself, so an
`action-required` issue would assert an action that does not exist and would grow the backlog for
nothing. They are recorded in `specs/…/decision-challenges.md` and rendered here instead.

## Verification

- One file in the evidence commit; 1 non-`knowledge-base/` file vs `main`.
- `gitleaks` clean; `.gitleaks.toml` unmodified. No Better Stack source id, team id or table name is expanded into the transcribed SQL (literal `$BS_TABLE` tokens only).
- Suites: `git-data-rung2-rehearsal` 83/83, `test-git-data-birth-readiness-gate` 80/80, `git-data-emit` 59/59.
- Full local gate **not** run — five sibling full-gate runs were in flight (`CAPACITY_CONTENDED`); reported as skipped-for-contention rather than forced.

Two acceptance criteria were amended, with the reason recorded in the plan: AC1 expected 2 `# QUERY:` lines (the capture emits 3 — the Sentry cross-check writes one) and AC5 expected 1 `remote($BS_TABLE)` (there are 3). Both were stale plan-time counts. **The artifact was not edited to match**; AC5's load-bearing assertion — that no source identifier is expanded — measures 0 as specified.

Ref #7025 — the birth is not performed here. `git_data_host_create` still runs under `web-platform-infra-apply` (1 required reviewer), and ADR-149's release checklist is unchanged by this PR.

---

Filed: #8009 #8010

**Net-issue-flow override.** This PR closes nothing (`Ref #7025`) and files two, so it is net +2.
Both are the criterion the gate names verbatim — *a discovered defect in another subsystem that
must stay its own issue* — and neither is inlineable here without breaking the diff discipline
this PR's own AC4 asserts:

- **#8009** — the birth gate does not assert the three `authorized_keys` pubkeys are distinct.
  A defect in `tests/scripts/lib/git-data-birth-readiness-gate.sh` and
  `rung2-rehearsal/rehearsal.tf`, found by review of *this* evidence. Inlining it would add a
  second and a third product file to a PR whose AC4 requires exactly one.
- **#8010** — the same gate checks assertion *shape*, not that a rehearsal passed. Its cheapest
  fix touches the gate; its other two touch the capture script, its tests, and the evidence
  **format** — regenerating the evidence would rewrite its `# Captured (UTC)` stamp and void the
  capture provenance this PR exists to establish.

Both were filed at CPO sign-off because the body had claimed them "tracked" while no tracker
existed — an unbacked tracking claim converts a disclosed risk into a forgotten one.


deruelle:
## CPO sign-off — SIGNED-OFF with conditions

`requires_cpo_signoff: true`; `brand_survival_threshold: single-user incident`. Recorded per
tasks.md 5.4. The verdict was taken against the **corrected** User-Brand Impact framing, not the
original — the five-agent review inverted one of that section's own failure shapes.

**Verified rather than read** (each claim re-derived against the repo and the live API): the gate
prints `RELEASED` rc=0 at HEAD; production pins three distinct forced commands to three distinct
pubkeys while `rehearsal.tf` collapses them to one `tls_private_key`; the environment is
`prevent_self_review:false` with reviewers `["deruelle"]`; `deploy-script-tests` is the only
failing check and is not required; the diff is one non-`knowledge-base/` file.

**Does `## User-Brand Impact` still hold after the encryption correction?** Yes — and better than
before. The section does not merely delete the inverted failure shape; it carries a dated
scope-out naming the inversion, citing the AC30 wording in `git-data-bootstrap.sh`, and stating
why the original was dangerous (a reader could infer encryption-at-rest from a green boot). The
surviving failure shapes are all still real and all within what this evidence attests.

**Is the residual risk adequately disclosed?** For this PR's reviewer: yes, exceptionally. For the
operator who must later approve the **birth dispatch**: not yet — a PR body has a readership of one
review, and at dispatch time the operator sees a runbook banner and a bare "Approve deployment"
button carrying none of this. That gap is the substance of the conditions.

### Conditions

**A — before ready. DISCHARGED.** Two risks the body called "tracked" had no tracker. An unbacked
tracking claim converts a disclosed risk into a forgotten one, and this PR's whole credibility
rests on its disclosures being literally true. Filed **#8009** (three-pubkey capability parity —
weighed most heavily: a future collapse silently grants the web app's transport key the Art. 17
erase capability, and rehearsal-derived evidence would still read PASS) and **#8010** (gate checks
assertion shape, not that a rehearsal passed). Both numbers now replace the "tracked" prose in the
body.

**B — before ready. OPEN.** `git-data-birth.md` currently tells the operator *"That file does not
exist"* and *"A dispatch today exits 1 before planning anything."* Both become **false at the merge
instant**: the gate releases, and the banner becomes the only prose hold, backed by a
one-reviewer self-approvable environment. Deferring to #7025 is accepted **only if not deferred to
memory** — the correction must land in the same session as the merge. A separate one-file PR
preserves AC4. The ⛔ *instruction* survives intact, which is why this is a condition and not a
block.

**C — before any birth dispatch, not before ready.** The static three-distinct-pubkey assertion
must be in the birth gate and green, and the dispatch surface must state in plain language the
three things a green boot does **not** mean: repos are not encrypted at rest pre-cutover; the
environment approval is not two-party; the evidence attests *the final stage was reached*, not
*four invariants were measured*. Without that, the approval is ceremonial and "one human clicking
twice" becomes the real control.

### Why not BLOCKED

Merging births nothing, and the banner's ⛔ instruction survives the merge. Blocking is not free:
the hash binds 13 files, so any drift voids the evidence and costs another paid Hetzner rehearsal —
an asymmetric cost that argues for merging promptly once green.

### Two corrections to the PR body — both applied

1. The body claimed *the file's header* calls merging it the second of two human gates. It does
   not (`grep` returns 0); that framing is the **capture script's stdout**. The committed artifact
   is clean of it. Corrected.
2. The body cited 23 required contexts; the measured value is **25**. Corrected.

Issue #8016, opened 2026-09-09
release: canary_sandbox_failed blocked v0.264.6 while the payload's own canary reported pass (recurrence of #2276)


deruelle:
## `canary_sandbox_failed` blocked the v0.264.6 deploy while its own canary reported `pass`

Production did not take the build from merge commit `6012a87bd`. The rollback worked and prod is
healthy on the prior release, but `main` is now ahead of what is deployed.

**This is a recurrence.** #2276 — *"release: canary_sandbox_failed blocking v0.35.15+ deploys
(host-level drift, PRs not at fault)"* — is CLOSED. The same reason string is back.

### The contradiction

`Web Platform Release` run 34409004195, `deploy` job, step 5 "Verify deploy script completion":

```
ci-deploy.sh exited 1 (reason=canary_sandbox_failed, tag=v0.264.6)
```

But the run's own status payload carries a **passing** canary:

```json
"reason": "canary_sandbox_failed",
"sandbox_canary": {
  "verdict": "pass",
  "reason": "ok",
  "sdk_version": "0.3.197",
  "checked_at": 1788991188,
  "consecutive_pass": 435,
  "first_pass_at": 1783102021
}
```

`checked_at=1788991188` is **~31 minutes before** the deploy's own `start_ts=1788993060`. So the
gate did not observe a failing canary — it observed a **stale** one, and the reason string names a
failure that the embedded evidence contradicts. That is the "names a cause the job did not measure"
shape, in the deploy path.

Container health in the same payload is clean:

```
container_exit_code: 0     restart_count: 0     oom_killed: false
seccomp_profile_loaded_matches_host: true     restart_rate_per_hour: 0
```

### Impact

- Prod serves `0.264.5` / `build_sha 6abab7638` (verified via `/health`: `status ok`,
  `supabase connected`). The deploy for `6012a87bd` did not land.
- `live-verify` was skipped (gated on deploy success), so the deployed artifact went unverified.
- The failure notification path worked: `Deploy-failure email sent to ops@jikigai.com (HTTP 200)`.

### Not attributable to the triggering PR

PR #8003 is 10 markdown files — one `plugins/soleur/skills/work/SKILL.md`, the rest
`knowledge-base/`. Zero application code, zero `apps/web-platform/**` source, and plugin skill
files are read from the repo rather than baked into the web-platform image. It changed no deploy
gate either, so `postmerge` Phase 3.7's gate-suspect trigger correctly did not fire. The same
[...]

Issue #6982, opened 2026-07-27
git-data: pre-birth hardening + observability (blocks the first birth)


deruelle:
## Context

#6570 repinned `soleur-git-data` off the unorderable `cax11` and made its arch derived. That change
is IaC-only and the host remains unborn (#6977 — no birth route). The 11-agent review panel on PR
#6974 surfaced a cluster of host-hardening and observability gaps that share one trigger: **they must
land before the host is first born**, because several are unfixable afterwards.

## Items

- [ ] **No log shipper.** `cloud-init-git-data.yml` has **0** occurrences of vector/betterstack/journald
      (measured; the inngest cloud-init has 31/11/5, the web host's 14/2/7). Every boot-time failure on
      a deny-all-public-ingress host is readable only via on-host `/var/log/cloud-init-output.log`.
- [ ] **Heartbeat absent, not merely paused.** Live `GET /api/v2/heartbeats` (2026-07-27) returns 7
      heartbeats; `soleur-git-data-prd` is not among them. Related: #6548, #6975.
- [ ] **No `set -e` on the Doppler runcmd item.** A failed `sha256sum -c -` still `tar xzf`s the
      unverified tarball and `chmod +x`s it to `/usr/local/bin/doppler`, which then runs as root. The
      web host's `cloud-init.yml` already wraps the identical block with `STAGE=doppler_dl` under a
      top-armed `trap on_err EXIT` + `set -e` (#6090); git-data has neither.
- [ ] **No gc/pack tuning on a 2 vCPU / 4 GB box with no swap.** `git-data-bootstrap.sh` sets only
      `core.hooksPath` and `receive.advertisePushOptions`. `receive.autogc` is default-ON,
      `pack.windowMemory` defaults to unlimited, and `gc.autoDetach` means an OOM kill is invisible to
      the pushing client. `git-data-transport-wrapper.sh` denies `git gc`, so there is no manual route.
- [ ] **No quota / reclaim on a 10 GB store.** `replicateToGitData` force-pushes per-worktree
      namespaces at every session end; `--force` orphans prior objects and nothing prunes. No disk
      monitor is installed (`disk-monitor.sh` already exists in-tree).
- [ ] **No `MaxStartups` / `MaxSessions`** in the git-data sshd hardening block, and no client-side
      concurrency limiter. Fine at `replicas=1`; the first thing to bite at ADR-068 Phase-3.
[...]

PR #8026, opened 2026-09-10
fix(ci-deploy): the bwrap probe discarded its own diagnosis

deruelle:
## The bwrap deploy gate discarded its own diagnosis

On 2026-09-09 the blocking sandbox probe rolled back the v0.264.6 deploy and left no evidence of
why. The entire journald record was:

```
DEPLOY_ROLLBACK: bwrap sandbox non-functional in <registry>/<image>:v0.264.6
```

No cause, no exit code. The probe ran `docker exec ... bwrap ... 2>&1` **inside an `if !`**, which
merges its stderr into the deploy script's *stdout* — a stream journald never sees. Only
`logger -t "$LOG_TAG"` lines reach journald, and that line carried none of the probe's output.

This makes the probe self-report: exit code, wall-clock duration, canary container state,
pre-sanitization output length, and the sanitized output ride the same rollback line.

## Ref #8016 — and why this PR does not close it

Deliberately `Ref`, not a close keyword. The closing criteria are stated on the issue and neither is
satisfiable here: Better Stack shows **one** occurrence in 48h and none since, so the
"next occurrence names a cause" arm has nothing to fire on, and the soak arm cannot complete in this
PR **by construction** — landing the marker is its precondition. Closing on merge would be closing
on an assumption.

The issue's original framing was also wrong, and that correction is posted there rather than edited
into the body: `reason=canary_sandbox_failed` comes from this blocking probe, while the passing
`sandbox_canary` block in the same payload comes from `run_faithful_sandbox_canary` — a
**non-blocking dark-launch** canary (ADR-079) invoked as `|| true`, whose own comment says it
"never gates/rolls back this deploy". A naming collision, not a contradiction.

## Three things worth reading

**The exit code gates the rollback, not just the diagnostic.** `VAR="$(cmd)" || RC=$?` is the only
form that preserves it — `if ! VAR=$(cmd)` detects the failure but `!` *consumes* the status, so
`$?` reads 0 (measured). Under mutation, reverting to that form emits **zero** rollback lines: the
`(( BWRAP_RC != 0 ))` branch never fires and the gate **fails open**, shipping a broken sandbox.
That is recorded at the call site so a future "simplification" cannot reintroduce it quietly.

**`<empty>` is load-bearing and retained.** The capture was empty on *both* production occurrences,
so a message-only fix would have rendered `<empty>` and named nothing. An empty message beside a
`128+n` code says *signalled child*; beside `rc=1` it says something else entirely.

**Capturing the stream reopened a sanitizer's scope-lock.** The canary starts with `--env-file`, so
the live prd secret set sits in `Container.Config.Env` and is re-injected into every `docker exec`;
runc can format an env entry into stderr. `_cred_err_tail` therefore gains four shape-anchored
redaction rules and now **fails closed** via `pipefail` — a mid-pipeline tool death yields
`<sanitize_failed>` rather than a partially-sanitized leak.

## What review changed — read this part

A five-seat panel found **a live defect introduced by an earlier revision of this PR**, and a
shipped gap whose false premise I had written into a code comment. Both are fixed; recording them
because the mechanism is the interesting bit.

**`ci-deploy`'s stdout is not dark, and I asserted that it was.** It runs under
`adnanh/webhook -verbose`, which captures the hook command's combined output and re-logs it, and
`vector.toml:191` allowlists `SYSLOG_IDENTIFIER="webhook"` right next to `"ci-deploy"`. Verified in
production — `Verifying bwrap sandbox...`, a plain `echo`, is queryable in Better Stack under the
webhook tag. So the raw re-emit was shipping **unsanitized** bytes off-box seven lines before the
same bytes were sanitized for the `logger` leg. The plan had measured this and said in terms that
the comment must not repeat the discard claim; the implementation repeated it anyway. The re-emit
now prints the sanitized value and the comment states the measured topology.

**My fix for a redaction bypass over-corrected into a live defect.** Dropping the `\b` anchors
closed a word-prefix bypass and made the JWT rule match *mid-word*:
`libkeyring.so.1` → `libkeyJ.REDACTED`. A missing-shared-object error is one of the likeliest real
causes of a failing bwrap exec, so the sanitizer was destroying the diagnostic this PR exists to
produce — and every retention canary in the file is a bare `[A-Z_]` token with no `.`, `:`, `/` or
`ey`, i.e. structurally immune to the failure mode. Now `(^|[^A-Za-z0-9])` for the prefix rules
(stricter than `\b` — it admits `_`) and a literal `eyJ` for JWT, which is self-anchoring because
`eyJ` is base64 for `{"`. Retention is now fixtured with the shapes over-redaction actually eats.

**A prefix list is the same claim as a name list.** Thirteen real prd secret classes matched none of
the shape rules — including `BYOK_ENCRYPTION_KEY` (64 bare hex chars, whose loss makes every
customer's stored BYOK credential permanently unrecoverable) and `ANTHROPIC_API_KEY` (`sk-ant-`,
hyphens). Added a **value-based arm** that substitutes the container env file's own values, so it
needs no prediction; precedent is this repo's own Art. 30 register, where the #6982 `git-data-emit`
bracket records a value arm for "the LUKS passphrase, which no pattern could match".

**Eight test vacuities closed**, the sharpest being that `assert_blocking_probe_line` owns its own
verdict — so one character (`failed=1` → `failed=0`) voided every field anchor in four scenarios
while `TOTAL`/`PASS`/`FAIL` reconciled exactly, and no verdict-machinery control in the file could
see it. Closed with an in-suite positive control, using the pattern `T-7095-3` already established
in this same file.

## Verification

- `ci-deploy.test.sh`: **249/249** after the second review round (245 before it; baseline 216/216).
- `scripts/followthroughs/bwrap-probe-selfreport-8016.test.sh`: 18/18, mutation-proved 5/5 killed
  (unisolated liveness, unisolated rollback, free-text reprint, ignored threshold, PASS-on-FAIL).
- **12-row mutation battery, then a five-seat review panel** — the panel found what two batteries
  did not, and the full ledger including the two findings the test seat itself retracted is at:
  `knowledge-base/project/specs/archive/20260910-160213-feat-one-shot-8016-bwrap-probe-self-report/mutation-battery.md`.
- The fail-closed case was **vacuous on its first cut** and mutation is what caught it: it shadowed
  `sed`, which is the pipeline's *last* stage, so the status is non-zero with or without `pipefail`
  — it asserted "the last command failed". Re-pointed at `tr` (stage 2, never last) it discriminates
  exactly, and the mutant returns an *empty* string, which would render as `<empty>` and falsely
  report "the probe said nothing" when the sanitizer had died.
- `shellcheck -S warning`: no findings on added lines.
- Duration uses bash `EPOCHREALTIME`, not `date +%s%3N` — this host ships uutils coreutils 0.8.0,
  which ignores the `%3N` precision suffix and emits full nanoseconds.

## Second review round (ship-time coverage gate)

The first panel ran 5 of the 12 seats the classification calls for and emitted no coverage
trailer; with the threshold at `single-user incident` that blocks ready, so the missing seven
ran at ship time (git-history, pattern-recognition, performance, data-integrity, agent-native,
code-quality, user-impact). Fixed inline:

- **Order-dependent value arm (P2, code-quality).** `DB_HOST=db.example.com` listed before
  `DATABASE_URL=postgres://u:PASSWORD@db.example.com/x` rewrote the composite so it no longer
  matched — the password shipped in clear beside a `<redacted:DB_HOST>` marker certifying that
  redaction ran. Now longest value first, pure bash (no `sort`/`awk` on the redaction path, so a
  missing binary cannot downgrade the arm to a no-op). The helper's test now asserts the secret's
  **absence**, not just the marker's presence — the old check passed on exactly this leak.
- **PASS-path line carries `bwrap_err`** (P2, pattern + agent-native). `err_chars>0` alone said only
  that bwrap spoke; the text was on the untagged webhook leg where no literal can find it.
- **Runbook** (P1, agent-native): `canary-probe-set.md` now has the query, the `jq` decode with
  `SYSLOG_IDENTIFIER` isolation, and a field table — nothing outside the archived plan named the
  new fields.
- **64 KiB pre-clamp before the value arm** (P3, performance — measured 148 s on a 4 MB stderr made
  of a repeated env value; bash substitution is quadratic in match count). Shape rules still run on
  the full input; the 200-char truncation still comes after redaction (F14).
- F14 migrated to the shared extractor (it printed `command not found` for the value arm on every
  run); unterminated last env line now read; replacement quoted against `patsub_replacement`; the
  `$(( ))` and `err_chars` comments corrected to what was measured; non-ASCII purity leg now asserts.

Declined, with reasons: renaming `cstate` to the registry host's `state_status` (the plan, tests
and runbook all name `cstate`; cross-host queries are a P3 for a follow-up if one is ever written);
unifying the three bash redactors across cloud-init files (out of scope; cloud-init `$$` escaping
makes sharing non-trivial).

## Soak enrollment (#8016)

Ship Phase 5.5's soak gate fired: #8016's close criterion is a soak and it was not enrolled. Now
enrolled via `scripts/followthroughs/bwrap-probe-selfreport-8016.sh` (`earliest=` set past the
first post-merge deploy plus seven days): any `DEPLOY_ROLLBACK` row → exit 1 with the trusted
fields and the query (arm a, close by hand once the cause is fixed); ≥20 `SANDBOX_PROBE_OK` rows
and zero rollbacks → exit 0 (arm b, closed as environmental). Never reprints `bwrap_err` into the
public issue.

### Web Platform (deploy pipeline)

- The blocking bwrap sandbox probe in the deploy gate now reports why it failed: exit code, duration,
  container state, raw stderr length and a sanitized stderr tail ride the existing rollback line, and
  every green deploy emits a matching `SANDBOX_PROBE_OK` line. `<empty>` marks "bwrap said nothing".
- Captured stderr is sanitized fail-closed before it leaves the host: shape rules for known token
  formats plus a longest-first substitution of the container's own env values; bounded to 64 KiB
  before substitution and 200 chars on the wire.
- Runbook: `canary-probe-set.md` explains the two "sandbox" canaries, the Better Stack query, and each
  field.
- #8016 is enrolled in the follow-through sweeper; a post-incident report records the two
  undiagnosable rollbacks (2026-09-04, 2026-09-09).

### Plugin

- `review` skill: one sharp-edge bullet on retention canaries and sibling sinks (no behaviour change).

## Legal

One additive dated bracket on PA-8 §(g) of `knowledge-base/legal/article-30-register.md`, per a
binding CLO ruling (routed to the `clo` agent, not the operator — legal posture, not scope). It
ships as a **documentation obligation, not an acceptance criterion**: no test, no CI gate, no code
dependency. The reviewers' split and the ruling are recorded in `decision-challenges.md`.

Implementation note: the CLO's drafted text carried eight bare `|` characters inside backticked
regex alternations. GFM splits a table row on pipes *before* inline-code parsing, so pasted verbatim
the row would have split into nine cells and the amendment would have been **discarded at render
while surviving in the raw file**. They are escaped to `\|`, matching the convention the zot bracket
in that same cell already uses.

## User-Brand Impact

- **Artifact:** `apps/web-platform/infra/ci-deploy.sh` — the blocking bwrap probe inside the deploy
  gate, on its failure path plus one `SANDBOX_PROBE_OK` line on the success path.
- **If this lands broken, the user experiences:** an errexit-unsafe capture aborts the deploy at the
  canary stage. Production keeps serving the previous image (the cutover is downstream), but the
  EXIT trap writes `reason=unhandled` and every subsequent release fails red until the change is
  reverted — the founder cannot ship. Recovery is a revert plus a re-release. The instrumentation
  is errexit-safe by construction (`|| BWRAP_RC=$?`, `_now_ms` returns empty rather than aborting,
  a non-numeric timestamp yields `ms=unknown`), and the success-path twin means every green deploy
  now exercises the same capture code. The gate's genuine **fail-OPEN** mode is the one this PR
  guards against: reverting the capture to `if ! VAR=$(cmd)` consumes the exit code, `BWRAP_RC`
  reads 0 on a failing probe, and the rollback never fires (measured under mutation: zero
  rollback lines). A canary image with no `bwrap` binary yields rc 126/127 and fails closed.
- **If this leaks, the user's operational data is exposed via:** free text captured from the
  canary's stderr — the canary runs with `--env-file`, so prd secrets sit in `Config.Env` and runc
  can format them into an error. Two sinks: journald `ci-deploy` → Vector → Better Stack (the
  `logger` leg) and journald `webhook` → Better Stack (the guarded stdout re-emit, which
  `adnanh/webhook -verbose` re-logs). Both receive only `_cred_err_tail` output: fail-closed
  (`<sanitize_failed>` on any pipeline stage death), five shape-anchored rules, plus a value-based
  arm that substitutes the env file's own values regardless of shape, longest value first so a
  short public value cannot unmask a longer composite secret. **Preconditions and residuals,
  stated:** the value arm reads `$ENV_FILE`, which `resolve_env_file` writes before the probe and
  the EXIT trap removes after it — if that file is ever unset or unreadable at the probe site the
  arm silently degrades to the shape rules alone; values under 12 characters, values runc
  `%q`-escapes (containing `\` or `"`), and values that appear percent- or base64-encoded are
  outside the value arm; `err_chars` is the raw pre-sanitization length and is therefore a
  low-entropy length oracle on any redacted entry, accepted for its truncation-discriminator
  value (CLO on the record). The plan's third sink (a Sentry leg) was deliberately not built —
  see the archived `tasks.md` §"Deliberately NOT built".
- **Brand-survival threshold:** `single-user incident` — a secret in a log line is a
  single-tenant-class exposure; a broken gate is a founder-cannot-ship incident.

Plan (archived at compound time; `## Observability` block declares `credentials_required` because
the sink has no anonymous read):
`knowledge-base/project/plans/archive/20260910-160213-2026-09-10-fix-bwrap-probe-self-report-plan.md`

## Deploy-gate expectations

`apps/web-platform/infra/ci-deploy.sh` is a `deploy_pipeline_fix` trigger file, so
`apply-deploy-pipeline-fix.yml` auto-applies on merge — there is no operator terraform step.
/ship Phase 3.7 (`wg-dark-launch-deploy-gates`) will fire because the diff touches a deploy gate;
that is correct and expected, since a deploy-gate change cannot be validated by any deploy but the
first post-merge one. **If that release rolls back with a canary reason, suspect this change:
revert first, investigate second.**

Net issue flow is +3 and all three filings are mandatory rather than discretionary — each is a
discovered defect in a **different subsystem**, which review §5 requires to stay its own issue
rather than be bundled or fixed inline here:

- **#8036** — `docker login ghcr.io` fails on every deploy (53x in 7 days, `stage=relogin_failed`
  *after* a Doppler re-fetch), masked by the zot mirror fallback. A live credential problem behind a
  working fallback; not fixable in a deploy-gate diff.
- **#8037** — image signature verification has **never** succeeded (53/53 `cosign_absent`), so the
  soak gating `IMAGE_VERIFY_MODE=enforce` can never pass and flipping it today would fail-closed
  every deploy. Same root as #8036: `error from registry: denied` on ghcr.io.
- **#8071** — the cloud-init inngest-bootstrap pin (`v1.1.31`) is behind the published `v1.1.35`
  tag, turning the non-required `deploy-script-tests` check red on every infra PR. Pre-existing
  (both cloud-init files are unchanged here) and a recurring class (#6286 was the same chore); ship
  Phase 4 requires a tracker rather than a silent bypass.

Filed: #8036 #8037 #8071

https://claude.ai/code/session_01TZ679aF6rvtJ8CQbxNabBa


deruelle:
[deploy_pipeline_fix-drift-gate] This PR touches a trigger file. `apply-deploy-pipeline-fix.yml` will auto-apply on merge — no action required. To skip the auto-apply, add `[skip-deploy-fix-apply]` to a commit message. If the auto-apply fails (transient outage), run the workflow manually from the Actions tab, or as a last resort: `doppler run -p soleur -c prd_terraform -- terraform apply -target=terraform_data.deploy_pipeline_fix -input=true`.

Issue #8054, opened 2026-09-10
cutover op=execute is unrunnable: 2.0 needs a reachable dedicated host, but P1-5 keeps it dark on every pre-arm flag


deruelle:
## What is wrong

`op=execute`'s **2.0 empty-registry pre-flight** requires the dedicated inngest-server to be
*reachable* over GQL. Under the P1-5 arm-atomicity guard the dedicated host is **dark on every
pre-arm flag**, so 2.0 can never pass in the state it exists to guard. The only configuration that
satisfies it is `INNGEST_DIAGNOSTIC_BOOT=1` — which `op=arm`'s **G3.6 diagnostic-boot hard gate**
refuses. The two gates in the same pipeline cannot be satisfied at the same time.

## Measured

`cutover-inngest.yml -f op=execute` (run 34529824513, 2026-09-10T21:01Z), against host 165451537
(`boot_id=402c0d5b`, `probe_schema=8`):

```
##[notice] 2.-1 pool pre-check CLEAN — inngest_conns=0 ≤ readiness ceiling 12
##[error]  2.0 registry-probe returned HTTP 500: inngest-registry-probe: FATAL /v0/gql functions
           query failed or non-array (errors=["__FETCH_FAILED__"] data_keys=[]); is the dedicated
           inngest-server reachable at http://10.0.1.40:8288/v0/gql? — refusing to emit a false-…
Process completed with exit code 1.
```

The refusal is *correct* — the probe cannot distinguish "unreachable" from "empty", and emitting a
false empty would be the fail-open. The defect is that nothing can make it reachable pre-arm.

The host is dark for a good reason, and it says so itself:

```
inngest-server-flip-guard: BLOCK: prod Postgres URI with cutover flag='aborted' not in
{armed,flipping,flushed,done} — refusing inngest-server start (P1-5)
```

## The cycle

1. 2.0 needs the dedicated GQL endpoint reachable ⇒ `inngest-server` must be running.
2. `inngest-server-flip-guard` starts it only when `INNGEST_CUTOVER_FLIP ∈ {armed, flipping,
   flushed, done}`. Every pre-arm value (`aborted`, `rolled-back`, `unset`) is refused.
3. The flag leaves that set only via `op=arm` — which is supposed to run **after** `execute`.
4. The one escape, `INNGEST_DIAGNOSTIC_BOOT=1`, makes the host reachable with an empty registry and
[...]

Issue #8017, opened 2026-09-09
inngest recut is structurally unreachable: G14 compares findmnt's kernel device name against a /dev/disk/by-id path


deruelle:
## What is wrong

**G14 of `inngest_host_dark_gate` cannot be satisfied before the recut, so
`apply_target=inngest-volume-recut` can only ever be authorized on a host that has *already* been
recut.** The gate is circular and the destructive capability built by #7695 cannot legitimately fire.

## The two halves that disagree

**The emitter** (`apps/web-platform/infra/inngest-bootstrap.sh:655`):

```sh
data_mount_src="$(timeout 5 findmnt -no SOURCE "$data_mount" 2>/dev/null | head -1 || true)"
```

**The gate** (`tests/scripts/lib/inngest-host-dark-gate.sh`, G14):

```sh
expected_dev="/dev/disk/by-id/scsi-0HC_Volume_${expected_volume_id}"
if [[ "$data_mount_src" != "$expected_dev" && "$data_mount_src" != "/dev/mapper/inngest-redis" ]]; then
  _ihdg_verdict "mount_mismatch"; return $?
fi
```

`findmnt` reads `/proc/self/mountinfo`, which records the **canonical kernel device name**. It never
reports a `/dev/disk/by-id/...` path for a plain block device — the by-id path is a symlink, and the
kernel stores the resolved target. So the `expected_dev` arm is **unreachable code**. Only the
`/dev/mapper/inngest-redis` arm can ever match, and that device does not exist until the LUKS cut —
which is what the recut produces.

## Measured, three ways

**1. Live probe row** — host 165360464, boot_id `906c015b-b648-400f-93ea-ced42e048ed9`,
`2026-09-09T22:08:48Z`:

```
probe_schema=7 host_role=dedicated data_mount_src=/dev/sdb data_bytes=63728843
```

`/dev/sdb` matches neither accepted value. Volume id is `106261946`, so `expected_dev` would be
`/dev/disk/by-id/scsi-0HC_Volume_106261946`.

**2. Against the real binary** (not a stub — per the #8005 lesson):

```
$ findmnt --version
findmnt from util-linux 2.41.3
$ findmnt -no SOURCE /
/dev/nvme0n1p3
$ ls -l /dev/disk/by-id/ | grep nvme0n1p3
nvme-KXG80ZNV1T02_..._63SC63Z3EFNK-part3 -> ../../nvme0n1p3
```

[...]

Issue #7800, opened 2026-09-03
ship-incident-pir-gate: a citation of another post-mortem plus the adjective 'live' requires a PIR


deruelle:
## What

`scripts/ship-incident-pir-gate.sh` fires `INCIDENT-SIGNAL: yes` on a PR that fixes no
production incident, because `OUTAGE_RE` matches a **citation of someone else's
post-mortem** — including a bare file path — and `PROD_RE` matches the adjective
**"live"** in unrelated senses.

## Measured, on PR #7770

That PR is a scheduled burn-down of deferred scope (closes #7709). It touches only test
guards and one comment-only `.ts` line. The gate required a PIR.

`OUTAGE_RE` hits, after the gate's own strip:

```
186:   post-mortem. The isolation belongs in the consumer, not in a second pass.
344:- `knowledge-base/engineering/operations/post-mortems/net-issue-flow-gate-timeout-silent-passthrough-postmortem.md`
809:gate has a post-mortem about — a silent pass at `rc=124`.
851:gate's own timeout post-mortem.
```

Every one is a **reference to a different, already-written post-mortem**. Line 344 is a
file path in a citation list.

`PROD_RE` hits:

```
15:live worktree, whose `.git/config` is shared with every worktree on the machine.
24:- The live count is **167**, not the 160 in the issue body.
58:one reproduced live, creating a `.git` in the probe's CWD).
69:**The production guard could not fire.**
```

"live worktree", "live count", "reproduced live", "production guard" — four different
senses, none of them a production incident.

So: *a citation of an unrelated post-mortem* × *the adjective "live"* = PIR required.

## Why this is the #6813 class, narrowed but not closed

The gate's header records that its predecessor "fired on essentially every
`single-user incident` plan (incl. the preventive-hardening PR #6782), training the
operator to dismiss it." #6813 fixed that by stripping the threshold label and requiring
past-tense outage vocabulary.

It did not anticipate that a repo whose culture is *cite the post-mortem* would put
`post-mortem` in the prose of PRs that are not incidents. The richer a plan's citations,
the likelier it trips — so the gate fires hardest on the plans that did the most homework.

## Why it matters

The gate's disposition set is (a) run `/soleur:incident`, (b) author a PIR inline, or
(c) defer with a tracked issue "only when the PIR genuinely needs data not yet available."
**None of the three is "the gate mis-fired."** So the honest response to a false positive
is unrepresented, and the two cheap responses both fabricate an incident record — which is
[...]

Issue #7797, opened 2026-09-03
Two live API tokens were printed into an agent transcript by `bash -x` (Sentry + BetterStack)


deruelle:
## What happened

While building `cutover-verify.sh` for the ADR-194 apex cutover, I ran the script
under `bash -x` to debug a failing guard. The script authenticates with
`curl -H "Authorization: Bearer $TOKEN"`, and `set -x` traces every argument of
every command — so both bearer tokens were printed in full into the agent
session transcript.

**Exposed (both from Doppler `soleur/prd_terraform`):**

- `SENTRY_AUTH_TOKEN` — a `sntryu_`-prefixed Sentry user auth token
- `BETTERSTACK_API_TOKEN_READONLY` — a BetterStack API token

Values are deliberately not repeated here. They are in the agent session
transcript for session `01XR757S4JS1Ejx1mcA3ZdQj`, 2026-09-03.

## Why it is not "just a debug line"

`bash -x` is the natural next step when a token-carrying script misbehaves, and
nothing in the loop warns about it. The same reflex leaks any bearer token in any
script that passes it on an argv. Treat this as a recurring class, not a one-off.

## Blast radius

- `SENTRY_AUTH_TOKEN` is present in Doppler `prd` and `prd_terraform`, and is
  read by ~8 `scripts/followthroughs/*.sh`.
- `BETTERSTACK_API_TOKEN_READONLY` is present in `prd_terraform` and is read by
  `apps/web-platform/infra/cutover-verify.sh`.
- The Sentry token could not rotate itself: `GET /api/0/api-tokens/` returns
  **403**, so it lacks token-management scope. Rotation is a dashboard action.

## Remediation

1. Rotate `SENTRY_AUTH_TOKEN` (Sentry → User settings → Auth tokens) and write
   the new value to Doppler `soleur/prd` and `soleur/prd_terraform`.
2. Rotate `BETTERSTACK_API_TOKEN_READONLY` (BetterStack → API tokens) and write
   it to `soleur/prd_terraform`.
3. Do NOT paste either value into an agent session
   (`hr-never-paste-secrets-via-bang-prefix`).

## Prevention (the part worth keeping)

Reading a token from the environment is not enough — the leak is in the TRACE,
not the source. Two candidate fixes:

- Pass bearer tokens via `curl --config -` on stdin (`header = "Authorization: Bearer ..."`)
  rather than on argv, so `-x` cannot render them.
[...]

PR #8056, opened 2026-09-10
fix(cutover): op=execute 2.0 grades a positively-dark dedicated host instead of exiting on the P1-5 refusal (#8054)

deruelle:
## Summary

`scripts/cutover-inngest.sh op=execute` step 2.0 required the dedicated Inngest host to answer over GQL with `registry_empty=true`. The host is dark by design before `op=arm` (the P1-5 guard refuses a prod-URI start on every pre-arm flag), so the cutover's own orchestrator could not run its first step (run 34529824513: HTTP 500, `__FETCH_FAILED__`). Same class as #8017 — a predicate written against a world the surrounding guards later changed, failing closed and unnoticed until run.

2.0's property is "the dedicated host carries no function registry that could double-fire". A host that cannot start satisfies it more strongly than one that answered empty. This PR keeps the reachable-empty arm byte-for-byte and adds a second reading:

- **Lib** (`tests/scripts/lib/inngest-host-dark-gate.sh`): a second entry point, `inngest_execute_registry_gate`, grading darkness positively from the host's own hourly probe row (E1–E12) and requiring the flip FSM's ~1/min heartbeat on the SAME boot (joined on the journald `_BOOT_ID`) to attest a pre-arm flag within 15 min (E13) and to carry no FSM transition newer than the probe row (E14). E1–E7 are extracted once as `_ihdg_graded_row`, shared with the recut destroy gate. Tokens: `dark` (rc 0) plus ten refusals, each with a `gh`-verb remedy. Silence is never darkness.
- **Script**: the dark arm is admitted only on the web-host probe's own fetch-failure signature (HTTP 500 + `__FETCH_FAILED__`); any other non-200 refuses as `webhook_path` without reading Better Stack. Two reads through one generalised reader (`_bs_query_rows`, one `--grep` term each — an OR-combined read was measured starving the probe row), `set -e`-safe call shape, a `case` per token, `--emit-file` for the notice's validated fields. The 2.2 first-run "designed stop" is now a `::warning::` that precedes the `op=quiesce-web` remedy and names #6921/#8077. The script also gained the #7797 xtrace refusal and `--disable --noproxy '*'` on all 24 credentialed curls (CI's `--changed` lint unbaselines any touched file; the file left both baselines).
- **Recut workflow**: its `stale_schema` remedy derived the schema with a grep the constant refactor broke; fixed and pinned (Row 6d).
- **Suites**: dark-gate 124 → 282 (exact floor, 37 predicates, 19/19 tokens, `mutate()` classifies kills and refuses crashed mutants, shared rows redden both entry points); wiring 497 → 556 (exact; the render runs the REAL reader and remedy with `doppler` stubbed at the process boundary — argv pinned — plus `mutate_file` rows 17–20); FSM suite pins the heartbeat contract it now serves.
- **Docs**: ADR-100 addendum 2026-09-11 (#8054) §1–§8 + §4b/§4c; both Better Stack C4 edges amended (Better Stack, not Vector, parses the heartbeat).

Live read-only run at HEAD against the warehouse: `dark` rc 0 — `boot_id=402c0d5b-…`, probe row 258 s old, heartbeat 47 s old, both flags `aborted`.

Plan: `knowledge-base/project/plans/archive/20260911-191118-2026-09-10-fix-cutover-execute-dark-host-registry-gate-plan.md` (Guard Contract, mutation matrix, AC1–AC19). Learning: `knowledge-base/project/learnings/2026-09-11-the-gate-i-built-for-a-dark-host-was-blind-to-the-byte-shape-of-nothing.md`.

Ref #8054 — closed by hand after AC19 (the first post-merge `op=execute` dispatch reaching 2.2) has been read, per the closes-after-verification rule. Not `Closes`: the fix's proof lands post-merge.

Filed: #8072 #8077 #8078 #8079 #8080 #8081

Net-issue-flow override — every filing is a plan-review-surfaced defect in a DIFFERENT gate or subsystem that must stay its own issue (this PR closes nothing by keyword, by design — see Ref above):
- #8072: the reachable-empty arm accepts a host that answers pre-arm (out of sequence under P1-5) — a design question for 2.0's OTHER arm, deliberately not changed in the same PR that widens 2.0.
- #8077: `scheduled-inngest-health.yml` auto-restarts the web scheduler inside the cutover window — a watchdog-workflow defect, not this script.
- #8078: the recut destroy gate's G8 (`server_active == inactive`) refuses today's live `activating` host — a destroy-authorizing gate is not widened inside a cutover fix.
- #8079: `op=registry-probe` is unrunnable pre-arm for the same reason 2.0 was — a separate op with its own remedy path.
- #8080: post-cutover hardening (a cron satisfiability witness) — deliberately out of this PR's scope.
- #8081: the plan skill's Guard Contract needs an "assumed world, and who owns it" field — a skill-template change.

INCIDENT-SIGNAL: not-an-incident — `ship-incident-pir-gate.sh` fires on the linked plan's text, but the matched tokens are the deepen-plan-mandated `### Network-Outage Deep-Dive` heading, the `plan-network-outage-checklist.md` citation, a learning-file title and a negated Non-Goal ("No … is started"). No live event took place: the co-located web scheduler served throughout and nothing was quiesced, armed or flushed. Recorded as a new false-positive shape on #7800 rather than fixed here (different subsystem).

Soak-followthrough override: `ship-soak-followthrough-gate.sh` matched the linked plan's text, but every hit is a NEGATION or a citation — "Soak follow-through (2.9.1): not applicable … no acceptance criterion here is time-gated", "Nothing here is soak-gated", and a Non-Goal that names ADR-100's `adopting → accepted` soak (owned by #7674, which stays OPEN by operator decision). This PR declares no post-deploy close-criterion: AC19 is a single live dispatch, not a window, and `Ref #8054` closes by hand on that evidence.

Recorded in the archived spec's `decision-challenges.md`: T1 (no freshness bridge at all — not adopted: the hourly row alone cannot bound darkness to now), T2 (the reachable-arm `::warning::` — kept), T3 (ceremony of the two-stream read — kept; the one-term-per-read rule is measured, not stylistic).

## User-Brand Impact

- **If this lands broken, the user experiences:** every scheduled Inngest function firing twice (a reminder email delivered twice, a workspace reconciliation applied twice) — 2.0 is the outer repo-side layer between the cutover flip and a second scheduler registering against prod Postgres; the on-host P1-5 guard and `op=arm`'s G1 stand behind it.
- **If this leaks, the user's workflow is exposed via:** the Better Stack warehouse rows the two new reads pull. They land only in a 0700 `mktemp -d` under `RUNNER_TEMP`, are never echoed, and the notice's fields cross the gate boundary through `--emit-file` behind a shape regex; the rc-22 read-failure body is never printed (a ClickHouse 403 body names the query username).

### Web Platform (infra)
- `op=execute` 2.0 passes on a positively-dark dedicated host (probe row + same-boot FSM heartbeat) and refuses every state in which darkness cannot be established, each with a no-SSH remedy.
- The 2.2 designed-stop output warns before naming `op=quiesce-web` and cites #6921 (a second execute after quiesce fails at 2.1 capture) and #8077 (the watchdog un-quiesce).
- `apply-web-platform-infra.yml`'s recut Guard 2 `stale_schema` remedy derives the schema from the lib's constant.

### Plugin
- Recut gate lib gains `inngest_execute_registry_gate` and the shared `_ihdg_graded_row` prelude; `review`, `work` and `observability-coverage-reviewer` gain one bullet each from the review learning; constitution Testing/Never gains the process-boundary stub rule.

## Test plan

- [x] `bash tests/scripts/test-inngest-host-dark-gate.sh` → `282 passed, 0 failed` (exact floor 282; drop-one 37/35; tokens 19/19)
- [x] `bash apps/web-platform/infra/cutover-inngest-workflow.test.sh` → `556 passed, 0 failed` (exact floor 556)
- [x] `bash apps/web-platform/infra/inngest-cutover-flip.test.sh` → `150 passed, 0 failed`
- [x] `shellcheck -S warning` clean on every touched shell file; `lint-shell-trace-credential-refusal.py --changed --base origin/main` OK (0 baselined)
- [x] `python3 scripts/lint-guard-contract.py <plan>`; `bash scripts/lint-orphan-test-suites.sh`; C4 syntax + render suites; `terraform-target-parity.test.ts` 194/0
- [x] Live read-only gate run against the warehouse at HEAD → `dark` rc 0
- [x] AC15: `git diff --quiet origin/main -- .github/workflows/cutover-inngest.yml apps/web-platform/infra/inngest-bootstrap.sh apps/web-platform/infra/cloud-init-inngest.yml apps/web-platform/infra/inngest-server-flip-guard.sh apps/web-platform/infra/inngest-cutover-flip.sh` — no emitter, guard, FSM or workflow-secret change; no image bump, no host replace
- [ ] AC19 (post-merge, dispatched by this session): `gh workflow run cutover-inngest.yml --ref main -f op=execute` — expected `expected pre-arm (P1-5)`, the `dark-host arm PASSED` notice naming `boot_id=` and the flag, `2.1 capture: Σcaptured=`, `quiesce check (LB-reachable host): inngest STILL RUNNING`, then the 2.2 designed stop; the red X at 2.2 is the pass condition; `INNGEST_BASE_URL` unchanged; nothing quiesced, armed or flushed


deruelle:
## Advisory consult (fable, payload-only) — disposition

Read-only completeness opinion on the curated payload; no blocker found. Each item checked against the tree at `cafd90f43`:

| # | Advisor item | Disposition |
|---|---|---|
| 1 | Probe row ≤90 min is the only "server not active" evidence; a start outside the flag guard within that window passes | Accepted residual, same bound as the recut sibling (`max_row_age`); E14 closes the arm-then-abort window, E11 the flag-guarded start. The `::notice::` already says the P1-5 refusal is the *expected pre-arm* shape, not evidence of health. |
| 2 | E10 `server_active != active` is looser than G8 `== inactive` | Deliberate and documented in the lib (`THE DELIBERATE DIVERGENCE FROM G8`): `activating`/`failed`/`inactive` are all not-serving; E9 (`http_code != 200`, presence-checked) is the load-bearing signal. |
| 3 | `set -u` inside the reader could kill the job before any `::error::` | Reader references only its 4 args + locals; the live read-only run at HEAD executed the real reader (rc 0, verdict `dark`). |
| 4 | Reader could return 0 on a failed curl/jq under inhibited `set -e` | Reader captures rc explicitly (`|| rc=$?; return "$rc"`); does not rely on `set -e`. E1 refuses on non-zero `--query-rc`. |
| 5 | `trap … EXIT` clobbers a pre-existing trap | `cutover-inngest.sh` has exactly one `trap` line (this one). |
| 6 | Future-dated heartbeat → `fsm_silent`, probe path → `stale_row` | Both `dt` values are Better Stack ingest time, so skew affects both alike; `fsm_silent`'s first-occurrence remedy is "re-dispatch after ≥15 min", so a seconds-scale skew self-heals. Accepted. |
| 7 | AC19 depends on exact bytes of the untouched emitter/FSM | Settled by the live read-only run at HEAD against the real streams: `dark` rc 0 (boot_id 402c0d5b-…, row 258 s, hb 47 s, flag `aborted`). AC19 itself lands post-merge. |

No code change. Informational — recorded so the consult is auditable beside the 10-agent review.

Issue #7095, opened 2026-07-30
P1: prod has not deployed since 2026-07-29 — web-1's baked Doppler token was revoked (11:19:30Z), so the zot gate goes dark and the pull falls through to an unauthenticated GHCR fetch


deruelle:
**Production has not deployed since 2026-07-29.** Eight consecutive `Web Platform Release` runs have failed at `image_pull_failed`, spanning **v0.244.3, v0.245.0, v0.246.0, v0.246.1**. The site is up, serving stale code.

## The fault is isolated to the host-pull leg

PR #7071's new instrumentation separated the two legs for the first time. On the merge run (`30561788389`):

| Leg | Credential / transport | Status |
|---|---|---|
| CI **push** → zot | `REGISTRY_PUSH_ACCESS_TOKEN_*` over CF Tunnel + CF Access | **healthy** — `mirror_verified=true`, image confirmed present at the build's digest |
| Host **pull** ← zot | `ZOT_PULL_*` over the private NIC, no tunnel, no CF Access | **BROKEN** — `ci-deploy.sh exited 1 (reason=image_pull_failed, tag=v0.246.1)` |

So the operator's 2026-07-29 ~21:00 CF Access token fix repaired the push side only. The image genuinely reaches zot; the host cannot read it back.

## Suspect list (from ADR-096 clause (f), which names this exact residual)

Clause (f) states the mirror gate "asserts the manifest is in zot AND readable by the PUSH credential over the CF Tunnel. It does NOT prove the host can pull." Its three enumerated uncovered causes are the live suspects:

1. The private-net `10.0.1.10 → 10.0.1.30:5000` path is down.
2. `ZOT_PULL_*` has gone stale — **the same rotation-staleness class as the original incident trigger, one credential over.** This is the highest-prior suspect: the original trigger was a CF Access service-token rotation that did not propagate, and `ZOT_PULL_*` is subject to the identical `lifecycle.ignore_changes = [value]` invisibility.
3. zot `accessControl` grants push-read but not pull-read.

`apps/web-platform/infra/web-zot-consumer-probe.sh` is the purpose-built probe for exactly this question and should be the first thing run.

## Why this went a day without being named

Every release since the incident went red, and each red was read as "the known incident" rather than as a distinct, still-live fault. Nothing alerts on "N consecutive release-deploy failures" — the release run failing is not itself a monitored condition. That gap is worth closing alongside the pull path.

## Notes

[...]

Issue #6921, opened 2026-07-24
chore(inngest cutover): op=execute re-capture-after-quiesce ordering hazard breaks the documented "re-run op=execute" loop


deruelle:
Post-cutover follow-up from #6178 (the Phase 2 dedicated-host cutover is complete/verified; this is a documented-recovery-loop robustness defect, not a live incident).

## Problem

`op=execute` re-runs its **2.1 reminder capture** on *every* invocation. But once `op=quiesce-web` has stopped the web scheduler, the 2.1 capture can no longer reach the (now-stopped) server. So the runbook's own documented recovery instruction — "re-run `op=execute`" — fails at step 2.1 on the second pass.

This was **benign in the #6178 run** because run-1's capture is atomic-safe (the captured reminder set was already persisted on-host and consumed by `op=rearm`), so no data was lost. But the *documented loop is broken*: an operator following "re-run op=execute" after a quiesce hits a 2.1 failure that looks like a new fault.

## Fix direction

Make `op=execute`'s 2.1 capture **idempotent / skip-if-already-captured** (or gate 2.1 on "web still reachable", falling through to the retained on-host capture when quiesced), so a re-run after `op=quiesce-web` resumes from the persisted capture instead of trying to re-read a stopped server. Update the runbook step to match.

## Also (same runbook file, tiny correction)

Runbook §3b's post-cutover token-revoke command uses `doppler configs tokens revoke ... --slug <slug> --yes`, but the installed Doppler CLI **rejects `--yes`** (`Error: unknown flag: --yes`). In the #6178 revoke the orphaned key was already destroyed by the `-replace` (terraform destroy revoked it), so the manual `revoke` was a no-op — but the documented command as written would error. Drop `--yes` (revoke prompts differently) and note that `-replace` already revokes the old slug, so step (b) is a confirmation, not a required action.

Ref: `knowledge-base/engineering/operations/runbooks/inngest-server.md` (§ post-cutover, §3b).

deruelle:
## Re-measured at #8054 review (2026-09-11) — newly REACHABLE, plus a second artifact

Until #8054 (PR #8056) `op=execute` 2.0 exited 1 on every pre-arm dispatch, so this ordering hazard was unreachable in practice. #8056 makes 2.0 pass on a positively-dark host, which makes the documented loop the FIRST thing an operator meets:

[...]

PR #8135, opened 2026-09-13
fix(cutover): op=execute 2.1 capture read Doppler on the token revoked 2026-07-30 — re-read the re-delivered credential in both webhook scripts

deruelle:
## Summary

`op=execute` step 2.1 (reminder capture) died on every dispatch that reached it since 2026-07-30:

```
::error::2.1 capture returned HTTP 500: ERROR: INNGEST_MANUAL_TRIGGER_SECRET unavailable (env + doppler both empty)
```

The web host's `inngest-rearm-reminders.sh` (and its sibling `inngest-wiped-volume-verify.sh`) read Doppler with the webhook unit's `DOPPLER_TOKEN` from `/etc/default/webhook-deploy` — revoked 2026-07-30T11:19:30Z (`server.tf` "WHY THIS FILE EXISTS"). #7095 re-delivered a fresh credential at `/etc/default/soleur-doppler-token` and re-pointed `ci-deploy.sh` (parsed block) and four units (drop-ins); these two webhook scripts were not re-pointed, so every read 401'd into `2>/dev/null` and the fail-closed branch fired with no reason logged. Measured: of the three `op=execute` runs since the revocation, two died at 2.0 (#8054) and the one that cleared 2.0 — [34772240250](https://github.com/jikig-ai/soleur/actions/runs/34772240250), after #8056 — died here. The last successful capture was run 30021475229 on 2026-07-23, before the revocation.

## Change

Class fix, one block, two byte-identical copies (pinned by the new suite):

- `soleur_refresh_doppler_token()` — the #7095 parse shape (`IFS='=' read -r … || [[ -n "$k" ]]`, never sourced; the script runs as `deploy`, which holds NOPASSWD sudo on the installers), four keys (token + the three baked Sentry DSN components the same file carries), later-wins over the unit export, empty value skipped, `SOLEUR_CRED_FILE_STATE` = present|unreadable|absent.
- `soleur_doppler_read_class()` + `soleur_log_doppler_read_failure()` — on a failed/empty read, ONE classified line (`SOLEUR_DEPLOY_CRED_FAIL secret=… rc=… class=invalid_auth|secret_not_found|transport|empty_value|binary_absent|other token_file=… err="<Doppler Error line, ANSI/control-stripped, dp.xx.… scrubbed, 160 B>"`) to journald → Better Stack, mirrored to stderr → the hook's response body → the cutover run log, and a Sentry event (`op=doppler-read-failed`, class enum only) with the destination pinned by shape. Nothing here writes stdout — it runs inside `$(read_secret)`, whose stdout IS the secret.
- `read_secret()` calls the refresh before the doppler read; `mktemp` degrades to `/dev/null` instead of aborting under `set -e`; a missing `doppler` binary is classified rather than silent.
- #7797 hygiene on both files (xtrace refusal, `curl --disable --noproxy '*'` on the credentialed curls); both removed from the trace-credential baselines.
- ADR-159 dated addendum: the consumer inventory by mechanism, and the recorded decision against a `webhook.service` `EnvironmentFile=-` drop-in (empty-value blanking, sha256-pinned unit on the SSH bootstrap leg, no restart needed for the in-script read). Runbook: a no-SSH triage table for the new line.

## Verification

- `apps/web-platform/infra/webhook-doppler-token-reread.test.sh` — 76/0, exact count. Reproduces the live failure with a mock doppler that answers only to the fresh token and prints the REAL CLI's four-line coloured stderr otherwise; pins byte-identity of all three helpers; `DOPPLER_TOKEN` unset in env; present/unreadable/absent; every class; hostile `$(…)` line is data; wiped-side scrub; Sentry beacon present/refused; §I derives the webhook-executed Doppler-reader set from `hooks.json.tmpl` and requires each to carry the re-read (a fourth reader reddens). Registered in `infra-validation.yml`.
- 16-mutant sandbox battery across SUT / harness / fixture-direction / cardinality / class axes — all killed.
- rearm 41/0, wiped 23/0; `fixture-relative-assert` 62/62 (no new rows); `fixture-dir-operand-assert` 71/71; `lint-shell-capture-exit` 0 new; `lint-shell-trace-credential-refusal` OK; shellcheck; markdownlint.
- Post-merge: `apply-deploy-pipeline-fix.yml` delivers both scripts to the host via `/hooks/infra-config` (no SSH). Then re-dispatch `cutover-inngest.yml op=execute`: expect `2.0 dark-host arm PASSED` → `2.1 capture: Σcaptured=N` → the designed 2.2 stop (`STILL RUNNING`) while the web scheduler is up. Note #6921 (re-capture after quiesce) is now REACHABLE rather than masked by the dead token; it is unchanged by this PR.

### Web Platform (host scripts)
- The Inngest cutover's reminder capture (`op=execute` 2.1) and the wiped-volume verify now authenticate to Doppler with the re-delivered credential instead of the token revoked on 2026-07-30, so the capture succeeds again.
- A failed secret read now reports WHY — classified (`invalid_auth`, `secret_not_found`, `transport`, `empty_value`, `binary_absent`) in Better Stack, in the cutover run log, and as a Sentry event — instead of a bare "unavailable".
- Both scripts refuse to run under xtrace and transport-confine their credentialed curls (#7797).

### Docs
- ADR-159 addendum: consumer inventory of the re-deliverable Doppler credential by mechanism, and the recorded decision against a `webhook.service` drop-in.
- Runbook: a no-SSH triage table for the new failure line.

## User-Brand Impact

- **Artifacts exposed by this change:** `reminder.fire_at` / `reminder.action` records captured to `/var/lib/inngest/cutover-capture.json` and re-POSTed with the `INNGEST_MANUAL_TRIGGER_SECRET` Bearer; the Doppler token and three Sentry DSN components parsed from `/etc/default/soleur-doppler-token`; the classified failure line shipped to Better Stack, the hook response body and Sentry.
- **Vectors and mitigations:**
  - *Reminder loss on cutover* (a user's armed reminder never re-fires): capture previously failed closed on every dispatch; now succeeds on the re-delivered token. A failed read still fails closed with zero POSTs (C1/C9, G4). No partial capture file can be written (write is atomic; enumerate failure aborts before the file).
  - *Wrong value used as the Bearer secret*: any stdout from the failure path or a failing CLI's partial stdout would have become `SECRET`; the beacon is `>/dev/null` and `out` is cleared on rc≠0 (C28/C29). Real curl/doppler never wrote there; the mock proved the invariant.
  - *Credential in a third-party log store*: the reason line carries the `Doppler Error:` cause with `dp.<family>.<body>` tokens scrubbed, control/ANSI bytes stripped, 160-byte cap; the Sentry event carries only the class enum. Pinned per copy (C4/C5, G7/G8) and on the stderr mirror.
  - *Credentialed POST redirected by a hostile token file*: the Sentry destination is pinned by shape (`*.ingest.[xx.]sentry.io`, numeric project id) — a non-Sentry host receives nothing (C16b). The file is parsed, never sourced (E1/E2).
  - *Silent regression*: §I derives the webhook-executed Doppler-reader set from `hooks.json.tmpl`; a fourth reader without the re-read reddens CI.

## Review

8 agents (security, test-design, structural-enumeration, observability, code-quality, architecture, data-integrity + git-history, pattern + perf): 41 findings — 1 P1 (P1b ratchet red on the branch), 5 P2, rest P3 — all fixed inline in `ec7a7ae00`. Then a fable completeness consult (ADR-083): six items taken inline in `21fc84f1f`.

Full local battery at `21fc84f1f`: 411/416, 0 killed; the two reds were the trap-ownership lint (fixed in `b8c4136ac`) and the reaper mutation battery's own wall-clock ceiling under three sibling batteries (604/604 isolated at 179 s; green in CI) — filed as machinery, not caused here.

Filed: #8163

Net-issue-flow override: #8163 is a discovered defect in another subsystem (the reaper mutation battery's load-dependent ceiling), filed with `meta/machinery`; this PR closes nothing because #8054 was closed by hand on its live evidence before this PR existed.

Soak-followthrough override: no post-deploy close-criterion is declared; the verification is a single `op=execute` dispatch, not a window.

INCIDENT-SIGNAL: not-an-incident — the failing step was a pre-flight capture in a gated cutover orchestrator; the co-located web scheduler served throughout and nothing was quiesced, armed or flushed.

Ref #8054 · Ref #7095 · Ref #6921 · Ref #6178 · Ref #7695


deruelle:
[deploy_pipeline_fix-drift-gate] This PR touches a trigger file (`inngest-rearm-reminders.sh`, `inngest-wiped-volume-verify.sh`). `apply-deploy-pipeline-fix.yml` will auto-apply on merge — no action required; that apply is what delivers the fix to the host via `/hooks/infra-config`. To skip the auto-apply, add `[skip-deploy-fix-apply]` to a commit message. If the auto-apply fails (transient outage), run the workflow manually from the Actions tab, or as a last resort: `doppler run -p soleur -c prd_terraform -- terraform apply -target=terraform_data.deploy_pipeline_fix -input=true`.

Issue #8076, opened 2026-09-11
issue-flow: mandated run-reports have no honest exit from the filing gate (first live contact, #8059)


deruelle:
## Summary

Within 11 hours of #8038 merging, `cron-community-monitor` was denied on its plain filing and complied by taking exit 1 (`meta/machinery`) for #8059 — a community digest, not a machinery finding. ADR-216 says no filing exists that "must bypass the gate, cannot be labelled machinery, and cannot name a user impact". Four do: the `TASK_INVENTORY` crons whose `scheduled-<task>` issue is the heartbeat's liveness signal and the persistence handshake's precondition (community-monitor, content-generator, roadmap-review, competitive-analysis). legal-audit is in the inventory but its label covers per-gap findings, which stay gated.

Consequences today: the machinery-drain pool is 2 reports / 0 findings; the weekly measurement cannot tell relabel from reduce (class-3 run-reports are 54 of the 330 headline filings and inside the gate's supposed reach); the denial left no telemetry; run-reports have no lifecycle (43 open digests; daily triage labelled 21 of them p1-high and hid #8027's "Credit balance is too low").

## Scope

1. Substrate-written `run-report-label scheduled-<task>` directive in `cron-allow.txt` for the four crons; hook passes a filing iff a real `--label` token equals it. Not a narratable exit — the agent can neither read nor write the file.
2. `issue-flow-measure.sh`: class-3 run-report floor + exit-1 share lines; drain pool query excludes #8068 itself.
3. Lifecycle: auto-close SUCCESS run-reports at 3× `maxGapDays`; never `- FAILED` / `[cloud-task-silence]`; `scheduled-*` excluded from daily triage.
4. Hook appends denials to `.claude/filing-denials.jsonl`; substrate emits `SOLEUR_CRON_FILING_DENY` WARN post-exit.
5. Drop `meta/machinery` from #8059; amend ADR-216; update `wg-defer-only-after-inline-triage` body + ack line.
6. Two interactive-hook defects met while filing THIS issue: (a) `xargs -n1` aborts on an apostrophe anywhere in the command (a heredoc body with "Soleur's"), so exit 1 becomes unreachable and the refusal tells the filer to add the flag they passed; (b) the `--body-file` unreadable deny runs before the exit-1 pass is honoured, so a machinery filing is refused for a body it does not need.

## Artifacts

[...]

PR #8136, opened 2026-09-13
fix(inngest): resolve the taxonomy path without `new URL(…, import.meta.url)` so the Docker build compiles

deruelle:
## Summary

#8074 put the cron containment hook into the Next.js server bundle for the first time (`cron-filing-deny-marker.ts` imports `filingShape` from it). The hook spelled its taxonomy path as `new URL("../../../../.claude/hooks/lib/user-surface-taxonomy.txt", import.meta.url)` — the bundler's static asset-reference syntax — so `next build` resolved the file as a module. Every CI build passed (a full checkout has the file); the release's Docker build, whose context is `apps/web-platform`, failed with `Module not found` ([run 34773058045](https://github.com/jikig-ai/soleur/actions/runs/34773058045)) and **production stayed on the prior image**.

- Hook: `resolve(dirname(fileURLToPath(import.meta.url)), …)` — opaque to the bundler; the bundle never evaluates a filing (only the standalone `node <hook>` run does, from the sandbox's checkout), and an unreadable taxonomy is the hook's documented exit-2-does-not-apply degradation.
- Tripwire row (comment-stripped) in `cron-bash-allowlist-hook.test.ts`; fails on main's spelling.
- `web-platform-build` (ci.yml) now builds the Dockerfile's `builder` stage from the `apps/web-platform` context — the same build, from the same context, the release runs — instead of `npm run build` on the full checkout. Read-only cache from the release scope; image discarded; placeholder `NEXT_PUBLIC_` args.

Ref #8076 (the tracker stays open; this is the post-merge repair of #8074's release).

## User-Brand Impact

- **Brand-survival threshold:** none
- threshold: none, reason: no user data or user-facing surface is touched — a path spelling inside the cron sandbox's containment hook (the bundle never evaluates a filing) and a CI job; the incident's only user-visible effect was production staying on the prior image, which this PR ends.

## Test plan

- [x] Real `docker build --target builder` locally: main's hook → rc=1 with the exact release error; this hook → rc=0.
- [x] Tripwire mutation-checked against main's hook file (fails), passes on this one.
- [x] hook / deny-marker / substrate / repo-wide-containment vitest suites: 238 passed; `tsc --noEmit` clean; eslint clean.
- [x] Every suite that reads `ci.yml` green (c4-model-freshness, ci-concurrency-key, ci-test-aggregator-diagnosis, required-checks-canonical-parity, scripts-shard-runtime-coverage, scripts-shard-totality, workflow-run-deploy-invariants); `lint-workflow-install-sites` OK; actionlint only reports a pre-existing SC2034 in an unrelated block.
- [x] `web-platform-build` on this PR's own run builds the builder stage green (run 34774037658, 1m52s cold).
- [ ] Post-merge `Web Platform Release` → deploy → live-verify green.

Review (10 seats, 14 findings, 0 P1, all inline — `db67b57b4`): `web-platform-build` is now the 4th leg of the `test` aggregator (it had been advisory since creation); the job vendors `plugins/soleur` as the release does; `cache-from` removed (the release exports `mode=min`, so nothing could hit — measured zero CACHED steps); timeout 10m from the CI-measured 1m52s; `docker-context-import-containment.test.ts` now walks `.mjs`, extracts `new URL(…, import.meta.url)`, and flags targets above the app root (mutation-proven against main's hook); the taxonomy path is computed lazily; ADR-191 and the install-sites lint's BOUNDARY comment amended for the deps stage now running on `pull_request`.

https://claude.ai/code/session_01QzMAxVhPMjugs817zc4Cde


deruelle:
Review findings from a duplicate one-shot (#8139, being closed in favour of this PR — same `filingTaxonomyPath()` fix, opened later, without the PR-CI Docker builder stage this PR lands). Its review panel ran against a byte-equivalent hook, so the hook-scoped findings transfer verbatim. No P1.

**Verified, no action:** `data-integrity-guardian` ran the old (`origin/main`) and new hook from `cd /tmp` across 16 (tree × case) pairs — taxonomy present, taxonomy absent, a clone path with a space, a path with a literal `%20` — stdout/stderr/rc byte-identical; absent taxonomy degrades to DENY ("names no user-visible consequence"), never allow; `cron-bash-allowlist-hook.test.ts` still reads the real taxonomy through the lazy function (143/143, 30 reads from the worktree root). `semgrep` (73 rules, non-vacuous): 0 net-new. Docker `builder` stage on the fixed tree: exit 0, `✓ Compiled successfully`, zero hook diagnostics.

**P2 (architecture-strategist)** — the hook's CONTRACT header (lines 1–44) still describes ONE identity (stdin + argv[2]); the bundled identity lives only in the taxonomy-path comment. Suggested one-liner for the CONTRACT block: "TWO IDENTITIES since #8074: (1) standalone CLI in the sandbox clone — the only identity under which `filingTaxonomyPath()`/`main()` are meaningful; (2) an ESM import in the Next server bundle (`cron-filing-deny-marker.ts`) — module scope MUST stay side-effect-free and asset-free, and this file MUST keep importing only `node:*` builtins, because every runtime import is inside the `WRITE_PROTECTED_PATTERNS` boundary (a sibling `.mjs` the pure functions were relocated to would be Edit-able by the contained agent — that, not line count, is why option (a) is a containment change)."

**P3 (architecture-strategist, AP-021)** — when `readTaxonomy` was supplied and `surfaces.length === 0`, the deny reason names a cause the hook did not measure ("filing names no user-visible consequence"). Appending `(taxonomy unreadable at <path>)` would let the substrate's `SOLEUR_CRON_FILING_DENY` line carry the real cause under the bundled identity.

Both are ≤10 lines; posting here rather than filing (net-issue-flow).

deruelle:
Addendum from the same panel (`security-sentinel`), both **pre-existing on `main`** and outside this PR's diff — noted here rather than filed:

- **P3, hook `argumentInjectionReason` (~L415–418):** the `--body-file` secret-path denial is string-pattern only (`..`, `/proc|/etc|/root|/home`, `.git`, `.env`, `@`); a symlink under the spawn cwd pointing at `.git/config` evades it (the hook only regex-tests the bytes, but `gh` would post them). Needs a symlink-creating verb the cron allowlists do not grant today. Fix ≈ `realpathSync` the arg before pattern-testing.
- **P2-as-documentation, `scripts/lib/no-cross-context-import.test.sh` header (pre-#8139 text):** "Test files, e2e … are NOT in the Docker build" is false — `.dockerignore` prunes only `test/` (L106), has no `e2e` line, and `tsconfig.json` includes `**/*.ts`, so 9 colocated `*.test.ts(x)` files and `e2e/**` ARE in the `COPY . .` context and type-checked by `next build` (`.dockerignore` L124–127 records a release failure from exactly that). The guard's `:!**/*.test.*` / `:!e2e/**` exclusions therefore exempt build-included files (zero escapes measured today). With this PR's Docker `builder` stage in PR CI the real build catches that class, which is the better fix; the enumerating guard's sentence is just stale.

Hook verification on the identical `filingTaxonomyPath()`: allow ↔ allow before/after on plain / space / `%20` / `#` clone paths; absent taxonomy → DENY; no stdin/argv/allowlist input reaches the path computation.

deruelle:
`test-webplat (2/2)` on 0d2a7deeb is a real, PR-introduced red (run 34777959003, job 103779577523), not a flake:

```
FAIL test/repo-wide-containment.test.ts > repo-wide suite containment (#7498) > the manifest matches what actually escapes apps/web-platform
AssertionError: REPO_WIDE_SUITES has drifted from what is on disk.
+ "missing": ["test/docker-context-import-containment.test.ts"]
```

The extended `docker-context-import-containment.test.ts` now reads outside `apps/web-platform` (the hook's `.claude/…` target, presumably), so #7498's containment gate wants it registered in `test/repo-wide-suites.ts` — otherwise it sits in the GATED project and is declined on diffs that touch nothing in the app, i.e. exactly the diffs it guards. One manifest line. Everything else in that shard passed (150/151 files); the other `Error:` lines in the log are fixture noise from passing cases.

PR #8173, opened 2026-09-14
fix(inngest cutover): make execute → quiesce-web → execute → arm drivable, and stop the watchdog un-quiescing the window

### LINKED ISSUE #8077: scheduled-inngest-health auto-restart un-quiesces the web scheduler inside the cutover window — a 15-minute three-verb loop
## What is wrong

`.github/workflows/scheduled-inngest-health.yml` runs every 15 minutes and its "Auto-dispatch inngest
restart" step fires `restart-inngest-server.yml` when it reads `inngest_down` on the web host. It has
no awareness of a cutover in progress: `grep -i quiesce .github/workflows/scheduled-inngest-health.yml`
returns nothing.

`op=quiesce-web` (`scripts/cutover-inngest.sh`) stops and disables the co-located `inngest-server`
on both web hosts — which is exactly the `inngest_down` shape the watchdog remediates. So inside the
cutover window the sequence is:

1. `op=quiesce-web` stops the web scheduler (the maintenance window opens).
2. Within ≤15 min the watchdog reads `inngest_down` and dispatches `restart-inngest-server.yml`.
3. `ci-deploy.sh` on the web host runs `systemctl restart inngest-server.service` — on a DISABLED
   unit, `restart` STARTS it.
4. The second `op=execute` fails 2.2 `QUIESCE HARD GATE … STILL RUNNING`, and the operator is told
   to quiesce again.

A three-verb loop the operator cannot see the cause of, with a 15-minute fuse.

## Where it was found

Plan-review (spec-flow-analyzer, P1) of the #8054 plan, which unblocks `op=execute` and therefore
makes this the next step an operator actually reaches. The #8054 plan records it as risk R7 and
mitigates only by telling the operator in the 2.2 message to dispatch `quiesce-web → execute → arm`
back-to-back inside 15 minutes. That is a warning, not a fix.

## What is wanted

The watchdog must not un-quiesce a deliberate quiesce. Options, cheapest first:

- Read `INNGEST_CUTOVER_FLIP` (read-only, `soleur-inngest/prd`) in the health workflow and suppress
  the auto-restart while it is `armed`/`flipping`/`flushed` — the flip is in progress; or
- Have `op=quiesce-web` write a repo variable / Doppler marker the watchdog reads and honours for a
  bounded window (e.g. 2 h), with the marker cleared by `op=verify` or `op=rollback`; or
- Have the watchdog distinguish `inactive (dead) … disabled` (deliberate) from `failed` /
  `activating` (fault) in the `deploy-status` payload and only remediate the latter.

Whichever lands, add a test that a quiesced-and-disabled reading does NOT dispatch the restart.

## Where it lives

- `.github/workflows/scheduled-inngest-health.yml` — the auto-restart step
- `.github/workflows/restart-inngest-server.yml` → `apps/web-platform/infra/ci-deploy.sh` `inngest)` arm
- `scripts/cutover-inngest.sh` — `quiesce-web)` and the 2.2 gate

Related: #8054 (the plan that surfaced it, R7), #6178 (the cutover).

Mandated-By: wg-when-deferring-a-capability-create-a

deruelle:
## Summary

Makes the Inngest cutover loop `op=execute → op=quiesce-web → op=execute → op=arm → op=rearm` drivable end to end. It also stops the watchdog from un-quiescing the web scheduler inside the cutover window.

- **A quiesce marker proves the capture belongs to this stop.** The `quiesce inngest` handler captures first, then writes the sha of that capture into `/var/lib/inngest/quiesced-by-op` (with epoch, boot_id, host_id and run_id), and only then disables and stops the unit. A second `op=execute` resumes that capture only when its sha matches the marker. A stale, retired or already-consumed capture is refused.
- **One tri-state, byte-identical in three delivered scripts.** `inngest_quiesce_state` returns `quiesced`, `disabled_unattributed` or `not_quiesced`. Restart and deploy refuse on the unit shape alone, so a lost marker pages instead of restarting a deliberately stopped scheduler. Attribution (the inventory verdict, capture resume) needs both the shape and the marker.
- **The watchdog pages on the two silent states.** `DISABLED_UNATTRIBUTED` is a failure with an issue and an error check-in, and never triggers a restart. If the web scheduler stays quiesced past `INNGEST_QUIESCE_GRACE_MIN=60` while the dedicated host is not healthy, the watchdog raises `[ci/inngest-no-live-scheduler]`.
- **Step 2.2 certifies every answered probe.** Every non-200 answer must carry the sentinel. `op=quiesce-web` refuses to run while the host copies of its scripts are stale (sha preflight via `/hooks/infra-config-status`). The restart poller survives lock contention.
- **Re-arm reports what it held back.** Output is `re-armed=N failed=F held_back=H total=K`, and the cutoff includes the stop window.
- **schedule-reminder tags its two 503s.** A 503 now carries `X-Soleur-Unavailable: cutover-quiesce` or `backend-refused`. A refused `inngest.send` (TypeError whose cause code is `ECONNREFUSED`, measured on inngest 3.54.2) returns 503 + `Retry-After` instead of a terminal-looking 502. This is plan FR16. The plan said to file it as its own issue, but the issue gate refused because the fix is 20 lines / 2 files, inside the fix-inline threshold, so it is folded in here (plan C14).
- **Docs.** The runbook gains a window procedure, ADR-100 gets an amendment, and the deploy-status reason table and the learning are updated.

Closes #8077
Ref #6921. It will be closed with live evidence once the loop runs end to end during the #6178 cutover window.

Plan: knowledge-base/project/plans/archive/20260914-223454-2026-09-14-fix-inngest-cutover-loop-drivable-plan.md (User-Brand Impact threshold: `single-user incident`; Observability block with discoverability_test).

Soak gate: false positive. The only match is plan line 582, "nothing about it is soak-gated". #6921 closes on live evidence from the in-session cutover run, not on a time window.

Headless planning recorded three departures from the brief (full record: `knowledge-base/project/specs/archive/20260914-223454-feat-one-shot-6921-cutover-loop-drivable/decision-challenges.md`):

1. **No `CUTOVER_HOSTS` change.** The brief's claim that web-2 was destroyed was measured false. The Hetzner API shows `soleur-web-2` running at 10.0.1.11, and Better Stack shows 21 accepted peer deploys over 72 h. Stale prose was corrected instead.
2. **Capture-at-quiesce added beyond item (1).** Without it, a reminder armed between capture and stop is silently lost. Review later hardened it with the sha-bound marker.
3. **Plan-review simplifications applied.** The `web_scheduler` step output and its auto-close gates were deleted. Review later added the DISABLED_UNATTRIBUTED and no-live-scheduler pages, which cover the silent states that deletion left.

### Web Platform
- The cutover quiesce path now records a marker proving which stop a capture belongs to. Re-running `op=execute` after `op=quiesce-web` resumes that capture instead of failing, and refuses stale captures.
- The scheduled Inngest health check treats a deliberately quiesced web scheduler as QUIESCED instead of restarting it. It pages when a scheduler is disabled with no quiesce on record, and when no scheduler has been live for more than 60 minutes.
- `POST /api/internal/schedule-reminder` answers 503 + `Retry-After` when the Inngest backend refuses connections. Both 503s carry `X-Soleur-Unavailable`.
- Re-arm reports held-back reminders, and cutover step 2.2 certifies every probe answer.
- ci-deploy.sh's Sentry POSTs are transport-confined (`curl --disable --noproxy '*'`), and the Sentry ingest domain and project id are shape-pinned. `workspaces-cutover.sh` gains the xtrace refusal. This is the credential-lint baseline drawdown for the scripts this PR touches.

### Plugin
- Plan and review skills gain a check that any state used as a resume, replay or suppress signal carries provenance and time.

## Test plan
- [x] Integration run: all 19 affected suites plus the route test green (ci-deploy 305, rearm 176, enumerate 37, webhook-doppler 84, classify 104, wiped-volume 49, LUKS gate 34, cutover workflow 628, route 17)
- [x] Mutation rows for the quiesced arm, the declaration placement and the wiped-volume gate order
- [x] Parity sha of `inngest_quiesce_state` identical across ci-deploy.sh, inngest-inventory.sh and inngest-rearm-reminders.sh
- [x] actionlint, markdownlint, capture-exit, trap-tempfile and trace-credential lints clean
- [x] 12-agent review; 58 findings fixed inline (`Reviewed-Coverage: full 12/12 agents`)
- [ ] Post-merge: `apply-deploy-pipeline-fix.yml` delivers the new scripts, and `/hooks/infra-config-status` sha256 values match the merge commit


deruelle:
[deploy_pipeline_fix-drift-gate] This PR touches trigger files (ci-deploy.sh, inngest-inventory.sh, inngest-rearm-reminders.sh, inngest-enumerate-reminders.sh, inngest-wiped-volume-verify.sh). `apply-deploy-pipeline-fix.yml` will auto-apply on merge; no action required. To skip the auto-apply, add `[skip-deploy-fix-apply]` to a commit message. If the auto-apply fails, dispatch the workflow manually.

PR #8242, opened 2026-09-17
docs(7535): reconcile Phase 2 plan, hand impl to #8249, fix the collision gate that missed it

### LINKED ISSUE #7535: test(infra): stop the rehearsal's apt failures reading as emitter findings (image cut on measurement)
#7501 bounded the rehearsal's apt startup **in place**, at one of six container sites: `-o Acquire::Retries=3` on update, a 3-attempt backoff loop around install, and `fixture_fail` rc checks so a starved container names its own cause. That mitigates the symptom. It does not remove the dependency.

Measured (#7501 plan): `apt-get update && apt-get install -y curl python3` in `ubuntu:24.04` = **107.99 s** wall, against a **0.70 s** bare container spin. The image ships neither `curl` nor `python3`. Six sites in `apps/web-platform/infra/` pay it on every run.

## The design question

At least two further mechanisms trade off differently, and choosing between them is a design cycle rather than an edit:

1. **Pre-bake an image** carrying `curl` + `python3`. Eliminates the mirror dependency instead of retrying it, but requires an image build, somewhere to publish it, and pinning/refresh policy — and it invalidates the recorded rationale for *not* using `--network none` at these sites.
2. **`--network none`** for the sites that do not genuinely need the network once their tools are present. Cheaper than a registry, but only reachable *after* (1), and it changes what each rehearsal is actually rehearsing.

Both are durability-vs-cost-vs-complexity trades against the shipped in-place retry, which is the third option and is what #7501 chose deliberately.

## Scope-Out Justification

**Criterion: contested-design.** `code-simplicity-reviewer` (CONCUR gate) independently named the two alternatives above, judged them to trade off differently on durability (mitigate vs eliminate), cost (107.99 s × 6 per run vs image build and registry upkeep) and complexity, and recommended a design cycle outside #7501.

Two rationales I originally offered were **struck by that gate and are not claimed here**:

- *"touches six sites, therefore cross-cutting-refactor"* — fails the criterion's own text. All six live in `apps/web-platform/infra/`, the same top-level directory as #7501's primary changed file, which the criterion defines as **core**, not unrelated.
- *"pre-existing"* — false. #7501 actively rewrote the apt block at one of the six sites.

**Re-eval by (event-grep):** any R3/R4 arm reporting a fixture-starvation failure (`FIXTURE-FAIL: apt-get`) in a CI run after #7501 lands. That is the signal that the in-place retry was insufficient and the dependency has to be removed rather than bounded.

## Not part of this

The rehearsal workflow's bounded rc=2 poll was raised alongside and is **withdrawn**: `.github/workflows/git-data-rung2-rehearsal.yml` is untouched by #7501, its bound is `deadline=$(( SECONDS + 16 * 60 ))` rather than the "20 × 30 s" originally recorded, and its rc=2 step-summary is not hardcoded — it interpolates the observed rc. One residual is real and recorded in the plan's Non-Goals rather than filed: on a derivation fault the summary still reads `TRANSIENT (no verdict)` while the corrected step log reads `DERIVATION FAULT (deterministic, NOT transient)`.

Ref #7501

Ref #7507 — filed while implementing that PR (scoped out per its `deferred-scope-out` label).

deruelle:
## What this is

Plan/spec reconciliation for #7535 Phase 2, **plus two measured fixes to `one-shot`'s collision gate** that this run's own near-miss exposed.

**No product code.** #7535 Phase 2's *implementation* shipped in **PR #8249** from a parallel session; this PR carries the plan/spec reconciliation and records the handoff. Zero file overlap between the two, so both landed without conflict.

**This PR carries the close — see `## The close came back here`.** For most of this branch's life it deliberately did not: the keyword was diverted to #8249 so a docs-only PR could not resolve the issue ahead of the implementation. #8249 then merged as `4dbd1affe` carrying **no** closing link, leaving #7535 open and closed by nothing, so the close reverted here. The implementation is already on `main`, so nothing is now resolved ahead of its work.

## Artifacts

- Plan: `knowledge-base/project/plans/2026-08-13-test-remove-rehearsal-apt-dependency-plan.md`
- Spec/tasks: `knowledge-base/project/specs/feat-prebake-rehearsal-image-7535/`

(Linked explicitly because `ship-incident-pir-gate.sh` builds its corpus from a plan path in this body. Without one it emits `PIR-CORPUS-BODY-ONLY` and reports "no incident signal" having read **zero bytes of plan** — which is byte-identical to a real all-clear. Measured on this PR: it did exactly that until this section was added.)

## Why two PRs exist

Two sessions raced Phase 2. A sibling session implemented the T17-mutation and `_s1_run` sites in worktree `feat-7535-t17-rc-capture` and opened #8249 (non-draft, 107+/14-). My worktree's copy of that code was **byte-identical** and has been discarded.

The Step 0a.5 collision gate ran ~40 min *before* #8249 existed, so it cleared correctly and still missed the collision — a point-in-time PR probe cannot see a sibling that has not opened its PR yet. Recorded in the plan's `## Issue disposition` and fixed inline rather than filed as a separate issue. To be precise about what was wrong: the gate's **implementation** was correct — it ran all four probes and discriminated every hit properly — but its **specification** had two gaps, which is what the two commits below change. An earlier revision of this body said "the gate behaved as specified", which cannot be right when the specification is the thing being edited.

## Corrections this PR makes to the plan

The plan's Phase 2 target list named four sites; two were already owned elsewhere:

| Site | Owner |
|---|---|
| `run_case` split, T5-mutation split | **merged** `45ea9f7e9` (#7510) |
| R4 / `fixture_fail` | **merged** `dfcf7bd26` (#7507) |
| T17-mutation, `_s1_run` | Phase 2 → now #8249 |

Also corrected, both measured against `main`:

- **The assertion floor is `-lt 92`, not `-lt 44`.** The Phase-1-era "44 passed" baseline no longer describes this file (raised 77 → 92 by the S2 arm). Any handoff still saying "update the `-lt 44` floor" is stale.
- **The frozen 19-era baseline list above the floor must not be edited when raising it** — the file says so explicitly, and a prior review caught exactly that mistake. Raises get their own `# RAISED 92 -> N` stanza.
- Frontmatter `closes: 7535` → `refs: 7535` → **back to `closes: 7535`** once #8249 merged without the link, each step with the measured cause recorded. (`implementation_pr:` was dropped in review: a novel key, 1 of 1 in the corpus, and no code reads plan frontmatter at all — which is exactly why none of these edits ever controlled the close.)

### The keyword trap this PR walked into

An earlier revision of this body argued, in prose, that this branch must not take the close — and in doing so placed a GitHub closing keyword immediately before the issue reference. **GitHub's parser matches keyword-then-reference and does not model the negation.** So the sentence written to prevent the close is what assigned it:

```
# measured 2026-09-17, mid-session, BEFORE the fix:
gh pr view 8242 --json closingIssuesReferences  ->  [7535]   # the DOCS pr held it
gh pr view 8249 --json closingIssuesReferences  ->  []       # the IMPLEMENTATION pr held none
```

Merging this docs-only PR would have resolved the issue with the implementation still unmerged — the exact inversion this PR exists to prevent. Surfaced by the design-validity review pass, not by me.

This body now keeps every closing keyword non-adjacent to the reference, enforced by an assertion over the body text before each push. That assertion rejected my first two fix attempts, because each one re-armed the trap inside the sentence explaining it — which is the same defect class one level up, and the reason the check is mechanical rather than a careful read.

## Finding raised on #8249 rather than fixed here

The code is that PR's to own, so this went to [#8249 as review feedback](https://github.com/jikig-ai/soleur/pull/8249#issuecomment-5714128495):

#8249's T17 verdict chain reports a **capture-server bind failure** as `a genuine vacuity finding` — the #7535 defect class one hypothesis over, and a *stronger* false claim than the code it replaces. `echo T17M_APT_OK` precedes `bash /work/drive.sh`, whose bind guard exits **2**; `_T17M_ENV_RCS='100 125'` excludes 2 while the marker is present, so both marker-absent branches are false and the `else` arm certifies the finding as genuine on evidence that only ever proved apt succeeded. T5 already carries the equivalent rung in the same file. Also flagged there: its `closingIssuesReferences` measured empty.

## The collision gate fixes (two commits, both measured)

The gate cleared #7535 at 13:10 having run all four probes, correctly. #8249 opened at 13:53 — 43 min into planning. Two independent reasons it stayed invisible:

**1. The re-probe's scope was too narrow.** The post-planning re-probe added after #7247 said "re-run against any ref *not already checked* at Step 0a.5". #7535 *had* been checked, so nothing was owed. Generalized to **every** ref: the gate is point-in-time, not a lock, and the pipeline's cost sits after it, so re-asserting its own conclusion is the only thing covering the window.

**2. Both probes the gate runs are blind to an open prose-ref sibling.** Measured while running the re-probe the first commit mandates:

| probe | sees #8249? |
|---|---|
| `linked:issue #7535 --state all` | **no** |
| `#7535 in:body --state merged` (what the gate runs) | **no** — wrong state |
| `#7535 in:body --state open` | **yes**, immediately |

#8249 has `closes: []` and references #7535 in prose only, so the documented "prose-`Ref` blind spot" probe closed the *merged* half and left the *open* half — the half that matters for a live race. Body probe now runs both states.

Learning: `knowledge-base/project/learnings/workflow-patterns/2026-09-17-the-collision-gate-is-point-in-time-a-cleared-ref-goes-stale-during-planning.md`

## Verification — stated precisely

No executable **product** code. The diff is 5 markdown docs (plan, spec, tasks, a new learning, and the regenerated `knowledge-base/INDEX.md`), prose instructions in `plugins/soleur/skills/one-shot/SKILL.md` (not eval-gated — confirmed absent from `gated-skills.json`, no `eval-gate:block` markers), and **one behavioural change: `scripts/markdown-lint.test.sh`**, the mutation suite behind the required `markdown-lint` gate. That file is called out separately because an earlier revision of this body omitted it from this list entirely — review caught that, and a diff enumeration that hides its only executable file is the kind of inaccuracy this PR exists to correct elsewhere.

**The re-probe fix is self-applied:** I ran the now-mandated re-probe against #7535 after planning, and it surfaced #8249 — the fix demonstrated on the very case that motivated it.

| Check | Result |
|---|---|
| `lint-guard-contract.py` | PASS |
| `lint-credential-path-literals.py` | PASS |
| `lint-encryption-posture.py` | PASS |
| `lint-infra-no-human-steps.py` | FAIL — **pre-existing**: 518 findings on both this HEAD and `origin/main`, **zero in my files** |
| `markdown-lint.sh --repo-sweep` (CI's exact command) | PASS — 1295 files clean |
| `markdown-lint.test.sh` (the gate's own mutation suite) | PASS — 32 passed, 0 failed |
| `bash -n scripts/markdown-lint.test.sh` | PASS |
| `lint-rule-bodies --check --base origin/main` | PASS |
| `scripts/test-all.sh` | **not obtained** — refused `EXIT=4`, a sibling full-gate run was in flight (`feat-one-shot-8231-parallel-test-all`). The gate itself directs targeted suites in that case. |

Two environment problems encountered, neither caused by this diff and neither worked around silently:
- `bun` is installed (1.3.14) but **no mise version is pinned**, so `test-all.sh` aborts at startup — identically at the bare repo root. Worked around with `mise exec bun@1.3.14 --`.
- `lefthook` is not in PATH, so pre-commit hooks did not run; the lints above were run explicitly instead.

## The close came back here

The trap above was fixed and both fields re-verified `[]`. AC30b was then added to the plan for the
*other* side, because nothing asserted it: **#8249 must carry `[7535]` before it merges**, or the
issue stays open once the implementation lands. It was raised on that PR.

It failed. Measured after #8249 merged:

```
gh pr view 8249 --json state,mergedAt,closingIssuesReferences
  -> MERGED 2026-09-17T13:47:30Z, closingIssuesReferences []
gh issue view 7535 --json state,closedByPullRequestsReferences
  -> OPEN, closedBy []
```

The implementation is on `main` at `4dbd1affe`; the issue it completes was left open, closed by
nothing. A merged PR's body cannot close anything, so the remedy is for this PR — which carries
#7535's residual documentation scope — to take the close it originally owned. Safe now for the
reason it was unsafe before: the implementation has already landed, so the close cannot run ahead
of it. AC30 is inverted to assert `closingIssuesReferences == [7535]`; AC30b is kept and marked
**FAILED**, because the failure is the finding.

**The finding: an acceptance criterion whose subject you cannot write to is a monitor, not a gate.**
AC30b named the right field on the right PR and was escalated to its owner — and none of that could
satisfy it, because the control surface sat in another session's write scope. Measuring, reporting
and raising it all happened, and it was still unsatisfied at merge. `ship/SKILL.md`'s close bullet
gated only the must-not-close direction; it now covers a close assigned to a sibling PR, and says
plainly that such an assertion needs a fallback inside your own write scope.

And the original trap's inverse holds with equal force: a body that *says* `Closes` is no more
evidence than one that said `Refs`. Only `closingIssuesReferences` is, re-read after every edit.

**I did not run the docker rehearsal suite against this branch, and claim no green from it.** It is irrelevant to a markdown-only diff, and #8249 owns the code it covers.

Closes #7535


deruelle:
Heads-up on a split, to avoid duplicated work on #7535 Phase 2.

Two sessions ended up live on this branch at once. Rather than race commits here, the **code** for Phase 2 has moved to a sibling branch and PR: **#8249** (`feat-7535-t17-rc-capture`), branched from `main` and carrying no planning artifacts.

**#8249 covers the two remaining apt sites only:**
- T17-mutation — split the `&&` **and** capture rc through the `|| true` (`_t17m_rc` + a `T17M_APT_OK` in-container marker + `_T17M_ENV_RCS`). Splitting alone does not fix this site.
- `_s1_run` — named, not restructured (its two apt calls are already separate statements, so `set -e` already fires).
- Registry updates the drift guards demand: skip ceiling `7 -> 8`, call sites `3 -> 4`, `_T17_SKIPS` in `_PROBE_NAMED`, and `SKIP (loud): T17 ` registered in `scripts/followthroughs/t5-skip-persistence-bound-7510.sh`.

**This branch keeps the planning artifacts** (plan/spec/tasks), which #8249 does not touch. The two diffs are disjoint — `ff3809022` touches only `plan/spec/tasks`; #8249 touches only the test script and the follow-through probe.

Two measured findings worth folding into the plan here:

1. **The assertion floor should stay at 92.** Phase 2 as implemented adds **no** assertions — every route through the new branch contributes exactly one counted outcome, and `total = passes + fails + SKIPPED_ASSERTIONS`. Measured 92 before and after. (Separately: the floor is already `92`, not the `44` in the original brief.)
2. **The plan's `discoverability_test` was stale on `main` before this phase.** It expects `"7"` from a count of `apt-get update|install` lines; `main` measures **15**, because #7507/#7510 added `fixture_fail` *message strings* containing those words and the regex matches a literal as readily as a command. Preflight Check 10 *executes* that probe, so it would fail at ship.

Verified on #8249 with a real daemon (docker 29.7.2, terraform 1.9.8, dash 0.5.13.4 — not a `_skip`): `92 passed, 0 failed, Skipped: 0 (92 assertions)`, strip-parity `15 passed`, shell-capture lint `0 new findings, baseline untouched`. Mutation-proved in both directions plus a pre-fix control.

Issue #6894, opened 2026-07-24
encryption-posture: hcloud_volume.inngest_redis is plaintext ext4 (in-flight job payloads)


deruelle:
`hcloud_volume.inngest_redis` (`apps/web-platform/infra/inngest-host.tf:288`) is provisioned `format = "ext4"` with **no guest-side LUKS apparatus**. It holds the Inngest queue + run-state AOF — i.e. in-flight job payloads (user prompts and agent output). This is the highest-sensitivity plaintext store in the audit.

Ledgered as `mechanism: plaintext-exception` (see `scripts/encryption-posture-ledger.json`). Remediation = a guest-side LUKS cutover with its own drain/freeze/blast-radius analysis (mirror the workspaces_luks / git_data_luks apparatus). Not done in the audit PR.

`domain/engineering`, `type/security`. Ref #6588

deruelle:
**#7674 cross-reference (2026-08-25).** Work on the dedicated inngest host re-measured this volume's posture and reached the same finding this issue already records, so nothing new is filed — this comment adds the context.

Confirmed independently: `hcloud_volume.inngest_redis` is `format = "ext4"` with **zero** LUKS apparatus in `apps/web-platform/infra/inngest-host.tf`, matching the audit table row (`encryption-posture-audit-2026-07-23.md`).

Two things #7674 adds:

1. **The blast radius is wider than "in-flight job payloads".** The same volume also carries the cutover **flush latch** (`/mnt/data/inngest-cutover/flip-done.latch`), which is the durable record that a FLUSHALL has already been performed for this host. So an attacker with volume-snapshot or detached-volume access reads the queue payloads *and* the latch record in the clear.
2. **A recut vehicle is now designed and tracked** — #7695, the gated `inngest-volume-recut` `apply_target`. That is the natural place to change `format`, because a recut is already destroying and recreating the volume: switching it to LUKS at that moment costs no additional outage. Today no such target exists at all, which is part of why this posture has persisted.

Re-evaluate point recorded in the #7674 plan: the next `inngest-host-replace` window, when a recut is already being performed. Exception expiry recorded as 2026-11-30.

deruelle:
[...]

PR #8248, opened 2026-09-17
feat(inngest): additive blue-green LUKS cutover for the Redis AOF store (#6894)

deruelle:
The ADR-142 apparatus that moves the dedicated Inngest host's Redis AOF store — the sole copy of
every armed reminder, user prompt and agent output — from a plaintext Hetzner volume onto an
encrypted one, **without losing a byte and without SSH**.

**It merges inert.** Nothing is encrypted by this PR. The cutover is a separate, reviewer-gated
dispatch that an operator runs afterwards, and the two follow-through probes below report whether it
landed.

Ref #6894
Ref #7695
Ref #8017
Filed: #8316

- #8316 — the retire-or-keep decision for the `inngest-volume-recut` apply target, which this PR makes dormant. A discovered decision in a DIFFERENT subsystem (a workflow target with its own five-layer guard set), so scope discipline keeps it its own issue; it cannot be fixed inline (a target retirement is a Terraform + workflow change), and it closes on an operator decision AFTER the cutover this PR only makes possible. This PR closes nothing by design: every `Ref` above closes post-cutover, never at merge (plan ship gate 7.2).

Plan: `knowledge-base/project/plans/archive/20260918-164711-2026-09-17-infra-inngest-adr142-bluegreen-luks-apparatus-plan.md` (archived by compound; its `## User-Brand Impact` and `## Observability` sections are the ones the ship gates read).

## User-Brand Impact

- **Brand-survival threshold:** `single-user incident` — the Inngest Redis AOF is the sole copy of every armed reminder and in-flight agent run, so a torn or partial copy is a job that was accepted and then never ran, with no failure anywhere to point at.
- **Artifact:** a missing reminder in the Command Center; an agent run that silently disappears mid-flight.
- **Vector:** the copy tears, the swap lands on the wrong device, or a crash mid-swap leaves the scheduler dark — each of which the FSM now refuses or repairs (six P1 sequences fixed at review), rolls back by reverse copy, and reports off-box through its own `SyslogIdentifier` so silence is itself a finding.

## What this adds

| | |
| --- | --- |
| On-host FSM | `inngest-luks-cutover.sh` + its unit and timer: freeze every writer under `/mnt/data`, copy the **whole mount**, prove the copy byte-identical, swap, verify on the new store, and roll back by **reverse copy** if the verification fails |
| Operator verbs | `op=luks-cutover` / `op=luks-rollback` on `cutover-inngest.yml`, reviewer-gated, four fail-closed guards before the write, and a Better Stack confirm anchored at the write |
| Plan-shape gates | `inngest_host_shape_gate` (new, per-address permitted actions) and a hardened `inngest_host_replace_gate` |
| Detection | A Better Stack alert that pages when the host reports `/mnt/data` on anything other than the encrypted volume — it ships **paused**, because before the cutover the plaintext alias is the correct value |
| Records | ADR-142 amendment, ADR-225 (the latched-flag control channel), a runbook with no SSH step in it, the C4 node, the encryption-posture ledger, and Art. 30 PA-21/PA-22 |

## Why additive, and what it costs

The store is populated (442 keys), so the cheap destroy-and-recut route is refused by ADR-199. The
additive design creates a second volume alongside the live one and swaps the mount, which makes the
rollback a mount move plus a reverse copy rather than a restore from a snapshot taken at an unknown
moment.

The cost is honest and bounded: the plaintext volume **stays attached** afterwards as the rollback
backstop, so two 10 GB volumes bill concurrently. Retiring it is tracked with an expiry in #8285,
both rows are in the cost ledgers, and the store actually being on the wrong volume is *detected*
rather than assumed.

## What review changed

Seven agents. The panel found six crash sequences that would each have left the scheduler dark or
the store wiped — demonstrated against a sandbox copy, never on the live host, which stayed serving
throughout (1081 keys, `server_active=active`) — all of which were **green on the suite** — because the harness modelled a mount as a
symlink swap, so two mountpoints of one device looked like two devices. That is exactly the
distinction that decides whether the copy wipes the live store.

- A kill inside the swap left the recovering process with no phase, so it resumed nothing and then
  refused terminally — a recoverable kill became a permanently dark scheduler on a host with no way
  in. Now: resume from the freeze **record**, and repair forward when the mount says the swap
  happened.
- `copy_store` compared **paths**. A rollback killed after it remounted the plaintext volume
  re-enters with that volume at both paths, and the destination clear empties the live store before
  `cp` notices. Now compares devices, validates the source is a real mount, and short-circuits.
- The restore branch encoded "plaintext is canonical" — false for every line of rollback.
- The durable pointer was written **first**, so a crash left Doppler naming the encrypted volume
  while fstab still mounted the plaintext one. It is now the last write in both directions.
- A Redis error reply exits 0, so the keyspace sum read `0 0` and both post-swap population
  predicates went vacuous — an empty store could cut over and be marked done.
- The freeze stopped the flip **timer** but not a flip already running, which owns the one
  authorized `FLUSHALL`. The two FSMs are now mutually exclusive.

And the harness audit's headline: **deleting T2's checksum comparison left the suite green.** Every
arm handed T2 a pair `cp -a` had just made, so eight of its ten failure reasons were unreachable and
the claim "byte-equality is measured" was vacuous. T2 is now driven as the pure predicate it
advertises, the silent verdict-owners have a discrimination self-test, and the stubs reject calls
they do not model.

Review also caught a page this PR would have caused: the drift reconciler treats any live-paused
declared alert as drift, and the wrong-volume alert ships paused on purpose — a false
`logs-alert-paused` page twice daily until the cutover.

## Verification

- `inngest-luks-cutover.test.sh` — 89 assertions, driving the **real** script under `--fixture-seams`
  over a modelled block layer, so `cp -a`, `find` and `sha256sum` run for real
- `inngest-redis-luks.test.sh` 57, `cutover-inngest-workflow.test.sh` 663, `inngest.test.sh` 414,
  `ci-deploy.test.sh` 305, shape gate 79, replace gate 33, wrong-volume alert guard 19
- Mutation batteries on the FSM, both gates, and the alert; the alert's SQL was live-probed with two
  controls, including one that shows why the match carries a trailing space

## Ship-gate record

- Incident-PIR gate: `INCIDENT-SIGNAL: yes` came from the **plan**, not the body (body alone: rc 1 after one accuracy reword). Measured: the plan cites six prior post-mortems as design precedent and says "production" elsewhere, and the gate ANDs those across the whole corpus. No live event took place — the dedicated host stayed serving throughout this PR (`server_active=active`, 1081 keys, `cutover_flag=done`), and every crash sequence review found was demonstrated on a sandbox copy. No PIR is authored, because there is no incident to report; this line is the record of that decision.
- Net-issue-flow: overridden above for #8316, with the reason.
- Counsel-review: DISCHARGED by the `clo` agent — audit `knowledge-base/legal/audits/2026-09-counsel-review-8248.md`; its three conditions (a superseded-limb marker on PA-13, the two-concurrent-copies record on PA-13/21/22 with the #8285 expiry, a corrected retention sentence) are applied in `0641dff41`. Non-blocking O1 (the DPD's stale co-located serving-path sentence) is pre-existing and left for a follow-up; not filed here.
- Advisor consult (ADR-083): GAPS — three abort-path sequences that stranded a landed swap or left the scheduler dark until reboot, plus a fourth the fixture surfaced (the frozen record overwritten on re-entry). All fixed in `0641dff41`: 6 new worlds, 96/96, 26/28 mutants killed with their own FAIL line.
- Squash body: this repo squashes with `COMMIT_MESSAGES`, and commit `0ac3d5068`'s message carries an auto-close keyword next to the tracker number (in a conditional sentence GitHub's parser does not model). The merge is queued with an explicit `--body`, and `closingIssuesReferences` is asserted `[]` after every body edit — including this one, which briefly reproduced the trap by quoting the phrase.

### Web Platform (infra)
- The dedicated Inngest host can now move its Redis AOF store onto an encrypted volume without SSH and without losing a byte: a second volume, an on-host cutover state machine with byte-identity proof and automatic reverse-copy rollback, and `op=luks-cutover` / `op=luks-rollback` verbs on the cutover workflow.
- A Better Stack alert (shipping paused) pages when the store is found on the wrong volume.
- The `inngest-host` dispatch is now graded per address by `inngest_host_shape_gate`; the replace gate is hardened.
- Bootstrap image `v1.1.36` carries the cutover trio; all four pins moved with the digest.

### Plugin
- The heartbeat drift reconciler now honours a declared `paused` on a logs alert instead of paging on it.
- Review and work skills gained the Sharp Edges this round produced (a crash is not a mutation kill; `LEFTHOOK_EXCLUDE` as the narrow hatch).

## After merge

1. Observe the second volume STAGED on the next boot — #8294
2. `gh workflow run cutover-inngest.yml -f op=luks-cutover` (reviewer-gated) — #8295
3. Arm the wrong-volume alert — the runbook's §5 close-out
4. Flip the encryption-posture ledger once measured — #8296
5. Retire the plaintext backstop — #8285, expires 2026-10-22


deruelle:
Cross-link from the #7695 reconciliation (PR #8314): when this PR merges — or closes unmerged — the retire-or-keep decision for the dormant `inngest-volume-recut` target is due on #8316, together with #8285 (this volume's plaintext-backstop retirement). #8316's body lists the four stale-prose anchors in this PR's file set (`cutover-inngest.yml`, `cutover-inngest.sh`, ADR-199 §Consequences, `model.c4` `inngestRedis`) that the reconciliation PR left alone to stay disjoint from yours.

Issue #8043, opened 2026-09-10
git-data: five hash-bound hardening items, batched — erasure can report success having erased nothing, plus the authorized_keys2/hooksPath/AcceptEnv surface


deruelle:
Five hardening items on the git-data host, batched because they share one blocker and one
release: **every one of them edits a file bound by `RUNG2_TEMPLATE_SHA256`
(`3a2392fb…c1725`)**. Editing any of them — even a comment — voids the committed rung-2 boot
evidence and buys a fresh **paid** Hetzner rehearsal. Filing them separately would mean up to
five rehearsals where one discharges all five. Whatever change next legitimately moves that
hash should carry this batch.

Found during #8009 (the authorization-map gate) and co-signed by the simplicity review. All
five are things a **static** assertion over Terraform source structurally cannot see: they are
properties of a running host, or of files that host owns.

### F8 — erasure can report success having erased nothing (the serious one)

`git-data-remove.sh` derives `REPO_ROOT=/mnt/git-data/repositories` and guards with
`readlink -f`. **`readlink -f` succeeds on a path that does not exist** (verified: rc=0, prints
the path). So on a host whose volume failed to mount, both guards pass, `mkdir -p "$REPO_ROOT"`
writes into the unmounted mountpoint on the root disk, the repo is not found, and the script
prints `not present (no-op)` and **exits 0** — an Article 17 erasure success over a store
nobody looked at. A no-op and a real erasure are indistinguishable to the caller.

**What prevents it today is an accident, not an assertion.** The forced command runs as `git`;
cloud-init creates `/mnt/git-data` as root; `git-data-bootstrap.sh` chowns only the *symlink*
(`chown -h …/repositories`). So the `mkdir -p` takes EACCES and `set -euo pipefail` aborts.
Chown that mountpoint to `git` for any unrelated permissions fix and erasure begins silently
succeeding over nothing.

Fix: a `mountpoint -q` check before the guards. The invariant is documented in
`knowledge-base/engineering/operations/runbooks/git-data-birth.md` in the meantime (that file
is not hash-bound), so the trap is at least written down.

### F7 — the authorization map is rewritable by the principal it constrains

`/home/git/.ssh/authorized_keys` is `git:git 0600` inside a git-owned `.ssh`, and
`01-hardening.conf` pins no `AuthorizedKeysFile`, so sshd also honours `authorized_keys2`.
Fix: `root:root` ownership plus an `AuthorizedKeysFile` pin.

[...]

PR #8249, opened 2026-09-17
test(infra): stop a starved apt reading as a T17 mutation-battery vacuity finding

deruelle:
## Why

The T17-mutation arm's only assertion was `[ -s capture.log ]`, sitting behind an enclosing `docker run ... || true`. That pair cannot distinguish *"the mutant ran and emitted nothing"* from *"apt starved, so the mutant never ran"* — both leave an empty capture — and the arm reported both as:

> `T17 MUTATION: removing the rc guard did NOT make a healthy run emit — the check is vacuous`

An environment failure announced as a mutation-battery defect. That is the misattribution class #7535 exists to close.

## Two sites, treated deliberately differently

**T17-mutation — restructured *and* rc-captured.** Splitting the `&&` alone does not fix this site, which is why #7510 correctly declined to half-fix it. Two defects are stacked:

1. `apt-get update && apt-get install ...` — `set -e` does **not** fire on a failing **non-final** member of an AND-OR list. Measured:
   ```
   bash -c 'set -e; sh -c "exit 100" && true; echo reached'   ->  prints "reached"
   ```
   A failed update fell through into `drive.sh` with no `curl`/`python3`. The one-liner *looks* guarded under `set -e` and is not — the mechanism is now named in the code comment instead of left to be re-derived.
2. `|| true` discarded the container rc, so even after the split the `100` that `set -e` now raises was thrown away just as thoroughly as the fall-through.

The arm now captures rc via `; _t17m_rc=$?` (the `_t5m_rc` idiom) and gates on an in-container marker `T17M_APT_OK`, emitted only after both apt cycles succeed. Marker **absence** is positive evidence of a pre-mutation failure rather than an inference from an empty capture — matching `_S1_MARKER`.

**`_s1_run` — named, not restructured.** Its two apt calls are *already* separate statements, so `set -e` does fire and the container aborts with apt's own rc, which `_S1_ENV_RCS` already classifies. Nothing falls through, so there is no control flow to fix. What was missing was a **name**: `rc=100` alone does not say *which* cycle starved. Both handlers re-raise exactly the `100` `set -e` would have produced, and emit **before** `S1_FIXTURE_OK`, so classification stays `did-not-run` rather than `fixture-defect`.

## Registry consequences

Making T17 skip-eligible has knock-on effects — all caught by the suite's **own drift guards**, not by reading the edit:

- skip ceiling `7 -> 8`, itemised (`T17 mutation: 1`)
- `arm_skip` call sites `3 -> 4`, itemised
- `_T17_SKIPS` added to the `_PROBE_NAMED` roster parity check
- `SKIP (loud): T17 ` registered in `scripts/followthroughs/t5-skip-persistence-bound-7510.sh`, so the external probe is not left blind to the arm this change creates

No `Acquire::Retries` added at either site — retry stays owned by #7507 at R4.

## Verification — real daemon, not a `_skip`

The suite `_skip`s to exit 0 without docker off-CI, so a green from a docker-less box is worthless. Run with docker 29.7.2, terraform 1.9.8 and dash 0.5.13.4 all genuinely present:

| Gate | Result |
|---|---|
| `git-data-runcmd-rehearsal` | **92 passed, 0 failed, Skipped: 0 (92 assertions)** |
| `git-data-render-strip-parity` | 15 passed, 0 failed |
| `lint-shell-capture-exit` | 0 new findings, 203 baselined, **baseline file untouched** |

**The assertion floor is UNCHANGED at 92, deliberately.** Phase 2 adds no assertions — every route through the new branch still contributes exactly one counted outcome (`total = passes + fails + SKIPPED_ASSERTIONS`). Note the floor is `92`, not the `44` of Phase 1; #7510 and #8043 had already moved it.

### Mutation proof — both directions plus a pre-fix control

`--network none` starves apt deterministically without depending on a real mirror outage.

| Arm | Result |
|---|---|
| **A** pre-fix + starved | `FAIL: ... the check is vacuous` — reproduces the defect |
| **B** post-fix + starved | `SKIP (loud): T17 MUTATION did not run` — rc=100, named decline |
| **C** post-fix + healthy | `PASS (rc=0)` — the new message does **not** fire |

S1, both directions:
- starved -> `FIXTURE-FAIL: S1 apt-get install openssh-server starved`, `S1_FIXTURE_OK` absent -> `did-not-run`
- healthy -> `S1_FIXTURE_OK` present, no `FIXTURE-FAIL`

## Scope note

`Refs #7535`, **not** `Closes`. Phase 2 was briefed to close the residual scope, but a second session is concurrently working the planning artifacts (plan/spec/tasks) for this issue on branch `feat-one-shot-7535-rehearsal-apt-misattribution` (PR #8242). This PR carries only the code for the two remaining apt sites and touches no planning artifact, so closing #7535 from here would close it out from under that work. Whoever lands last should make the closure call.

## Second commit — a pre-existing gap this branch surfaced

`lint-bot-statuses` reddened on CI at the step `lint-shell-trace-credential-refusal.py --changed`. It was **not** a regression from this branch, and it is fixed here rather than deferred.

That linter runs in `--changed` mode, so it only scans files a PR touches. `scripts/followthroughs/t5-skip-persistence-bound-7510.sh` has bound `GH_TOKEN` with no xtrace refusal since it was written, and nothing had touched it since the gate landed — so the omission was real but unobserved. Registering `SKIP (loud): T17 ` in `SKIP_MARKERS` pulled the file into scope and it reddened immediately. Confirmed pre-existing: `git show origin/main:scripts/followthroughs/t5-skip-persistence-bound-7510.sh | grep -c 'refusing to trace'` returns `0`.

Without the guard, `set -x` expands the `gh` invocations carrying the token into a log that lands in a public Actions run.

Mutation-proved in both directions:

| Condition | Result |
|---|---|
| traced + `GH_TOKEN` set | `exit 78`, `[FATAL] refusing to trace with a live credential set` |
| traced + `GH_TOKEN` unset | no refusal — the guard is credential-conditional, not a blanket xtrace ban |

`lint-shell-trace-credential-refusal: OK, 2 scanned, 0 violations`


deruelle:
## Review from a parallel #7535 Phase 2 session (worktree `feat-one-shot-7535-rehearsal-apt-misattribution`, docs PR #8242)

Two sessions raced this. Yours has the implementation and mine has the plan/spec reconciliation — no file overlap, so I've discarded my copy of your code and am handing implementation to this PR. Two findings, both measured against `4c6254e1a`.

### 1. The bind-failure path still misattributes — and now asserts a *stronger* false claim

The arm closes the apt door and leaves the capture-server door open. Trace, with `drive.sh` failing to bind:

- `echo T17M_APT_OK` runs **before** `bash /work/drive.sh`, so the marker **is** present.
- `drive.sh` exits **2** on a bind failure — the guard at `(echo > /dev/tcp/127.0.0.1/8099) 2>/dev/null || { echo 'FIXTURE: capture server never bound :8099' >&2; exit 2; }`.
- `_T17M_ENV_RCS='100 125'` does **not** contain 2.

So in the verdict chain:

| branch | evaluates |
|---|---|
| `[ -s capture.log ]` | false — server never bound, capture empty |
| `! grep -qx marker && rc in allowlist` | false — marker **present** |
| `! grep -qx marker` | false — marker **present** |
| → `else` | **taken** |

The `else` arm fires:

> `T17 MUTATION: removing the rc guard did NOT make a healthy run emit — the check is vacuous`
> `… T17M_APT_OK present, so apt succeeded and this is a genuine vacuity finding`

That is the #7535 defect class one hypothesis over: an environment failure reported as a mutation-battery vacuity finding. It is *worse* than the pre-fix text, because the detail string now affirmatively certifies it as **genuine** on the strength of a marker that only ever proved apt succeeded — it was never evidence the capture server bound.

**The rung is already precedented in this file.** T5's mutation arm handles exactly this, ~100 lines up:

> `fail "T5 MUTATION: the capture server never bound :8099 — deterministic fixture defect, not an environment skip; the checksum-rejection premise is undemonstrated"`

T17 needs the equivalent branch before the `else`. `_T5_FIXTURE_MARKER='FIXTURE: capture server never bound :8099'` is already defined at the top of the file and `$TMP/out/t17m.stdout` captures stderr (`2>&1`), so the discriminator is a `grep -qF` on the text you already have — no new plumbing.

Suggested shape, keeping your three-way structure:

```sh
elif grep -qF "$_T5_FIXTURE_MARKER" "$TMP/out/t17m.stdout" 2>/dev/null; then
  fail "T17 MUTATION: the capture server never bound :8099 — deterministic fixture defect, not an environment skip; the vacuity check is undemonstrated" \
       "${_t17m_rc_note}; tail: ${_t17m_tail}"
```

A `fail` rather than an `arm_skip` matches T5's disposition: a bind failure is deterministic and owned by this file, not a precondition nobody owns.

### 2. `closingIssuesReferences` is empty — #7535 will stay open

```
gh pr view 8249 --json closingIssuesReferences  ->  {"closes":[]}
```

Phase 2 completes #7535's residual scope (Phase 1, merged `910f237f0`, correctly used `Refs`), so this PR body wants **`Closes #7535`**. My #8242 is docs-only and deliberately uses `Refs` so it cannot close the issue ahead of this implementation.

### Notes on things I checked and found correct

- The `&&` → two-statement split, and naming the `set -e` AND-OR exemption in the comment. I re-measured independently: `bash -c 'set -e; sh -c "exit 100" && true; echo reached'` prints `reached` (rc=0), and the same shape wrapped in `|| true` loses the rc entirely. Both halves of your fix are required, as your comment says.
- Reusing `arm_skip` from `45ea9f7e9` (#7510) rather than inventing a primitive.
- Registering `'SKIP (loud): T17 '` in `t5-skip-persistence-bound-7510.sh` in the same change — the roster-guard coupling your comment describes is real and easy to miss.
- The `>/dev/null 2>&1` confidentiality rationale (apt proxy credentials in error text).

### Verification status, stated honestly

I have **not** run the suite against this branch. Docker was unavailable in my session until mid-run (socket `root:docker 0660`, empty `docker` group) and is now reachable only via `newgrp docker`. Finding 1 is a static trace of the verdict chain plus the measured `exit 2` in the bind guard — it is not a suite observation. Worth reproducing with `INJECT=no-bind`, which exists for this (`[ "$INJECT" = "no-bind" ] && _PORT=8098`) and should drive the arm down the path above.

Issue #7898, opened 2026-09-07
Draw down the credential-forwarding confinement backlog (Rule D: 67 files, plus the YAML and CURL_BIN scope gaps)


deruelle:
Consolidated tracker for the credential-forwarding destination/transport confinement work that #7873 measured but deliberately did not close. One tracker rather than three issues: this is one defect class with one remedy, and splitting it would grow the backlog without adding information.

Rule D of `scripts/lint-shell-trace-credential-refusal.py` (added by #7873) enforces the property — *every credentialed `curl` carries `--disable` first and `--noproxy '*'`, and an env-settable destination needs an equality pin* — and its baseline + highwater keep the remainder visible on every run rather than forgotten.

## The census, measured 2026-09-07

| quantity | value |
|---|---|
| `curl` invocation lines in tracked `*.sh` | 472 |
| files with a credentialed `curl` Rule D reports | **67** (the baseline + highwater) |
| files already carrying `--disable` | 14 |

The ratchet is one-way: `--check-highwater` fails if that 67 grows. Drawing it down is the work below. `--changed` bypasses the baseline entirely, so **any PR touching a listed script must remediate it in the same PR** — that is the drawdown trigger, and it is enforced rather than aspirational.

## 1. Sites 2 and 3 — deferred for a deployment window, not for effort

`apps/web-platform/infra/soleur-host-bootstrap.sh:875-879` and `apps/web-platform/infra/web-private-nic-guard.sh:96-98` are one copied idiom (`soleur-host-bootstrap.sh:872` says so: *"Double-post mirrors web-private-nic-guard"*), so this is one edit applied twice.

They are not deferred because they are hard. `apps/web-platform/infra/server.tf:639-648` hashes `file("${path.module}/web-private-nic-guard.sh")` into `triggers_replace` on a `terraform_data` with an SSH `connection` block, and `apply-web-platform-infra.yml` has a `push:` trigger — so **editing site 3 re-provisions the live serving host on merge**. That is a host-replacement window and belongs to a PR that is already taking one.

Two constraints for whoever picks it up:

- **The refusal must be skip-and-report, never fail-stop.** `soleur-host-bootstrap.sh` runs under `set -e` from cloud-init and its post sits directly above the dark-host detector, so a fail-stop refusal would silence the very thing it protects.
[...]

PR #8297, opened 2026-09-18
feat(ship): operator-bootstrap skill, operator-script library, and the Merge Danger block

### LINKED ISSUE #8287: feat(ship): wizard-generated operator bootstrap scripts + Merge Danger block
Bundle 1 of 5 from the mattpocock/skills peer-plugin audit (Tier 1 entry in `knowledge-base/product/competitive-intelligence.md`, landed via https://github.com/jikig-ai/soleur/pull/8284).

**Narrative: "The founder can actually do the thing."**

User-Impact: the `/soleur:ship` generated PR body the founder reads before approving a merge; the `<feature>/bootstrap.sh` script the founder runs after merge; `/soleur:help` output; the `provision-cloudflare` / `provision-doppler` / `provision-github` / `provision-hetzner` CLI flows.
Fix-Size: ~620 lines / 12 files (measured: `template.sh` 204L + `wizard/SKILL.md` 44L ported as new assets; edits to `ship/SKILL.md` 2833L, `commands/go.md` 214L, `help/SKILL.md` 21L; refactor of 4 provision scripts totalling 1041L; 1 new 7L `wait-what` skill)
Mandated-By: hr-multi-step-post-merge-bootstrap-script
Mandated-By: hr-ship-message-no-operator-checklist

## Why

Soleur's operator-step machinery is an *accounting* system, not a *delivery* system. `wg-block-pr-ready-on-undeferred-operator-steps` blocks a PR whose body declares an operator step without a tracked automation-gap issue; `operator-digest` harvests `action-required` issues weekly. Every part answers *"has this step been deferred correctly?"* — none answers *"can the founder actually perform it?"* `ship/SKILL.md` already concedes *"a PR body is not an operator-visible surface."*

Critically, **Soleur already wrote the rules that demand a script and never built the template**:

- `hr-multi-step-post-merge-bootstrap-script` mandates a `<feature>/bootstrap.sh`
- `hr-ship-message-no-operator-checklist` mandates an `auto_command:` block
- Verified: **neither has an implementing template anywhere under `plugins/soleur/`** (both greps return zero)
- Meanwhile four `provision-*` scripts hand-roll the same `read -p "… Type 'yes'"`, and `provision-hetzner/scripts/provision-hetzner.sh` prints a numbered `echo` walkthrough with no URL opening, no progress counter, no idempotent re-run

This is live Phase 4 work: alpha tester #1 is onboarded and the per-tester process at `knowledge-base/engineering/operations/runbooks/alpha-tester-onboarding.md` is exactly the class of procedure a wizard replaces.

### G1 — wizard generator + shared library (primary)

Port `skills/engineering/wizard/` and its `template.sh` (204L). The architecture is the point: an **immutable library above a `STAGES` marker** (stage progress, confirmation gates, cross-platform URL opening incl. WSL, hidden secret entry, idempotent `.env` upserts, `gh secret`/`gh variable` writes, closing summary), with the skill authoring **only** the stages below it. "Never hand-edit the library" is the invariant that makes every wizard consistent.

Wire it as the delivery half of the two mandating rules above, and refactor the four `provision-*` scripts onto the shared library rather than leaving them hand-rolled.

### B9 — Merge Danger block in `plugins/soleur/skills/ship/SKILL.md` (Phase 6 body generation)

Add `**Door:** one-way | two-way` + `**Blast Radius:** <one word>` to the generated PR body, plus the evidence-tier ranking (screenshots S-tier when visual, execution-based evidence A-tier, show the exact test that fails then passes). `ship` is 2,833 lines of gates but the body never answers the founder's actual question: *can this be undone?*

**Constraint:** the heading must stay clear of the existing `ship-operator-step-gate` deny set (`Operator` / `Post-merge` / `Follow-up`) — mirror how the Model Dissents heading already does it.

### B12 — capability map in `plugins/soleur/commands/go.md` + `plugins/soleur/skills/help/SKILL.md`

Import `ask-matt`'s flow map (main flow / on-ramps / standalone / vocabulary-underneath) and `PHASE-BOUNDARIES.md`'s **ordered 5-option tree**: Continue → `/clear` → handoff → subagent → `/compact`, first *yes* wins, rule out **Continue** first because it costs nothing and loses nothing, `/compact` placed **last**. Soleur has `cm-when-proposing-to-clear-context-or` and `cm-delegate-verbose-exploration-3-file` as two unrelated one-liners; the ordered tree is the missing shape. At 99 skills vs their 38, the cognitive-load problem is 2.6x theirs.

**Constraint:** `go.md` routing edits are eval-gated (`eval-gate:block:go-routing`) — run `soleur:eval-harness`.

### G5 — `wait-what` (cheap rider)

7-line skill: "that didn't land — re-pitch it," in ASD-STE100 Simplified Technical English using the project's own vocabulary. `operator-digest` does plain-language translation weekly and asynchronously; nothing does it in the moment. Reads the glossary from bundle 3 when that lands.

## Founder outcome (CPO gate)

Manual steps become a script the founder runs, a reversibility verdict they can read before approving a merge, and a map telling them which of 99 skills to reach for.

## Attribution

Every file taking prose must carry:
`<!-- Inspired by mattpocock/skills/<path> (MIT, Copyright (c) 2026 Matt Pocock). -->`

---

## Naming + integration contract (operator requirement — applies to every new skill in this bundle)

**A new skill is not done when its SKILL.md exists. It is done when the Soleur workflow reaches it.**

### 1. Rename — peer names do not ship

The peer repo's names (`wizard`, `wait-what`, `to-questionnaire`, `domain-modeling`) are cute, author-specific, or opaque. They must not enter Soleur's catalog.

Note the live contradiction: `skill-creator/references/skill-structure.md:158` documents a **verb-noun** convention (`create-hooks`, `setup-stripe-payments`), but **no shipped skill uses that verb-first form** (measured: 0 of 99 match `create-*`/`setup-*`/`manage-*`/`generate-*`). What the corpus actually shows is **prefix-family grouping for related skills** so related skills sort together: `flag-create` / `flag-delete` / `flag-list`, `cron-list` / `cron-delete`, `provision-*`, `release-announce` / `release-docs`, `resolve-debt` / `resolve-parallel`, `legal-audit` / `legal-generate`, `kb-search`, `operator-digest`, `reproduce-bug`, `harvest-debt`, `merge-pr`, `drain-prs`.

**Follow the shipped convention, not the documented one** (the documented one is a defect tracked separately in the bundle-4 issue). Required renames:

| Peer name | Soleur name | Why |
|---|---|---|
| `wizard` | `operator-bootstrap` | Joins the `operator-*` family beside `operator-digest`; `hr-multi-step-post-merge-bootstrap-script` already names the artifact `bootstrap.sh`, so the skill and the mandated output share a word |
| `wait-what` | `operator-explain` | Same family; says what it does instead of being a catchphrase |
| `to-questionnaire` | `questionnaire-generate` | Matches `legal-generate` / `flag-create` verb-suffix shape |
| `domain-modeling` (glossary half) | `kb-glossary` | Joins `kb-search` / `archive-kb`; the artifact lives in `knowledge-base/` |

If the implementing plan finds a better name **inside the shipped convention**, take it — but do not keep a peer name, and do not invent a new prefix family for a single skill.

### 2. Wire it in — every surface, not just the skill file

For each new skill, all of the following must be true before the PR is ready:

- **Routing.** If the skill has a user-reachable intent, add a row to the `/soleur:go` routing table in `plugins/soleur/commands/go.md`. That block is eval-gated (`eval-gate:block:go-routing`) — run `soleur:eval-harness` and attach the result. If the skill is deliberately NOT routable from `/go` (invoked only by another skill), say so explicitly in the PR body and name the invoking skill.
- **Discoverability.** Add it to `plugins/soleur/commands/help.md` — that is the real source; `skills/help/SKILL.md` defers to it and must not be edited instead.
- **Predecessor wiring.** Name the skill in whatever existing skill should hand off to it, so it is reachable by the pipeline rather than only by a founder who already knows it exists. A skill nothing invokes is an orphan. Concretely: `operator-bootstrap` must be invoked from `ship` where the operator step is generated; `operator-explain` must be reachable mid-conversation; `questionnaire-generate` must be named where `hr-technical-fork-is-not-an-operator-question` currently dead-ends; `kb-glossary` must be read by `brainstorm`, `plan`, `spec-templates` and `triage`.
- **Lifecycle registration.** If it participates in the pipeline, add it to `plugins/soleur/lib/workflow-fidelity.ts` (`PIPELINE_SKILLS` / `HANDOFF_SKILLS` / `IMPLEMENTATION_TAIL`) — otherwise the anti-bypass guard cannot see it.
- **Manifest + docs counts.** Run `soleur:release-docs` so `plugin.json` and the README component counts stay honest.
- **Rule pointer.** Where a hard rule mandates the capability, point the rule body at the skill that now delivers it, so the rule stops being aspirational. Respect `cq-rule-ids-are-immutable` (bodies change, ids never) and the `B_ALWAYS <= 46000` byte budget.

### 3. Acceptance check

The PR body must show, per new skill: its routing row (or an explicit not-routable justification), its help.md entry, the named predecessor that invokes it, and the `release-docs` diff. A skill that appears only in its own directory does not pass.

---

**Correction (measured 2026-09-18).** An earlier revision of this issue claimed *all 99* shipped skills use `<noun|domain>-<verb>`. That does not survive measurement and is withdrawn: `resolve-*` (4) and `drain-*` (2) are verb-FIRST families, and 64 of 99 skills are not in a shared prefix family at all. The checkable claim, and the one the naming contract rests on, is narrower: **0 of 99 skills use the `create-*`/`setup-*`/`manage-*`/`generate-*` form that `skill-creator/references/skill-structure.md` prescribes**, and 14 multi-member prefix families (`flag-*`, `provision-*`, `resolve-*`, `cron-*`, `operator-*`, …) cover 35 skills. The rename table is unaffected — it was already consistent with the narrower claim.

deruelle:
## Summary

Bundle 1 of 5 from the `mattpocock/skills` peer-plugin audit. Closes the `wizard`-shaped gap Soleur had already written rules for but never implemented: `hr-multi-step-post-merge-bootstrap-script` and `hr-ship-message-no-operator-checklist` each had **zero** implementations before this PR.

Three things land:

1. **`plugins/soleur/scripts/lib/operator-script.sh`** — a shared bash library for generated operator scripts: stage progress, preflight, closing summary, exact-key `.env` upsert, stdin-only GitHub secret write, additive-only URL open, a JSON run ledger that records key *names* and never values, and a non-interactive contract (class-1 ladder value / class-2 destructive-write acknowledgement / class-3 barrier), with `exit 64` before any read when a required acknowledgement cannot be obtained.
2. **`soleur:operator-bootstrap`** — the skill that generates one of those scripts from an enumerated stage list, writing it to `knowledge-base/project/specs/feat-<name>/bootstrap.sh` (a tracked home, so it survives Phase 7 worktree reaping). It refuses to write when that directory is not gitignored for `.env`, exit 64, before the first write.
3. **`soleur:operator-rephrase`** — re-pitches a message that did not land, in a different register. Cites `operator-digest` §Register rather than restating it.

`provision-hetzner.sh` is refactored onto the library as the **proving consumer**, so the library cannot ship inert. A 21-assertion characterization suite pins its `--dry-run` stdout byte-for-byte against a committed golden captured before the refactor.

Closes #8287
Filed: #8310 #8335 #8340

Net is +2 with the override. Two filings are the "discovered defect in another subsystem that must stay its own issue" case the gate names; the third is the "filing forced by a SKILL.md phase mandate that carries no rule id" case:

- **#8310** — the remaining three `provision-*` refactors onto this library, 886 lines. Past the inline threshold, and each one carries its own blast radius (DC-3 records the measured reasons for sequencing them rather than bundling).
- **#8335** — five test suites that fail **identically on `origin/main`**, measured side by side in a pristine `--detach origin/main` worktree during the Phase 4 battery. `wg-when-tests-fail-and-are-confirmed-pre` requires the tracker; that rule is not `[mandates-filing]`-tagged, so the exemption is unavailable and adding the tag here would be a PR granting itself its own exemption (the gate reads the corpus at the merge-base precisely to stop that). Override rather than tag.
- **#8340** — the `action-required` + `decision-challenge` issue that ship Phase 6 step 2.5 (ADR-084 §5) requires whenever `decision-challenges.md` is non-empty, because `operator-digest` harvests issues, not PR bodies. Five auto-decided dissents from the headless plan/work phases, rendered under *Model Dissents* below. No rule id mandates this filing — a SKILL.md phase does — so no exemption applies.

## Naming and wiring

Both skills follow the shipped `operator-*` prefix family. No peer-repo names leaked (`wizard`, `wait-what`).

**Deliberately not routable from `/soleur:go`.** `plugins/soleur/commands/go.md` and `plugins/soleur/lib/workflow-fidelity.ts` are **byte-identical to `origin/main`** in this diff — verified, not assumed. The contract in #8287 permits a not-routable disposition provided a named invoker exists: `ship/SKILL.md:1472` invokes `skill: soleur:operator-bootstrap` at the Undeferred Operator-Step Gate, option 4. Adding a routing row instead would widen an enum, red two pinned literals, need a golden task and a regenerated projection, and cost roughly 144 API calls through the eval harness. The non-edit is the design, not an omission.

Both skills are registered in `commands/help.md` (all three harness blocks), `docs/_data/skills.js`, `plugins/soleur/NOTICE`, and the `README.md` component counts.

## Merge Danger

**Undo:** `git revert <squash-merge-sha>` — the diff is additive plus one refactor whose byte-identical behaviour is pinned by a committed golden; nothing here is stateful. The one `apps/web-platform` change is a widened read-only matcher (`MARKER_RE`); reverting it makes the `SOLEUR_BOOTSTRAP_*` family unobservable again and nothing else.
**Blast Radius:** web-platform

The plan and work phases ran headless. Five points where this session's judgement and a domain leader agreed the stated direction in #8287 should change are recorded verbatim in `knowledge-base/project/specs/archive/20260918-181155-feat-one-shot-8287-operator-bootstrap/decision-challenges.md`. Three of the five are User-Challenges against the issue text. Summarised, each with its named reversal:

- **DC-1** — the second skill is named `operator-rephrase`, not `operator-explain` as #8287's naming table says. The contract delegates this ("if the implementing plan finds a better name inside the shipped convention, take it"). Reversal: one directory rename plus three citations, under five minutes.
- **DC-2** — ASD-STE100 is cited as an *influence*, not claimed as conformance. `grep -rn "ASD-STE100|Simplified Technical English|controlled English" plugins/ knowledge-base/` returns zero, so there is no controlled vocabulary in the repo to bind the standard to and no gate that could check it. Reversal: one sentence; note it restores an unverifiable assertion.
- **DC-3** — the four-script refactor is reduced to one proving consumer. The four scripts had zero dedicated tests; the DPA gate runs before the `--dry-run` branch in all four; `provision-github.sh --dry-run` makes four network calls before that branch and therefore has no capturable baseline today. The remaining three are #8310 with a sequencing note. Reversal: re-scope #8310 back in — the characterization work lands either way.
- **DC-4** — the generated script `source`s the library rather than inlining it above a `STAGES` marker. The two shapes differ in their first five lines by construction, so no byte-identity guard can hold over both; four reviewers reached that contradiction independently, and the plan's own `SOLEUR_BOOTSTRAP_LIB_MISSING` failure mode is only meaningful for the sourced shape. Reversal is substantial and reinstates a guard whose population is gitignored and therefore invisible to CI.
- **DC-5** — `## Merge Danger` ships two fields, not three. `Door: one-way | two-way` is cut: it is derived from `Undo`, and the agent that writes the undo line runs only on PRs touching production data or migrations, so on a plugin-only PR (this one included) `Door` would be either unconstrained or forced to `one-way`. The Phase 5.5 merge hold is deferred with an activation criterion rather than a promise.

`prod-data` was renamed `user-data` in the `Blast Radius` vocabulary because `PROD_RE` in `scripts/ship-incident-pir-gate.sh:191` matches the former, which would have armed that gate on every PR declaring it.

### Plugin

- **New:** `soleur:operator-bootstrap` skill — generates a resumable, non-interactive bootstrap script for multi-step post-merge operator work, from a stage list enumerated by its caller.
- **New:** `soleur:operator-rephrase` skill — re-pitches a message the founder did not act on, in a different register.
- **New:** `plugins/soleur/scripts/lib/operator-script.sh` — the shared library those scripts source. Public surface is versioned via `SOLEUR_OP_LIB_API`.
- **New:** `## Merge Danger` block in `ship/SKILL.md`, at both PR-body template sites — two author-written fields, `Undo:` and `Blast Radius:`.
- **New:** ADR-228 — generated operator scripts are non-interactive by default.
- **New:** `operator-digest/references/phase-boundaries.md` and a capability map in `commands/help.md`.
- **Changed:** `provision-hetzner.sh` refactored onto the shared library; behaviour pinned byte-for-byte by a new characterization suite.
- **Changed:** three `AGENTS.rules.md` rule bodies now point at their implementations (`hr-multi-step-post-merge-bootstrap-script`, `hr-ship-message-no-operator-checklist`, `cm-when-proposing-to-clear-context-or`). Rule ids untouched.

## Test plan

- [x] `plugins/soleur/test/operator-script.test.sh` — 136/0 (floor 136). Guards 3 (secret handling), 4 (non-interactive contract), 5' (prologue), 7 (gitignore refusal, incl. the symlink case), 8 (pty-driven decline), 9 (consumer-side acknowledgement census), 10 (behavioural: an unattended run of the live consumer cannot reach the billable create).
- [x] `plugins/soleur/skills/provision-hetzner/test/provision-hetzner-characterization.test.sh` — 21/0, golden byte-identical to the pre-refactor capture.
- [x] `plugins/soleur/test/components.test.ts` — 1350/0, description word budget measured through `discoverSkills()`.
- [x] `plugins/soleur/test/ship-operator-step-gate.test.sh` — 22/22.
- [x] redact-sentinel — 96/0; Test 24 still sees `provision-hetzner.sh`.
- [x] `lint-shell-trace-credential-refusal` OK on 10 files; `lint-agents-rule-budget` OK; `lint-rule-ids` rc 0; `lint-orphan-test-suites` 0 orphaned.
- [x] 18/18 named mutations RED on a sandbox copy.
- [x] End-to-end founder journey driven in a scratch git repo: unattended run reaches the destructive-write acknowledgement and exits 64 naming it; a non-gitignored `.env` refuses before the first write; a re-run after an interrupt skips the completed stage rather than repeating a billable create; the EXIT trap prints the plain-language stopped-at line and writes a failed settle row to the ledger.
- [ ] Full `TEST_GROUP=all` battery — Phase 4, this run.

No browser surface in this diff; Playwright MCP is disconnected this session and no browser claim is made.

## Review

`Reviewed-Coverage: full 9/9 agents`. Two-lens design-validity pass, then the nine-seat panel, then a coverage consult. 34 findings, **all** fixed inline; zero `deferred-scope-out` issues filed from review.

The ship Phase 5.5 completeness consult then found five more, also all fixed inline (`094b81420`). Two were confirmed by execution rather than argument, and both were in the property this PR exists to establish:

- The gitignore probe read the symlink **name** while the writer `readlink -f`s before writing, so a `.env` symlinked at a shared credentials file passed the probe and landed on a path git reports as untracked-and-not-ignored. Guard 7 gains the symlink case plus a mutation row reverting the resolution — all three of its previous fixtures used a regular file.
- Guard 9 and the Guard 4 consumer sweep both stayed green when the consumer's acknowledgement was wrapped in `if [[ -z "${VAR:-}" ]]`: the call is still present and still matches, and an unattended `hcloud server create` bills a real server past it. No spelling rule closes that, because the dodge is control flow. New **Guard 10** asserts the property behaviourally — drive the live consumer with stdin closed and every escape variable the script itself names, and require the billable command was never invoked, with the variable set discovered from the script under test. Its two anti-vacuity rows both earned their keep during authoring: the own-dispatch probe caught the check deleting the script it was about to drive, and the reached-the-billable-stage assertion caught the live arm passing only because the copied script could not resolve the library and exited 64 before anything.

The Phase 4 battery then reddened four repo-global ratchets that no file-selected suite run can see, because a ratchet counts a property across the whole tree and references no changed file. All four were mine and all are fixed in `d69e67702`'s successor:

- **`guard-vacuity-floor`** — both new floors were routed *through* the suites' own `fail()`, the exact ADR-193 defect: neuter `pass()`/`fail()` and the floor that backstops them is disarmed by the same edit. Both now report via direct `printf` + `exit 1` and carry a pass+fail conservation check. The characterization suite additionally took the per-file promotion seam, so the deferral ledger **shrinks 48 → 47** rather than the ratchet growing — the guard's own instruction ("do NOT raise this number"), and the promoted entry is measured to score `FIRES`.
- **`lint-trap-tempfile-ownership`** — the library's `.env` tempfile had no owning trap. The lint's suggested remedy is wrong here: this is a *sourced* library, so `trap … EXIT` would **replace** the consumer's (`on_exit`, `cleanup`) — the terminal-outcome mechanism this feature rests on. Instead every path out of both helpers removes its own temp explicitly, both sites are annotated, and the class-b high-water is raised 80 → 81 deliberately with the reason recorded in the file, which is what that census asks for (and matches the 80th entrant, also a sourced library).
- **`lint-rule-bodies-live`** — my `AGENTS.rules.md` body edits left the corpus tamper-detection manifest stale. Regenerated.
- **`fixture-relative-assert`** — its baseline grew by exactly the 2 new test files. Regenerated with `--write-baseline` in the same commit.

CI then found one more, in the directory none of my suites touch: `apps/web-platform/test/git-lock-marker-telemetry.test.ts`'s drift guard requires every `SOLEUR_*` sentinel a plugin skill emits to be mirrored by the telemetry extractor. The `SOLEUR_BOOTSTRAP_*` family was mirrored by nothing — an operator-facing marker no layer could see. Registered by prefix, mirrored-not-paged, same closure #7898 made for `SOLEUR_TRANSPORT_DIAG`. **I had mis-triaged this**: the local battery showed it too, one of 19 failing web-platform files, and I classified the suite from its first named cause (the `localStorage` class, which CI passes) and stopped. #8335 carries the correction as a comment.

Three smaller: the proving consumer now rejects a relative `SOLEUR_OP_LIB`; `on_exit` no longer claims "Nothing else was changed" on the one path the library documents as *write may have landed*; and `phase-boundaries.md` is linked from its own SKILL.md rather than only from a backticked path in an always-loaded rule.


deruelle:
**`secret-scan-allow-rename` applied — reason on record.**

`rename-guard` flagged `plugins/soleur/skills/plan/SKILL.md → plugins/soleur/skills/plan/references/plan-sharp-edges.md`. That rename is **#8301's** (the workflow-FSM Sharp-Edges extraction, landed on `main` as `50af0436f`), not this PR's.

Why the guard attributes it here: it scans `git log --diff-merges=first-parent BASE..HEAD` with `BASE = pull_request.base.sha`, so each `Merge origin/main` sync commit on this branch is diffed against its first parent and everything `main` brought in reads as this PR's change — at the guard's 5% similarity threshold, #8301's extracted section registers as a rename into an allowlisted `references/` path. That first-parent scan exists to catch renames hidden inside conflict resolution, which is the right design; this is its false-positive shape for a branch that syncs `main` by merge four times while `main` carries such a rename.

Measured, this PR's own diff renames nothing: `git diff --name-status -M origin/main...HEAD | grep -E '^[RC]'` is empty, and `plan-sharp-edges` appears 0 times in it. The guard's own notices exempt every archive move this PR does make.

Label rather than a `Rename-Allowed-By:` trailer because a trailer would need a new commit — restarting the ~45-minute required-check cycle for a fifth time — and because a trailer naming me as allowing #8301's rename would misattribute it. The label is scoped to this PR and removable.

PR #8338, opened 2026-09-18
refactor(test): extract the cloud-init strip helpers into two shared functions with a mutation battery

### LINKED ISSUE #7968: Extract the three duplicated cloud-init strip helpers — the third copy shipped a fail-open that a green 42/42 suite certified
## The defect vector, measured

`plugins/soleur/test/cloud-init-user-data-size.test.ts` now carries **three near-verbatim copies** of two helpers:

- `registryStripRegex` / `inngestStripRegex` / `gitDataStripRegex`
- `registryStripIsApplied` / `inngestStripIsApplied` / `gitDataStripIsApplied`

They differ only in a local name and their error strings.

**This is not a hypothetical tidiness concern — the third copy shipped a fail-open, and review caught it rather than a gate.** During #7965 the first committed `inngestStripIsApplied` read the **raw** `.tf` source, where both siblings open with `stripHclLineComments(tfSrc)`. Its own docblock claimed "Mirrors registryStripIsApplied". Mutation-verified consequence: unwire `replace()`, orphan the local, and add one comment line naming the chain gives

```
control: true    MUT-A (strip UNWIRED + explanatory comment above it): true
```

i.e. **a green suite over a tree storing 42,360 B** against a 32,768 B cap — the exact tree that could not create the host. Fixed in #7965 by anchoring on the full three-link locals chain, but fixed *in the copy*, not in the class.

## Ask

**Extract Function**, parameterised on `(localName, fileLabel)` for the extractor and `(anchorRe, localName)` for the predicate, so the comment-strip-before-extract and the uniqueness requirement are structural rather than remembered three times. That makes the #7965 defect **unexpressible** instead of review-caught.

## Why it was not done inline in #7965

That PR was a production-down recovery — `hcloud_server.inngest` was destroyed and un-recreatable — so a three-way refactor of the measurement helpers was the wrong risk to take on the critical path. The arms are now well covered (47 in that suite), which is the right precondition for the extraction.

Raised by `code-quality-analyst` during #7965 review as P2-4, with the explicit sequencing note to do it after the parity arms and the walker land. Both have now landed.

Ref #7965, Ref #7695.

deruelle:
## Summary

- Collapses the three near-verbatim helper pairs in `plugins/soleur/test/cloud-init-user-data-size.test.ts` (`registryStripRegex`/`inngestStripRegex`/`gitDataStripRegex` and the three `*StripIsApplied` predicates) into two shared functions — `extractStripRegex(tfSrc, localName, fileLabel)` and `stripIsApplied(tfSrc, anchorRe, localName, chain)` — with six one-line wrappers keeping the original names, so the comment-strip-before-extract step and the local-uniqueness requirement are structural rather than remembered per copy.
- `stripIsApplied` requires the strip local to be the SECOND ARGUMENT of the `replace()` call inside the anchor's balanced span (string-aware walk + argument split), replacing the mention-in-span check every copy carried. `stripHclLineComments` strips all three HCL comment forms (line-leading, trailing, `/* block */`), string-aware and line-scoped.
- Twenty synthetic-fixture arms (`x_` host, A1–A12 incl. per-link and per-direction derivations of two must-PASS controls) plus three real-tree arms: `git-data template strip preserves #cloud-config`, `inngest user_data is unstripped when the first chain link bypasses the render`, and a walker requiring a wrapper pair for every render-wired `*_rationale_strip` local.
- New committed mutation battery `plugins/soleur/test/cloud-init-strip-helpers-mutation.test.sh` (auto-discovered by `scripts/test-all.sh`): rows R1–R9 mutate a scratch copy of the suite (comment-strip removed, uniqueness relaxed, anchor-call bound dropped, first chain link skipped, mention-vs-argument) and require the named arm under `(fail)` with `Ran N` equal to the measured pristine total; harness rows H1/H1b/H2/H3/H4 and K1/K2 drive `expect_red` and `row_red` with a green output, a wrong arm title, a synthetic load-time break, a known survivor and a known landing failure. `16 ok, 0 problem(s)`.
- Byte parity of the modelled `user_data` sizes re-measured after every helper change: 0 delta on all three hosts. Nothing under `apps/web-platform/infra/` changes.

Closes #7968
Filed: #8341

### Plugin

- `cloud-init-user-data-size.test.ts`: two shared strip helpers replace six per-host copies; the predicate now checks argument position, the comment stripper covers trailing and block comments, and 23 arms are added (70 pass / 174 expect() calls).
- New `cloud-init-strip-helpers-mutation.test.sh` battery proving the shared helpers red when the #7965 fail-open (raw-source predicate) is reintroduced.
- `work/SKILL.md` check 6.6 and `review/SKILL.md` sharp edges carry the two routed learnings (canonical `assert_fixture_dir` only; re-derive battery rows after a predicate strengthens).

## Test plan

- [x] `bun test plugins/soleur/test/cloud-init-user-data-size.test.ts` → 70 pass / 0 fail / 174 expect() calls (47 pre-existing titles unchanged)
- [x] `bash plugins/soleur/test/cloud-init-strip-helpers-mutation.test.sh` → `16 ok, 0 problem(s)`, exit 0; tracked suite never written (`git status` identical before/after)
- [x] Wrapper-drift and harness mutants driven on scratch copies (W4/W6/W7, N2/N9/N11) — all die
- [x] Byte parity old vs new helpers on registry/inngest/git-data: 0 delta
- [x] `fixture-relative-assert`, `fixture-dir-operand-assert`, `guard-vacuity-floor`, `lint-guard-contract`, `scripts-shard-totality`, `lint-orphan-test-suites`, `lint-shell-capture-exit`, shellcheck: all rc=0
- [x] Review: 10 seats, 25 findings, all fixed inline (`Reviewed-Coverage: full 10/10 agents`)
- [x] Suite + battery re-run green on the tree merged with `origin/main` (incl. #8248's `inngest-host.tf` changes)

## Gate record

- Battery: `battery-owed` reported OWED on every head because `origin/main` moved four times during the ship (#8248, 8efd7eeb3, 5396a71db, …) and was never an ancestor at check time; CI's required `test` context was green on `77f2526c8` and `cf251776e` (merge(HEAD, base) each time, 0 failures across 74–75 checks). A detached local `TEST_GROUP=all` run on `cf251776e` was killed (rc 143) after a sync merge landed under it; the run on the final tree (head `fa6dccd31`, force-pushed as `a7aef38d1` after a tree-identical commit-message reword for the auto-close scan) is recorded in a follow-up comment on this PR. The Phase 2 shard gate was refused (rc=4, sibling runs); the substitute set was every consumer suite plus the repo-global ratchets, all green.
- AC amendments: AC3's "one-line wrappers" is met as one-expression, constants-only wrappers (two span multiple lines for width); AC10's diff set additionally carries the two routed-learning edits to `plugins/soleur/skills/{work,review}/SKILL.md` (compound Route-to-Definition), and the merge commit with `origin/main`.
- Advisor consult (Phase 5.5): one fail-open in the new walker (`continue` on an unbalanceable span) fixed to throw; quote state in the span walk made line-scoped.
- Auto-close scan: one pushed commit body carried the closing keyword beside the issue number in prose; reworded via `filter-branch --msg-filter` (tree hash unchanged) and force-pushed with lease.
- rename-guard: the required check red on the sync merge with `main` `50af0436f`, whose first-parent diff carries main's own C058 copy `plugins/soleur/skills/plan/SKILL.md → references/plan-sharp-edges.md`; suppressed with a `Rename-Allowed-By: sync-merge-of-main-50af0436f` trailer commit (no rename of this branch's own is involved; the four archival moves are guard-exempt).
- Incident-PIR gate: no live event took place — the #7965 history cited in the docblocks is the mutation-found defect class this refactor makes unexpressible, already covered by ADR-152's amendment.

- T1 — env-var seam vs scratch-tree symlinks (Taste, CTO devex panel): the CTO suggested a `CLOUD_INIT_SUITE_REPO_ROOT` seam in the suite so the battery could copy it flat. Declined in-session: it adds a harness-serving seam to the SUT (the same objection that removed `// MUT:` markers), and the symlinked three-deep scratch tree is verified. No property difference; recorded in `knowledge-base/project/specs/archive/20260918-232150-feat-one-shot-7968-cloud-init-strip-helpers/decision-challenges.md`; surfaced to the operator as #8341.


deruelle:
## Phase 4 battery record (`TEST_GROUP=all`, local, detached)

- Launched 01:47 UTC+2 on head `fa6dccd31` (tree-identical to `a7aef38d1`), log `/var/tmp/ship-battery.6sOffrrA.log`, rc file read at completion.
- Terminal marker: `=== 447 suites: 441 passed, 1 failed, 0 killed, 5 skipped (declined — not relevant to this diff), 17 repo observation(s) ===`, rc=1.
- The single "failed" item is the runner's own write-boundary guard (`[FATAL] A SUITE WROTE TO THE LIVE REPOSITORY`): HEAD moved `a7aef38d1 → 8ea56ef85` during the run. That movement is this PR's two Phase 7 BEHIND auto-sync merges (`9a521229d`, `f032a17ba`) and the `Rename-Allowed-By` trailer commit (`8ea56ef85`) — operator/pipeline commits, not a suite write. Every executed suite reported `[ok]`; zero `[KILLED]`; the ten `[FAIL]` lines in the stream are suites' own expected self-test markers.
- Caveat stated plainly: the second sync (`f032a17ba`) pulled 26 non-KB files from `main` (workflow-fsm, `ci.yml`, `lefthook.yml`) under the running battery, so suites executed after ~02:17 ran against a mixed tree. The merge gate is CI's required `test` context on the pushed head, which was green on `77f2526c8` and `cf251776e` and is re-running on `8ea56ef85`.
- Contention banners during the run: `SIBLING_SUITE_DETECTED` (1 sibling suite), `LOCK_WAIT_HEARTBEAT` (60 s queue). No RED was attributed to them.

Issue #8281, opened 2026-09-18
feat(harness): WikiSkill-style skill evolution — pattern wiki, proposal ledger, gated skill creation


deruelle:
Apply the structure of Google Research's **WikiSkill** (arXiv 2608.27454, 2026-08-27, CC BY 4.0) to Soleur's existing compounding loop, in three gated phases. Brainstormed 2026-09-18.

## Why

Verified this session:

- `promotion-config.yml` is `enabled: true` since 2026-07-06, yet `promotion-log.md` has **zero rows** and no `self-healing/auto` PR or issue has ever been opened.
- The 2026-09-13 run made one Sonnet call with **516,512 input tokens / 7,623 output** (Better Stack `SOLEUR_CLAUDE_COST`, `id: cron-compound-promote`) and opened nothing. Sentry shows **0** events for `feature:cron-compound-promote` in 90d (a `has:feature` control query returns live rows). Every refusal marker is 0 over ~83 days. **The zero-output paths log no outcome**, so the cause is not currently knowable.
- `collect-corpus` sends the first 10 lines of all **2,309** learnings in one message.
- `weakness-digest.md` (2026-09-13) shows **15 learnings in one week** for a single failure class — WikiSkill forbids exactly this ("update existing patterns, do not duplicate").
- Rejected proposals are not retained anywhere a machine reads; the paper's `skill-impact.md` is what steers the next proposal.
- Skill descriptions are at **2,442 / 2,442 words — zero headroom**; **0 of 98** skills have a `PURPOSE.md`.

WikiSkill's ablation attributes **+15.0 points** (48.7% → 63.7%) to the persistent knowledge layer alone.

## Scope — 2+ phased (operator decision)

**Phase 1 — make the loop work and be visible**
- Emit `SOLEUR_COMPOUND_PROMOTE_OUTCOME` on every run (status, cluster count, each refusal reason); Sentry alert after 4 zero-output weeks.
- Fix **#8274** (diff path allowlist bypassable via `git apply -p1` prefix stripping) *before* any allowlist widening.
- Two-tier **pattern wiki**: learnings stay write-once per-session evidence; a new `patterns/` layer holds one page per pattern (root cause + fix + evidence links) plus a one-line-per-pattern index. **Full backfill** of the 2,309 learnings as its own reviewed, chunked phase — learnings are never deleted or edited.
- Proposer reads the index and pulls pages on demand, replacing the corpus dump.
[...]

PR #8344, opened 2026-09-18
fix(8281): emit the compound-promote outcome marker through pino, not Inngest's console-backed ctx.logger

deruelle:
## Summary

`SOLEUR_COMPOUND_PROMOTE_OUTCOME` — the marker #8281 exists to make observable — was emitted through Inngest's `ctx.logger`, the console-backed ProxyLogger the client deliberately leaves unconfigured (`server/inngest/client.ts`). `warn(obj, msg)` through it renders via `util.inspect` as **multi-line text** (`  SOLEUR_COMPOUND_PROMOTE_OUTCOME: true,` on one journald row, `} compound promote outcome` on another), so every field-isolated reader — the runbook decode, the #8281 follow-through probe's `.fn` select, PR #8276's own post-merge poll — reported the marker absent. Measured on the 2026-09-18 23:10:59Z manual fire: the marker landed at 23:11:54Z and was read as "not emitted" for twenty minutes.

- **New `server/compound-promote-marker.ts`** — the sibling-`*-marker.ts` form: a module-private `pino({ base: { component: "compound-promote" } })` on pino's default async, exit-flushed stdout stream (no `sync: true` — the handler runs in the long-lived Next serve process and the four siblings decode fine async), `emitOutcomeMarker(outcome)` single-argument and fail-open, holding `CompoundPromoteStatus` / `CompoundPromoteOutcome` / `REFUSAL_DETAIL_CAP`. No ADR-029 formatter, no `redact` paths: the row carries enums, counts, a run id, a cluster hash and one scrubbed error message.
- **The class, not the instance.** `SOLEUR_RUN_REPORT_SWEEP` (`cron-stale-deferred-scope-outs.ts`) had the identical defect — measured dark on the 2026-09-14 and 2026-09-15 live fires while its runbook decode claimed "at pino WARN". Routed through Marker 7 `emitRunReportSweep` in `cron-liveness-marker.ts`; a class guard in the new suite asserts no Inngest function writes a `SOLEUR_*` object (first key, literal or `[X_MARKER]`) through `ctx.logger` — RED on that site before the fix.
- **Scrub.** `error_message` and the Sentry `message` now pass `redactGithubSourcedText` (`safeDetail` strips control chars and caps at 200; `redactToken` knows only the current installation token); the always-loaded-missing message carries a repo-relative path, not the ephemeral root.
- **Tests observe the DEFAULT path.** `cron-compound-promote-outcome-marker.test.ts` partial-mocks pino with real serialization and calls `emitOutcomeMarker(outcome)` with no sink: one single-line JSON row at level 40 with the marker fields and `component`; the constructor is called once with `{ base }` only; a throwing sink never propagates; the handler's every `emitOutcomeMarker(` call is paren-balanced to exactly one argument (an optional second parameter is a mutation `tsc` cannot see). Mutants re-driven on the final tree — console sink, `process.stderr` destination, level override, dropped `try/catch`, second argument at a call site — all killed. The sweep test observes `emitRunReportSweep`, not a `ctx.logger` spy; the liveness `MARKERS` table gains the row.
- **Runbook** (`betterstack-log-query.md`): both decodes gain `select(type == "object" …)` (pre-fix rows are strings and jq-error rather than miss); `refusals_total`, the cap and `component` documented; per-refusal git `detail` is reachable in Sentry, not Better Stack; the SWEEP emitter named.

Review: 12/12 agents (`Reviewed-Coverage: full`), 24 findings, all fixed inline in `425d5ce0b` — including the P1 that the first version of this PR's own suite stayed 13/13 green with the default sink bound to `console`. Learning: `knowledge-base/project/learnings/test-failures/2026-09-19-my-fix-for-the-unpinned-wire-injected-the-seam-under-test.md`.

Refs #8281

## User-Brand Impact

- **Artifact:** the weekly self-improvement run's outcome marker and the daily run-report sweep summary (status enum, counts, run id, cluster hash, scrubbed error message) — no user data.
- **Vector:** none new; both markers gain a `component` binding and lose nothing. The pino instances carry no redaction (same as the sibling markers), which is why the field sets stay enum/count-only and the one free-text field is shape-scrubbed.
- threshold: none, reason: observability-only change on two crons that write to `knowledge-base/` and GitHub via reviewed PRs; no user-facing surface, no credential, no personal data reaches either marker.

### Web Platform

- The compound-promote outcome marker and the run-report sweep marker now reach Better Stack as single structured rows, so the #8281 runbook decode and follow-through probe can read them.
- The compound-promote error message is shape-scrubbed before it reaches Better Stack and Sentry.

### Plugin

- `review/SKILL.md`: a test that injects the seam under test leaves the default-parameter binding unpinned — drive the default path and pin call shape; `git-history-analyzer` lives under `research:`.

## Test plan

- [x] `cron-compound-promote-outcome-marker.test.ts` 5/5; 5 mutants re-driven on the final tree, all killed
- [x] 12 affected suites (marker, census, allowlist, graymatter, promoter, liveness, cert-reissue, claude-cost, sweep, repo-wide-containment) 167/167; `tsc --noEmit` clean; probe fixtures `compound-promote-outcome-8281.test.sh` 12/12; `bun test plugins/soleur/test/` 2897/2897
- [x] semgrep 79 rules: 0 findings, 0 baseline
- [x] Per-file pre-commit gates run by hand on each commit (gitleaks --staged, markdown-lint, tsc, conversations-update-callsites, agents-compound-sync, client-pii-sentry, skill-security-scan, gdpr-gate, generate-kb-index); the three commits used `LEFTHOOK=0` because a sibling worktree held the `test-all.sh` flock — the full battery was NOT run locally; the required `test` context on this PR head is the battery
- [ ] ⏳ Post-merge: after the deploy arm's `deploy` job succeeds AND `/health build_sha == merge sha`, fire `cron/compound-promote.manual-trigger` and decode with the runbook `jq` — expect one object row, `trigger=manual`, `status=error` while Anthropic credits remain exhausted (recorded on #8281); confirm the next 12:00Z stale-deferred-scope-outs fire lands a `SOLEUR_RUN_REPORT_SWEEP` object row if it closes anything


deruelle:
Admin-merge disclosure (operator-authorized green-but-BEHIND path).

Green verdict SHA: `b0ce3a6dd` — 26/26 required contexts incl. `test` on the PR head.
Merged head: `119f3e0dc` = the same PR patch rebased onto `main` @ `129fcd4d5` with `knowledge-base/INDEX.md` + `kb-tags.txt` regenerated against the merged tree (`Total files: 6650`).

Verification that nothing but the generated index changed between the green head and the merged head: `git diff <base> b0ce3a6dd` vs `git diff origin/main HEAD`, both excluding the three generated files, are byte-identical (sha256 `91749e10…`, 958 lines each). `kb-index-merge-driver.test.sh` (AC17 index parity) 78/78 on the merged tree; `rename-guard` clean; plugin suite 3102/3102.

Why not another CI cycle: this PR was resynced seven times in one pipeline — `main` landed a KB-touching PR (#8300, #8338, #8343, #8321, #8313, #8297 …) inside every ~35-min PR-CI cycle, and a squash of a stale index reds `main`'s AC17 (measured on #8343's merge). The only delta since the green verdict is the regenerated index, which is exactly the artifact AC17 verifies, and it was verified locally against the merged tree.

Issue #8010, opened 2026-09-09
git-data: git_data_rung2_rehearsal_gate checks assertion SHAPE, not that a rehearsal passed — bind the run id, resolve the run, make the Sentry crosscheck load-bearing


deruelle:
`git_data_rung2_rehearsal_gate` asserts **"a well-formed, template-bound assertion exists"** — not
**"a rung-2 rehearsal passed"**. Measured on PR #8002 at `47a00de87`:

- It **strips comments** before reading, so the host name, all four ARTIFACT query blocks and the
  capture timestamp are **not load-bearing**.
- It **never resolves the Actions run id** — the URL check is a `=~` shape match only. It does not
  check the run exists, its conclusion, or which workflow it belongs to.
- **`RUNG2_SENTRY_CROSSCHECK` is read by no consumer.** The capture writes it
  (`git-data-rung2-evidence-capture.sh`), its own test asserts it exists, and the gate's
  required-key loop covers only the other four. `UNAVAILABLE` and junk both RELEASE.
- The hash is a **staleness detector, not an authorship proof** — a pure function of tracked files,
  computable by anyone with a checkout.

Measured: a **four-line hand-written file** naming a nonexistent run releases the gate identically.

The gate knows the compensating control is absent — its own refusal message justifies the strict
URL regex by saying an unauditable claim is not evidence *"and this repository's main branch has no
required-review ruleset standing behind a hand-typed one."* That is still true, and
`deploy-script-tests` (which runs the freshness check) is **not** in the required-check ruleset, so
the gate library itself can be edited and merged with neither its suite nor a reviewer required.

## Three fixes, cheapest first

1. **Free, no artifact change.** `rehearsal.tf` builds the host name **from** the run id
   (`soleur-git-data-rehearsal-${var.rehearsal_run_id}`), so the URL's run id and the host-name
   suffix are already coupled — measured equal (`33888071954`) in the committed evidence. Assert
   `URL run id == host-name suffix`. Note the host name currently lives in a **comment**, and the
   gate strips comments precisely so comments are never load-bearing — so this either reads raw
   text (reversing that decision) or promotes the host name to a key (which needs a regeneration).
   **Decide deliberately.**
2. Resolve the run in CI (`gh api repos/{owner}/{repo}/actions/runs/<id>`) and assert
   `workflow_id`. Assert workflow **identity**, *not* `conclusion == success` — the current,
   legitimate evidence would fail that, because its run is marked `failure` (the in-run capture
[...]

PR #8393, opened 2026-09-19
chore(git-data): commit the rung-2 boot + reboot-reopen evidence from run 35465756680

deruelle:
## Evidence-only PR — nothing else is in it, by design

Commits `apps/web-platform/infra/git-data-rung2-boot-evidence.env` from rehearsal run
[35465756680](https://github.com/jikig-ai/soleur/actions/runs/35465756680), dispatched from `main` `d64430c26` with `dry_run=false`, `teardown_only=false`, approved through the `web-platform-infra-apply` environment (2026-09-19 19:54:12Z → 20:09:21Z). The committed file is byte-identical to the workflow artifact (`cmp` clean).

This is PM2 of #8210's post-merge sequence (PM1 = this dispatch, PM3 = the daily sweeper retires #8210 once `main` carries the key below). #8312 deliberately deleted the previous evidence, so the gate has been HOLDing by absence since; **merging this file re-RELEASES `git_data_rung2_rehearsal_gate`** for the current template. It is not auto-committed by the workflow (`permissions: contents: read`): the route that produces the evidence must not be the route that releases the hold.

### What the run established
- `RUNG2_BOOT_REHEARSAL=PASS` — `stage:boot_complete` reached, no `level:fatal` on Better Stack, source-liveness anchor answered.
- **`RUNG2_REBOOT_REOPEN=PASS`, `CHANNEL=both`, `RESTARTS=0`** — the #8312 reset arm rebooted the throwaway host and `/dev/mapper/git-data` reopened unattended, observed on both Better Stack and Sentry (`luks_reopen_ok`), with zero unit restarts. This is the datum #8210 waits on.
- Template sha256 `a0b5f37b2fefcb095c1d5e4a559dfac047840db87bb087ffdab5c78d514a0517` — the post-#8312 template; 17 hash-bound inputs.
- Teardown ran (no `teardown_only` recovery needed; the redact/upload-on-non-PASS steps were skipped because the run PASSed).
- `RUNG2_SENTRY_CROSSCHECK=UNAVAILABLE` — this key is the *other-hosts* Sentry liveness anchor over the ~2 min boot window (`count=0`), and the gate does not read it today. It is not in tension with the reboot bullet above: the reboot arm's Sentry read on *this* host answered with the `luks_reopen_ok` row three minutes later (`CHANNEL=both`), so the window was quiet, not the read path dead — in-band corroboration the 09-13 evidence only had after the fact. Making the key load-bearing is #8010 (PR #8388, in flight). The gate is evaluated on every workflow run, not at merge time, so merge order does not protect this file: #8010 must ship an explicit ack/grandfather path for `UNAVAILABLE`-with-live-reopen-read evidence, or this evidence gets re-captured under the new script.

### Gate, run against this commit
```
git_data_rung2_rehearsal_gate: RELEASED — … attests PASS for user_data sha256 a0b5f37b… (…/runs/35465756680) … provenance: PASS — … last modified in commit 1ad973bb8…, which touched none of the 17 hash-bound inputs.
```

## User-Brand Impact

- threshold: none, reason: the diff is a machine-generated attestation file with no runtime consumer — it changes no code path, credential, or user-owned data; it releases one arm of a dispatch gate whose dispatch is separately environment-approved.

### What it does NOT do
- It does not birth or replace anything. PM4 (`git-data-host-replace`) stays blocked on #8362 (apply workflow over the 512 kB ceiling).
- It does not retire the tracker (#8210) — the sweeper's `git-data-reboot-evidence-landed-8210.sh` probe does that once this is on `main` (earliest 2026-09-20T02:00Z).

Provenance guard HOLDs on any co-edit with a hash-bound file, so **please do not push additional commits to this branch**.

Ref #8210


deruelle:
**Sequencing, now measured rather than predicted.** #8010 (PR #8388's branch) makes `RUNG2_SENTRY_CROSSCHECK` load-bearing. Run against **this PR's exact bytes**, in a scratch worktree detached at `d64430c26`, with the real Actions API and no test seam:

```
git_data_rung2_rehearsal_gate: HOLD [SENTRY_UNAVAILABLE_UNACKED] — … records
RUNG2_SENTRY_CROSSCHECK=UNAVAILABLE … append RUNG2_SENTRY_CROSSCHECK_ACK=35465756680:<why>
```
rc=1. With the ack appended in its own commit, the same bytes go `RELEASED` (rc 0), naming `head_sha d64430c26…`, the re-hashed digest `a0b5f37b…`, the uploaded artifact, and `Sentry cross-check UNAVAILABLE (acknowledged)`.

**Nothing changes for this PR, and that is the decision, not an oversight.** The gate is evaluated on every workflow run, not at merge time, so merge order protects nothing — what matters is that `main` carries the ack by the time the stricter gate lands. #8010's branch therefore appends it as its own evidence-only commit after this merges (both Guard 4 arms pass: that commit touches no hash-bound file). Keeping this PR a pure evidence PR is worth more than saving one commit.

So: **do not push the ack here.** This branch stays byte-identical to the run artifact, which is the property its review panel verified.

